PR #2, opened 2015-11-23
Copy in the elvis files

Licenser:
This commit copies all 'core' elvis file form the elvis repository
including tests. It clens up code that was related to non core
tasks and removes the tests for that code.


jfacorro:
@Licenser Hi, thanks for your work on this PR! I have a few ideas and comments I wanted to share.

I was thinking that maybe we could make `elvis_core` leaner, although I don't have a clear idea of exactly what that looks like. For example all output to screen is actually related to the command-line tool, so it doesn't feel right that there are calls to [`elvis_utils:info/1`](https://github.com/inaka/elvis_core/pull/2/files#diff-54610b3451069d53e541dd9123ee5fd4R89) and friends in `elvis_core`.

My point is that `elvis_core` should **only** provide **core** functionality. How can we defined what that core functionality is? 

There are some modules that: 

1. could be brought from `elvis` **without modification**,
2. others that might need **minor changes** and
3. a few could use a **rewrite** to expose a better agnostic API.

I will list the modules I think belong to each group so we can start a discussion: 

#### 1. Without modification

- `elvis_code`
- `elvis_config`
- `elvis_file`
- `elvis_style`

#### 2. Minor changes

- `elvis_utils`: we could remove all of the function that handle the output since that is not really a core feature.
- `elvis_result`: the `print/1` function could be implemented by the user of `elvis_core`. We may have to modify the way we expose the types and their structure in this module.
- `elvis_project` (only because it has a couple of calls to `elvis_utils:error_prn/2`)

#### 3. Rewrite

- `elvis` (which I think should be renamed to `elvis_core`): maybe the module doesn't need a *full* rewrite, but it would be a good idea to remove the functions where no `Config` is provided so that `elvis_core` clients explicitly use `elvis_config:default/0`. It would also be a good idea to have an additional `Options` argument in all API functions. A possible option would be a callback function used to report partial results or logging messages, which could be disabled by not providing a value for such options.

We can merge this PR and work from there. Or we could discuss what would be the roadmap for `elvis_core` and create issues to organize work. Personally I prefer the second option.

One last detail, the file `test/examples/pull_request.js` is used for testing the webhook functionality, so it shouldn't be included in this project.

@brujo what do you think?

elbrujohalcon:
@jfacorro I think we can merge this PR, add your comments in an issue (or more than one) and start from here.

elbrujohalcon:
@jfacorro and I think `elvis` module should be renamed `elvis_core`, and an `elvis` module should be created in the elvis repo, which will use `elvis_core` for the core functionality and expose the extra stuff as it is now.

jfacorro:
> and I think elvis module should be renamed elvis_core, and an elvis module should be created in the elvis repo, which will use elvis_core for the core functionality and expose the extra stuff as it is now.

@elbrujohalcon Yeah, I included that in the comments.

jfacorro:
@Licenser To sum up. Please remove the file `test/examples/pull_request.js` and rename the `elvis` module to `elvis_core`. Thanks!

src/elvis_core.erl
```diff
@@ -0,0 +1,176 @@
...
+    Results = elvis_result:new(file, File, RulesResults),
+    elvis_result:print(Results),
+    Results.
+
+apply_rule({Module, Function}, {Result, Config, File}) ->
+    apply_rule({Module, Function, #{}}, {Result, Config, File});
+apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
+    ConfigMap = ensure_config_map(Module, Function, ConfigArgs),
+    RuleResult = try
+                    Results = Module:Function(Config, File, ConfigMap),
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 142. Only modules that define callbacks should make dynamic calls.

src/elvis_core.erl
```diff
@@ -0,0 +1,176 @@
...
+-spec rock_this(target()) ->
+    ok | {fail, elvis_result:file()}.
+rock_this(Target) ->
+    Config = elvis_config:default(),
+    rock_this(Target, Config).
+
+-spec rock_this(target(), elvis_config:config()) ->
+    ok | {fail, elvis_result:file()}.
+rock_this(Module, Config) when is_atom(Module) ->
+    ModuleInfo = Module:module_info(compile),
```
gadgetci: According to **Elvis**:
> Remove the dynamic function call on line 52. Only modules that define callbacks should make dynamic calls.

Licenser:
hmm the elvis gadget check isn't finishing o.O

jfacorro:
@Licenser There seems to be a problem with gadget when using elvis in this project. I'll check the logs to find out what is the problem and let you know.

jfacorro:
@Licenser Merged!

PR #6, opened 2015-11-26
Add group level ignore

### LINKED ISSUE #5: add option for global ignore
I'm having the situation that I want to exclude some modules entirely. It are auto generated modules for a lexer and parser and enforcing syntax rules on they would be pointless.

I would like to add a `ignore` rule on the top level, along with `dirs` and `filter` that entirely takes modules out of the check.

If there are no objections to this I'll implement it and open a PR.

Licenser:
This commit intorduces a `ignore` clause on the same level as
the `dirs` or `filter` clause, it can take both atoms or regex's.
Atoms are turned into a regex by taking advantage of erlangs
convention that module name and file name have to be the same.

This resolves #5.


jfacorro:
@Licenser Sorry it took so long to review this PR :disappointed:. I though that by only deleting the `rebar`it wouldn't conflict with this PR but it did. Could you update the branch so I can merge it? Thanks!

PR #18, opened 2016-01-08
[Fix #166] Extend default configuration

### LINKED ISSUE #166: smarter handling of DRY checks
the Dry-ing should probably not suggest to turn a flat list/tuple creation into a more dry version. That would only be mis helpful. Perhaps there could be some added value to complexity like depth?

``` erlang
start_link(Server, Port, Command, From, Timeout) ->
    gen_fsm:start_link(?MODULE,
                       [Server, Port, Command, From, Timeout], []).
%% and
call(Server, Port, Command, From, Timeout) ->
    supervisor:start_child(libchunter_fsm_sup,
                           [Server, Port, Command, From, Timeout]).
```

trigger the DRY detector but it would make little sense to extract `[Server, Port, Command, From, Timeout]` into an own function, it would neither make the code more readable nor more maintainable.

Upping complexity would prevent this but then other more valid claims are perhaps ignored.

harenson:


Makefile
```diff
@@ -10,8 +10,16 @@ TEST_DEPS = meck
 
 dep_meck = git https://github.com/eproxus/meck 0.8.3
 
+SHELL_DEPS = sync goldrush
```
elbrujohalcon: instead of adding goldrush as a dep, use lager not from hex but from github

Makefile
```diff
@@ -10,8 +10,16 @@ TEST_DEPS = meck
...
+
+dep_sync = git https://github.com/rustyio/sync.git 9c78e7b
+dep_goldrush = git https://github.com/DeadZen/goldrush.git 0.1.8
```
elbrujohalcon: instead of adding goldrush as a dep, use lager not from hex but from github

elbrujohalcon:
@harenson I finally found the proper word. It's not `rule_group`, it's [ruleset](http://www.linguee.es/ingles-espanol/traduccion/ruleset.html). Please rename everything.

src/elvis_config.erl
```diff
@@ -183,3 +187,43 @@ ignore_to_regexp(R) when is_list(R) ->
...
+%% @doc Merge user rules (override) with elvis default rules.
+-spec merge_rules(UserRules::list(), DefaultRules::list()) -> list().
+merge_rules(UserRules, DefaultRules) ->
+    UnduplicatedRules =
+        % Drops repeated rules
+        lists:filter(
+            % If any default rule is in UserRules it means the user
+            % wants to override the rule.
+            fun({FileName, RuleName}) ->
+                true /= is_rule_override(FileName, RuleName, UserRules);
```
elbrujohalcon: instead of `true /= |predicate|`, please use `not |predicate|` or just change `is_rule_override` to `should_use_default` and return the opposite value from it.

config/elvis.config
```diff
@@ -5,52 +5,19 @@
...
-                  {elvis_style, no_debug_call, #{ignore => [elvis,
-                                                            elvis_utils]}}
-                 ]
+        rule_group => erl_files
        },
       #{dirs => ["."],
         filter => "Makefile",
-        rules => [{elvis_project, no_deps_master_erlang_mk, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_erlang_mk, #{ignore => []}}]
+        rule_group => makefiles
```
elbrujohalcon: ruleset

config/elvis.config
```diff
@@ -5,52 +5,19 @@
...
        },
       #{dirs => ["."],
         filter => "rebar.config",
-        rules => [{elvis_project, no_deps_master_rebar, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_rebar, #{ignore => []}}]
+        rule_group => rebar_config
```
elbrujohalcon: ruleset

config/elvis.config
```diff
@@ -5,52 +5,19 @@
...
        },
       #{dirs => ["."],
         filter => "elvis.config",
-        rules => [{elvis_project, old_configuration_format}]
+        rule_group => elvis_config
```
elbrujohalcon: ruleset

src/elvis_config.erl
```diff
@@ -126,8 +126,12 @@ files(#{}) ->
     undefined.
 
 -spec rules(config()) -> [string()] | undefined.
-rules(_RuleGroup = #{rules := Rules}) ->
-    Rules;
+rules(_RuleSet = #{rules := UserRules, ruleset := RuleSet}) ->
```
elbrujohalcon: no need for the `_RuleSet` variable here

src/elvis_config.erl
```diff
@@ -126,8 +126,12 @@ files(#{}) ->
...
+    DefaultRules = elvis_rulesets:rules(RuleSet),
+    merge_rules(UserRules, DefaultRules);
+rules(_RuleSet = #{rules := Rules}) -> Rules;
```
elbrujohalcon: no need for the `_RuleSet` variable here

src/elvis_config.erl
```diff
@@ -126,8 +126,12 @@ files(#{}) ->
...
+rules(_RuleSet = #{ruleset := RuleSet}) ->
```
elbrujohalcon: no need for the `_RuleSet` variable here

src/elvis_config.erl
```diff
@@ -183,3 +187,43 @@ ignore_to_regexp(R) when is_list(R) ->
...
+               (_) -> false
+            end,
+            DefaultRules
+        ),
+    OverrideRules =
+        % Remove the rules that the user wants to "disable" and after that,
+        % remains just the rules the user wants to override.
+        lists:filter(
+            fun({_FileName, _RuleName, OverrideOptions}) ->
+                disable /= OverrideOptions;
```
jfacorro: I usually use `=/=` because it's an exact comparison. In this case it doesn't matter though since we are comparing an atom to a value and the match can only be exact.

PR #29, opened 2016-01-20
support ignore option in nesting_level and dont_repeat_yourself

stolen:
With this patch user can ignore some modules in `nesting_level` and `dont_repeat_yourself` rules.


stolen:
Also ignore code is copy-pasted a lot. Maybe there is better place for it?

elbrujohalcon:
@stolen would you mind adding some tests for this?

elbrujohalcon:
and if you manage to refactor the code in order to avoid all that copy/pasting… that would be awesome!

stolen:
OK, I'll try to refactor it when I have time.

PR #33, opened 2016-01-27
[Fix #41] rebar3 support

Euen:
[Fix #41] rebar3 support


elbrujohalcon:
This fixed inaka/elvis#307

PR #47, opened 2016-04-07
[#371] [Fix inaka/elvis#371] Move deps to app.src

elbrujohalcon:
In this PR:
- `katana_code` added to app.src
- `katana` removed (it was only used for `ktn_maps` and `maps` already comes with that functionality now)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ DEP_PLUGINS = inaka_mk hexer_mk
 
 dep_lager       = git https://github.com/basho/lager.git         3.1.0
 dep_zipper      = hex 0.2.0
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -9,7 +9,7 @@ DEP_PLUGINS = inaka_mk hexer_mk
 dep_lager       = git https://github.com/basho/lager.git         3.1.0
 dep_zipper      = hex 0.2.0
 dep_katana      = git https://github.com/inaka/erlang-katana.git 0.2.23
```
harenson: You removed `katana` from `DEPS`, I think you should remove it from here too.

PR #50, opened 2016-04-28
[Fix inaka/elvis#362] Correctly distinguish macros from variables

elbrujohalcon:
[Fix inaka/elvis#362] Correctly distinguish macros from variables

### DIFF
--- a/test/examples/pass_variable_naming_convention.erl
+++ b/test/examples/pass_variable_naming_convention.erl
@@ -13,7 +13,7 @@ should_pass(Should, Pass, Way2Home, Fun1, Fun2) ->
     Should = "Should",
     Pass = "Pass",
     Way2Home = "Way to home",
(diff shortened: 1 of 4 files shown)


test/examples/pass_variable_naming_convention.erl
```diff
@@ -12,5 +13,10 @@ should_pass(Should, Pass, Way2Home, Fun1, Fun2) ->
     Should = "Should",
     Pass = "Pass",
     Way2Home = "Way to home",
-    Fun1 = Should ++ Pass,
+Fun1 = Should ++ Pass,
```
amilkr: did you remove the indentation here?

elbrujohalcon: Yeah… intentionally… because I wanted to test the case when the variable started in column 1.

PR #64, opened 2016-10-17
[Fix #407] Fix bug in max_function_length rule

harenson:


src/elvis_style.erl
```diff
@@ -607,9 +607,16 @@ no_debug_call(Config, Target, RuleConfig) ->
 node_line_limits(FunctionNode) ->
     Zipper = elvis_code:code_zipper(FunctionNode),
     LineFun = fun(N) -> {L, _} = ktn_code:attr(location, N), L end,
+    % The first number in `lineNums' list is the location of the first
+    % line of the function. That's why we use it for the `Min' value.
     LineNums = zipper:map(LineFun, Zipper),
```
gadgetci: According to **Dialyzer**:
> The call zipper:map(LineFun::fun((_) -> any()), Zipper::#{'info':=#{'lefts':=[], 'parent_info':='undefined', 'parent_node':='undefined', 'rights':=[]}, 'node':=_, 'spec':=#{'children':=_, 'is_branch':=_, 'make_node':=_}}) does not have an opaque term of type zipper:zipper(_) as 2nd argument

PR #65, opened 2016-10-25
Remove Lager as a direct dependency

waisbrot:
Change the sole call to `lager:info` into `elvis_utils:info`

This is the elvis_core parallel to inaka/elvis#411

### DIFF
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -182,7 +182,7 @@ print_node(Node = #{type := Type}, CurrentLevel) ->
     Indentation = lists:duplicate(CurrentLevel * 4, $ ),
     Content = ktn_code:content(Node),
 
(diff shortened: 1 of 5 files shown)


src/elvis_code.erl
```diff
@@ -182,7 +182,7 @@ print_node(Node = #{type := Type}, CurrentLevel) ->
     Indentation = lists:duplicate(CurrentLevel * 4, $ ),
     Content = ktn_code:content(Node),
 
-    ok = lager:info("~s - [~p] ~p~n", [Indentation, CurrentLevel, Type]),
+    ok = elvis_utils:print("~s - [~p] ~p~n", [Indentation, CurrentLevel, Type]),
```
gadgetci: According to **Dialyzer**:
> Call to missing or unexported function elvis_utils:print/2

PR #66, opened 2016-10-27
Add rule to disallow using seqbind

gnuvince:
seqbind [1] is a parse transform for Erlang that allows the programmer
to assign to "the same variable" multiples times, e.g.,

```erlang
X@ = 10,
X@ = X@ + 1.
```

Although this transform can be useful at times, even its author
recommends using it sparingly since it obscures the functional nature of
Erlang, and can look quite ugly.

This commit adds an elvis_style rule to detect when a module declares
its intention to use seqbind.

[1] https://github.com/spawngrid/seqbind

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -23,7 +23,8 @@
          max_module_length/3,
          max_function_length/3,
          no_debug_call/3,
-         no_nested_try_catch/3
+         no_nested_try_catch/3,
+         no_seqbind/3
         ]).
 
 -define(LINE_LENGTH_MSG, "Line ~p is too long: ~s.").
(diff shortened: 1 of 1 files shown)


src/elvis_style.erl
```diff
@@ -1182,3 +1186,25 @@ check_nested_try_catchs(ResultFun, TryExp) ->
                              false
                     end,
                     elvis_code:find(Predicate, TryExp)).
+
+
+
+%% Disallow `-compile({parse_trans, seqbind})`
+no_seqbind(Config, Target, _RuleConfig) ->
```
gadgetci: According to **Compiler**:
> Warning: missing specification for function no_seqbind/3

gnuvince:
Selected the wrong base repo, not ready to be considered nor merged.

PR #67, opened 2016-10-31
Rules for warning about seqbind

gnuvince:
seqbind [1] is a parse transform for Erlang that allows the programmer to assign to "the same variable" multiples times, e.g.,

X@ = 10,
X@ = X@ + 1.

Although this transform can be useful at times, even its author recommends using it sparingly since it obscures the functional nature of Erlang, and can look quite ugly.

This PR adds:

    an elvis_style rule warning the user when a module declares its intention to use seqbind;
    an elvis_style rule warning the user when a module declares the seqbind parse transform, but never actually uses it (i.e., complete absence of Var@ in the module).

Issue ref.: https://github.com/inaka/elvis/issues/413

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -25,6 +25,7 @@
   {test, [
     {deps, [ {katana_test, {git, "https://github.com/inaka/katana-test.git", {tag, "0.0.6"}}}
            , {mixer,       {git, "https://github.com/inaka/mixer.git",       {tag, "0.1.5"}}}
+           , {seqbind,     {git, "https://github.com/lpgauth/seqbind.git",   {branch, "master"}}}
            , {meck,       "0.8.3"}
     ]}
   ]},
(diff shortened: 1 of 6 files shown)


rebar.config
```diff
@@ -25,6 +25,7 @@
   {test, [
     {deps, [ {katana_test, {git, "https://github.com/inaka/katana-test.git", {tag, "0.0.6"}}}
            , {mixer,       {git, "https://github.com/inaka/mixer.git",       {tag, "0.1.5"}}}
+           , {seqbind,     {git, "https://github.com/lpgauth/seqbind.git",   {branch, "master"}}}
```
elbrujohalcon: Don't use "master" branch, either choose a tag or use a commit SHA.

src/elvis_style.erl
```diff
@@ -1182,3 +1190,61 @@ check_nested_try_catchs(ResultFun, TryExp) ->
...
+
+
+uses_seq_bindings(Root) ->
+    SeqBindings = elvis_code:find(
+                    fun(Node) ->
+                            ktn_code:type(Node) =:= var andalso
+                            lists:last(ktn_code:attr(text, Node)) =:= $@
+                    end,
+                    Root),
+    case SeqBindings of
```
elbrujohalcon: Instead of a case, you can do `SeqBindings /= []` directly.

PR #72, opened 2017-02-01
Fix unnecessary loading of files caused by git-branch option (#421)

onno-vos-dev:
https://github.com/inaka/elvis/pull/421 introduced a git-branch option to elvis but this introduced a side-effect that files that are not checked by Elvis are still loaded and parsed causing unnecessary long run times.

### DIFF
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -69,11 +69,15 @@ rock_this(Path, Config) ->
                         [] =/= elvis_file:filter_files([File], Dirs, Filter,
                                                        IgnoreList)
                 end,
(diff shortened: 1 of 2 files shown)


anha0825:
Is there a way to rerun gadget/dialyzer? From what I understand there seems to have been a network / hex.pm problem.

elbrujohalcon:
Yes, close and reopen the issue… I'm doing that right now.

anha0825:
Thanks @elbrujohalcon. How do we get this into elvis? Are we able to create a new package out of elvis_core and point to that in elvis rebar.config? I guess we should have bumped the app version in elvis.app.src in this PR, or?

PR #81, opened 2017-03-07
Fix broken support for dashes in operator_spaces rule

tjarvstrand:


tjarvstrand:
I haven't really used gadget before but it looks like the failing compiler check is due to an environmental issue. Is it possible to rerun it?

test/style_SUITE.erl
```diff
@@ -244,6 +244,9 @@ verify_operator_spaces(_Config) ->
     SumOperation = #{rules => [{right, "+"}, {left, "+"}]},
     [_, _] = elvis_style:operator_spaces(ElvisConfig, Path, SumOperation),
 
+    MinusOperation = #{rules => [{right, "-"}, {left, "-"}]},
+    [] = elvis_style:operator_spaces(ElvisConfig, Path, MinusOperation),
+
```
tjarvstrand: Are you sure that's needed? The thing with dashes is that they are tokenized separately even if they are part of a compiler directive. This test shows that elvis doesn't choke on the module declaration (```-module(fail_operator_spaces).```) in ```fail_operator_spaces.erl``` when a rule with minus is applied

ferigis:
thanks @tjarvstrand , it is merged!

PR #83, opened 2017-04-14
Handle Hex deps with package name

varnerac:
Adds support for hex dependencies that specify a package name, like:

`{uuid, "1.7.0", {pkg, uuid_erl}}`


elbrujohalcon:
Thanks @varnerac !!
Would you mind adding test cases and (if applicable) doc for this change?

PR #88, opened 2017-06-06
add verbosity config option, output only errors unless set

srenatus:
`elvis_utils:print_info` will be suppressed when `verbose` is _not_ set. `elvis_utils:error` can only be suppressed by setting `no_output`.

### DIFF
--- a/src/elvis_utils.erl
+++ b/src/elvis_utils.erl
@@ -1,5 +1,7 @@
 -module(elvis_utils).
 
+-compile({no_auto_import, [error/2]}).
+
 -export([
          %% Rules
          check_lines/3,
@@ -16,6 +18,8 @@
          info/2,
          notice/1,
          notice/2,
+         error/1,
+         error/2,
          error_prn/1,
          error_prn/2,
          parse_colors/1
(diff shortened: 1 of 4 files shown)


src/elvis_utils.erl
```diff
@@ -138,8 +140,16 @@ notice(Message) ->
 -spec notice(string(), [term()]) -> ok.
 notice(Message, Args) ->
     ColoredMessage = "{{white-bold}}" ++ Message ++ "{{reset}}~n",
-    print(ColoredMessage, Args).
+    print_info(ColoredMessage, Args).
 
+-spec error(string()) -> ok.
+error(Message) ->
+    error(Message, []).
```
gadgetci: According to **Compiler**:
> Warning: ambiguous call of overridden auto-imported BIF error/2

srenatus:
@elbrujohalcon I've actually wondered how this _didn't_ break tests... I'm looking into adding some :+1:

elbrujohalcon:
@srenatus that's probably because we lost 100% coverage when we split elvis and elvis_core =/

elbrujohalcon:
Thanks @srenatus !!

srenatus:
@elbrujohalcon my pleasure -- thanks for being open to suggestions even when they change the defaults 😃

PR #93, opened 2017-07-27
[Fix #468] Add include_dirs/1 to elvis_config

harenson:
Also:
- Fix call to ktn_code:parse_tree/2
- [Fix inaka/elvis/issues/357] Document group and check level ignore

### DIFF
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -8,6 +8,7 @@
          normalize/1,
          %% Geters
          dirs/1,
+         include_dirs/1,
          ignore/1,
          filter/1,
          files/1,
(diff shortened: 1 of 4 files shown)


src/elvis_config.erl
```diff
@@ -116,6 +118,17 @@ dirs(_RuleGroup = #{dirs := Dirs}) ->
 dirs(#{}) ->
     [].
 
+% Only get `include_dirs' value for erl files RuleGroup, discard other ones.
+-spec include_dirs(Config::config() | map()) -> [string()].
+include_dirs(Config) when is_list(Config) ->
+    lists:flatmap(fun include_dirs/1, Config);
+include_dirs(RuleGroup = #{ruleset := "erl_files"}) when is_map(RuleGroup) ->
```
elbrujohalcon: Don't be so strict, let every group have an `include_dirs` option, even when it might be useless today.

src/elvis_config.erl
```diff
@@ -116,6 +118,17 @@ dirs(_RuleGroup = #{dirs := Dirs}) ->
...
+    maps:get(include_dirs, RuleGroup, ?DEFAULT_INCLUDE_DIRS);
+include_dirs(RuleGroup = #{filter := "*.erl"}) when is_map(RuleGroup) ->
+    maps:get(include_dirs, RuleGroup, ?DEFAULT_INCLUDE_DIRS);
+include_dirs(_) ->
```
elbrujohalcon: The default should not be `[]`. It should be `?DEFAULT_INCLUDE_DIRS` instead.

PR #97, opened 2018-04-13
Support ignore for ignored variable used

cypherfox:
This change adds the ability to add a list of ignored modules to the rule ignored_variable_used.

This is needed as elvis will fail on an ignored variable in  a eunit ?assertMatch macro.

```
    ?assertMatch({ok, {a_valid_message, _V}}, fut(Args))
```
Not ignoring the variable will cause the compiler to report a warning instead.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -364,20 +364,20 @@ invalid_dynamic_call(Config, Target, RuleConfig) ->
                             elvis_file:file(),
                             used_ignored_variable_config()) ->
     [elvis_result:item()].
-used_ignored_variable(Config, Target, _RuleConfig) ->
+used_ignored_variable(Config, Target, RuleConfig) ->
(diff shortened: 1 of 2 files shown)


elbrujohalcon:
I think the idea is useful, but what if you _fully_ ignore the variable (i.e. using `_`)?

elbrujohalcon:
Would you mind adding / adjusting some tests for this functionality?

cypherfox:
Ah! Fully ignoring works. While I dislike it from a standpoint of literate programming (IMHO it reduces the readability of the test), that explains why nobody else has been asking for it for a long time.

Yes, I will try to add some tests.

elbrujohalcon:
Thanks @cypherfox ! Don't forget to also update [the wiki](https://github.com/inaka/elvis_core/wiki/Rules#used-ignored-variable) with this change

cypherfox:
@elbrujohalcon done.

PR #98, opened 2018-06-28
Support for no_call and no_common_caveats

adrianroe:
A discussed in https://github.com/inaka/elvis/issues/484

I haven't update the docs as yet, but it all has tests and is dialyzer clean.

There is one issue at present, which is that the existing elvis_core code does not recognise calls to size as actually being calls to erlang:size so the test file, somewhat artificially has calls in it to erlang:size (whereas 99% of my code certainly does not usually use the erlang: prefix for this).

Perhaps one to talk through.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -22,7 +22,9 @@
          dont_repeat_yourself/3,
          max_module_length/3,
          max_function_length/3,
+         no_call/3,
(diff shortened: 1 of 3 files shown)


elbrujohalcon:
@adrianroe I'm merging the PR… please remember to update the docs.

PR #106, opened 2019-04-10
Add parsable option in order to output results in the dialyzer-like format

define-null:
I added another way to format the output, by providing the parsable env variable, for producing the output, that reminds dialyzer output. (see rationale https://github.com/inaka/elvis/issues/497)

Example of the output:
```
src/example.erl:1:god_modules:This module has too many functions (56). Consider breaking it into a number of modules.
src/example_a.erl:341:no_debug_call:Remove the debug call to io:format/2 on line 341.
src/example_a.erl:511:used_ignored_variable:Ignored variable is being used on line 511 and column 54.
src/example_a.erl:1252:used_ignored_variable:Ignored variable is being used on line 1252 and column 21.
src/example_a.erl:1259:used_ignored_variable:Ignored variable is being used on line 1259 and column 21.
src/example_a.erl:1309:used_ignored_variable:Ignored variable is being used on line 1309 and column 17.
src/example_a.erl:1311:used_ignored_variable:Ignored variable is being used on line 1311 and column 17.
src/other.erl:1:god_modules:This module has too many functions (73). Consider breaking it into a number of modules.
src/other.erl:150:no_spec_with_records:The spec in line 150 uses a record, please define a type for the record and use that instead.

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,17 @@ list of items with the following structure `{Module, Function, RuleConfig}` or
 `disable` certain rules if you want to just by specifying the rule in the `rules`
 key and passing `disable` as its third parameter.
 
+`output_format` is used to configure the output format. Possible values are `colors`,
+`plain` and `parsable`. The latter could be use for the automated parsing and has a
+format very close to dialyzer in a form `FILE:LINE:RULE:MESSAGE`:
+
+```
+src/example.erl:1:god_modules:This module has too many functions (56). Consider breaking it into a number of modules.
+src/example_a.erl:341:no_debug_call:Remove the debug call to io:format/2 on line 341.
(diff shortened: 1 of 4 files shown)


define-null:
Squashed commits, pls let me know what you think.

elbrujohalcon:
Looks reasonable to me. Do you think there is any way to add a test for this? At least to test that elvis does work when the new parameter is on?

elbrujohalcon:
Actually… Check out [my comment](https://github.com/inaka/elvis/issues/497#issuecomment-481648338) on inaka/elvis#497

elbrujohalcon:
Oh, and fix the dialyzer warnings reported by travis :)

define-null:
Ok, will do that.

define-null:
Let me know if there are any other changes needed. Otherwise I  will squash it before this could be merged

PR #107, opened 2019-04-22
Rock in parallel

define-null:
Implement parallel processing of the files. Option 'parallel'
takes pos integer as a number of processes allowed to run in
parallel, default to 1, to keep current behaviour.
    
Drop most of the file section from the result, and split loading
phase to reduce memory consumption and speed up gathering of the
results.

Some results, based on the https://github.com/basho/riak_core.git
Config:
```
$ cat elvis.config 
[
 {
   elvis,
   [
    {config,
     [#{dirs => ["src"],
        include_dirs => ["include"],
        filter => "*.erl",
        ruleset => erl_files
       }
     ]
    },
    %% Optional to select the output format, the default is colors
    {output_format, parsable}
   ]
 }
].
```
1) Before the change (current master):
```
riak_core$ time ../elvis/_build/default/bin/elvis rock > elvis.log

real	0m30.834s
user	0m30.711s
sys	0m1.371s
```
2) After the change, single process version:
```
riak_core$ time ../elvis/_build/default/bin/elvis rock > elvis.log

real	0m10.907s
user	0m11.523s
sys	0m0.257s
```
3) After the change, auto option (4 procs)
```
riak_core$ time ../elvis/_build/default/bin/elvis -p auto rock > elvis.log

real	0m5.933s
user	0m20.865s
sys	0m0.321s
```

### DIFF
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -37,7 +37,7 @@ rock() ->
 rock(Config) ->
     ok = elvis_config:validate(Config),
     NewConfig = elvis_config:normalize(Config),
-    Results = lists:map(fun do_rock/1, NewConfig),
+    Results = lists:map(fun do_parallel_rock/1, NewConfig),
     lists:foldl(fun combine_results/2, ok, Results).
 
 -spec rock_this(target()) ->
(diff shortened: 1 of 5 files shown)


define-null:
related: https://github.com/inaka/elvis/issues/382

define-null:
@elbrujohalcon please have a look

src/elvis_core.erl
```diff
@@ -74,23 +74,88 @@ rock_this(Path, Config) ->
...
+do_parallel_rock(Config0) ->
+    Parallel = application:get_env(elvis, parallel, 1),
     Config = elvis_config:resolve_files(Config0),
     Files = elvis_config:files(Config),
-    Fun = fun (File) -> load_file_data(Config, File) end,
-    LoadedFiles = lists:map(Fun, Files),
-    elvis_utils:info("Applying rules..."),
-    Results = [apply_rules(Config, File) || File <- LoadedFiles],
+
+    Results = do_parallel_rock0(Config, Files, Parallel),
```
elbrujohalcon: @define-null what if instead of implementing all of this logic ourselves, we just use [`rpc:pmap/4`](http://erlang.org/doc/man/rpc.html#pmap-3) here?
Would something like the following do the trick?
```erlang
  …
  elvis_utils:info("Applying rules..."),
  rpc:pmap({?MODULE, rock_file}, [Config], Files).

rock_file(Config, File) ->
  FileSrc = load_data_file(Config, File),
  apply_rules(Config, File).
```

define-null: I thought about that originally, but the problem with pmap, is that it does not allow to set the max amount of concurrency, that might be a problem when for instance you want to run elvis from the makefile that has -j option or on the CI, that does not limit amount of resources for particular job. Giving the ability to specify `-p auto`, or `-p N` solves both needs that I had in mind - to be able to limit concurrency, and an abillity to simply utilize all resources, available.

 Also if we just run pmap + lists:split - pmap would have to wait till all the processes are finished, before processing the next chunk, which would also not very beneficial for utilizing resources.

elbrujohalcon: I see… so you do want to limit concurrency. I originally thought "why would we want to be less concurrent than possible?" :P
Clearly, I don't run elvis on the same scenarios as you do :)
OK. I'll review this again either today or tomorrow and let you know my thoughts. I'm generally not in favor of having message passing, receiving, etc… code mixed in with the code that actually _does the job_ (that's why I prefer using high-order functions like `rpc:pmap/4`).
Maybe we can use another library that provides chunk-processing (I believe @ferd had one) or add some `ktn_task:chunk_map(…)` but only if it actually makes sense.
Give me a day or two and I'll do my research.

define-null: Lack of concurrency limit here, would also lead to huge memory consumption, when all the files are loaded, and syntax tree is generated for all of them and just reside in memory. With current implementation when the user does not have much memory with the large project he can just use `-p 1`.

In my case I have to deal with pretty large code base, for which the overall time is several orders magnitude higher then the timing, presented for riak_core, so I was thinking of optimizing both cpu/memory consumption if needed.

define-null: Yeah sure, no problem =) I will check ktn_task then.
I'm also not a fan of re-implementing common stuff, so fully understand your point about rpc:pmap.

igaray: I like this work and agree with @elbrujohalcon that separating elvis logic from a generic concurrency construction is desirable. ktn_task seems like a good destination for such a construction if it doesn't have it already.

define-null: Uhh, unfortunately elvis_core still relies on katana_code, "0.1.2", while the current version is 0.4.0, where kth_code does no longer exists. I'm not aware if there exists other non-incompatibilities. Also noticed that katana_code depend on elvis, which creates circle dependency.

I created separate fun, as you requested: https://github.com/define-null/erlang-katana/blob/feature/chunk_fold/src/ktn_task.erl#L54 but it seems in order to patch the katana version more non-related changes would be needed.

What about moving the code that I provided to the separate model under the elvis_code umbrella?

elbrujohalcon: Wait a sec.
Elvis depends on `katana-code` `0.1.2`, which is the latest version of that package: https://hex.pm/packages/katana_code
The one that's on `0.4.0` is `erlang-katana`, and yes… `ktn_code` is no longer there. It was moved to `katana-code` to avoid a cicular depend… OOOOOOH I see what you mean!
Yes, moving the code that you provided to something like `elvis_task` and adding some tests to it would be lovely. We can copy that code to `ktn_task` later.

src/elvis_task.erl
```diff
@@ -0,0 +1,115 @@
...
+%% every element Elem in JobItemList in parallel with max concurrcy factor
+%% equal to Concurrency. On succesfull evaluation FunAcc function is called
+%% with the result of succesfull execution as a first parametr and accumulator
+%% as a second parametr.
+-spec chunk_fold(FunWork :: {Module :: module(), Function :: atom()},
+                 FunAcc :: fun((NewElem :: term(), Acc :: term()) ->
+                                      Acc :: term()),
+                 InitialAcc :: term(),
+                 ExtraArgs :: list(),
+                 JonItemList :: list(),
```
igaray: did you mean JoinItemList ?

define-null: Yes ^^

test/elvis_SUITE.erl
```diff
@@ -422,6 +425,28 @@ to_string(_Config) ->
...
+
+    {error, {error, undef}} =
+        elvis_task:chunk_fold({?MODULE, chunk_fold_task_do_not_exist},
+                              fun(Elem, Acc) ->
+                                      {ok, Acc + Elem}
+                              end, 0, [Multiplier], lists:seq(1,10), 10).
+
+
+-spec chunk_fold_task(integer(), integer()) -> {ok, integer()}.
+chunk_fold_task(Elem, Miltiplier) ->
```
igaray: Multiplier*

igaray:
This LGTM, outstanding work! Pending @elbrujohalcon 's approval I want this in.

elbrujohalcon:
Let's wait on Travis… and we're ready to merge!

define-null:
Should be fine now. :)

elbrujohalcon:
It's merged. Now you will need a new release of elvis_core in hex.pm, right @define-null ?

PR #108, opened 2019-05-13
[#500] analyze abstract code if beam_files section is added

tothlac:


src/elvis_rulesets.erl
```diff
@@ -30,13 +30,18 @@ rules(erl_files) ->
...
   , {elvis_style, dont_repeat_yourself, #{min_complexity => 10}}
-  , {elvis_style, no_debug_call, #{ignore => [elvis, elvis_utils]}}
   , { elvis_style
     , variable_naming_convention
     , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$"}
     }
   , {elvis_style, no_nested_try_catch}
   ];
+rules(beam_files) ->
+  [
```
elbrujohalcon: To follow ROK-style / comma-first indentation, you should rewrite this as…
```erlang
  [ {elvis_style, no_debug_call, #{ ignore => [elvis, elvis_utils]
                                  , use_abstract_code => true
                                  }}
  , {elvis_style, invalid_dynamic_call, #{use_abstract_code => true}}
  ];
```

src/elvis_style.erl
```diff
@@ -634,7 +633,7 @@ no_call(Config, Target, RuleConfig) ->
 -type no_debug_call_config() :: #{debug_functions => [function_spec()],
                                   ignore => [module()]
                                  }.
--spec no_debug_call(elvis_config:config(),
+-spec no_debug_call(elvis_config:config() | map(),
```
elbrujohalcon: What's the situation in which this function is provided a `map()` that is _not_ an `elvis_config:config()`?

tothlac: If I debug I see it is always with a map, not with a list of maps as it is specified in `elvis_config:config()`. 

If I'm not wrong all the Config passed to the rules is always a map, not a list of maps. If I debug it I see.
```
elvis_style 642 Config: '#{,
                           dirs => ["_build/default/lib/myapp/ebin"],
                           files =>
                               [#{path =>
                                      "_build/default/lib/myapp/ebin/dapp.beam"}],
                           filter => "*.beam",
                           rules =>
                               [{elvis_style,no_debug_call,#{}},
                                {elvis_style,invalid_dynamic_call,
                                    #{ignore => [dapp_rest_handler]}}],
                           ruleset => beam_files}
```

Actually I don't understand why elvis_config:config is defined like this:
```
-type config() :: [map()].
```
Should not it be simply a map?

src/elvis_rulesets.erl
tothlac: Fixed this

src/elvis_style.erl
elbrujohalcon: Maybe? Try replacing that type definition and running dialyzer on the code. If a list is required somewhere, dialyzer will tell you :)

tothlac: This is actually why I added map(), because dialyzer recognized it will be called not with [map()], but with map().

elbrujohalcon: Cool! Let's change it, then!

src/elvis_rulesets.erl
tothlac: Do you have any ideas how this can be tested?

elbrujohalcon: @igaray can you help us out here?

paulo-ferraz-oliveira:
I think this is a great addition and would surely unblock stuff we'd want to do in the future. What's missing? @elbrujohalcon, @tothlac? (I can help working on it, if I can find some spare time)

src/elvis_style.erl
elbrujohalcon: @paulo-ferraz-oliveira this is one thing that's missing.

elbrujohalcon:
@paulo-ferraz-oliveira I left some comments that are not yet addressed in the code… and also the PR is missing some tests and a merge from `master`.
I believe you can totally take over the work from @tothlac … it shouldn't be hard to get it to a mergeable point.

paulo-ferraz-oliveira:
> @paulo-ferraz-oliveira I left some comments that are not yet addressed in the code… and also the PR is missing some tests and a merge from `master`.
> I believe you can totally take over the work from @tothlac … it shouldn't be hard to get it to a mergeable point.

Is it OK to just copy-paste the code to finish it? Aren't there any license related issues (@tothlac, do I have your permission to finish this implementation)? I don't mind attributing the "cherry-picked" commit to @tothlac and then continuing on top of that, since it'll be mostly his work, but it would be a shame to throw it out.

elbrujohalcon:
> > @paulo-ferraz-oliveira I left some comments that are not yet addressed in the code… and also the PR is missing some tests and a merge from `master`.
> > I believe you can totally take over the work from @tothlac … it shouldn't be hard to get it to a mergeable point.
> 
> Is it OK to just copy-paste the code to finish it? Aren't there any license related issues (@tothlac, do I have your permission to finish this implementation)? I don't mind attributing the "cherry-picked" commit to @tothlac and then continuing on top of that, since it'll be mostly his work, but it would be a shame to throw it out.

It's OK, @paulo-ferraz-oliveira … we'll give credit where credit is due.

src/elvis_style.erl
paulo-ferraz-oliveira: Yeah, but here `elvis_config:config()` really is a list of maps, so I guess the spec is right as-is.
```erlang
[ #{ dirs => ["src"]
   , ruleset => erl_files
  }
, #{ dirs => ["."]
   , ruleset => makefiles
  }
, ...
]
```

paulo-ferraz-oliveira:
I see `elvis_rulesets` with only two rules, but it's probably incomplete.

src/elvis_style.erl
```diff
@@ -1367,3 +1365,16 @@ uses_seq_bindings(Root) ->
                     end,
                     Root),
     SeqBindings /= [].
+
+-spec get_root(Config, Target) -> Res when
+      Config :: elvis_config:config() | map(),
+      Target :: elvis_file:file(),
+      Res :: ktn_code:tree_node().
+get_root(Config, Target) ->
```
paulo-ferraz-oliveira, line 1373: Does this seem to assume that the BEAM abstract tree and the `ktn` abstract tree are compatible?

paulo-ferraz-oliveira, line 1373: Does this assume that the BEAM abstract syntax tree and the `ktn` tree are "compatible"?

elbrujohalcon: It probably assumes that but that would be an incorrect assumption.

paulo-ferraz-oliveira:
Superseded by https://github.com/inaka/elvis_core/pull/138.

PR #110, opened 2019-07-23
treat FUNCTION_NAME as valid variable

aboroska:
Embedded macros might not expand the FUNCTION_NAME macro and it will be
treated as variable. Let's accept it.

Fixes inaka/elvis#505

### DIFF
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -33,7 +33,7 @@ rules(erl_files) ->
   , {elvis_style, no_debug_call, #{ignore => [elvis, elvis_utils]}}
   , { elvis_style
     , variable_naming_convention
-    , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$"}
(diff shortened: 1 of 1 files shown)


elbrujohalcon:
Even when this would certainly work, I would prefer a less _hacky_ solution. I will leave this in @igaray's hands, tho.

igaray:
@aboroska Thank you for taking the time to issue a PR, but although the end result is what is desired I really don't think this is the way forward, even temporarily.

The issue you linked remains and your helpful comment is present for others if they find themselves in the same situation. We will get around to addressing it, I think it's an important bug to fix.

What should really be done is investigate what happens during aleppo's parsing and see whether we can get it to expand all macros. I am going to close this PR but if you want to try a different solution we would be happy to help. 

Again, köszönöm!

aboroska:
@igaray you are fully right, I was about to to close it myself too. Cheers.

PR #112, opened 2019-11-25
[inaka/elvis#483] Operator spaces false positives

jfacorro:
Fixes inaka/elvis#483

### DIFF
--- a/test/examples/fail_operator_spaces.erl
+++ b/test/examples/fail_operator_spaces.erl
@@ -41,7 +41,8 @@ function5() ->
 function6() ->
     _MissingLeftSpace = 2+ 3,
     _MissingRightSpace = 2 +3,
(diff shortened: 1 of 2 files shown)


test/examples/fail_operator_spaces.erl
```diff
@@ -41,7 +41,8 @@ function5() ->
 function6() ->
     _MissingLeftSpace = 2+ 3,
     _MissingRightSpace = 2 +3,
-    _Successfull = 2 + 3.
+    _Successfull = 2 + 3,
```
elbrujohalcon: 🤦‍♂ 
We could use this chance to fix the typo ;)

jfacorro: We certainly can 😄. Fixed and pushed.

PR #114, opened 2019-12-04
Prepare release 0.4.3

jfacorro:
Let's create this release before bumping katana_code to not use the aleppo library.

### DIFF
--- a/src/elvis.app.src
+++ b/src/elvis.app.src
@@ -3,7 +3,7 @@
   [
    {pkg_name, elvis_core},
    {description, "Core library for the Erlang style reviewer"},
-   {vsn, "0.4.2"},
+   {vsn, git},
    {applications, [kernel, stdlib, zipper, katana_code]},
    {modules,
     [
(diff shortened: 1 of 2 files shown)


src/elvis.app.src
```diff
@@ -3,7 +3,7 @@
   [
    {pkg_name, elvis_core},
    {description, "Core library for the Erlang style reviewer"},
-   {vsn, "0.4.2"},
+   {vsn, git},
```
elbrujohalcon: Does this really work? At some point in the past we discarded this for… some reason I can't recall right now.
But… if we're going to use `git` for `vsn`s, we should change [our guidelines](https://github.com/inaka/guidelines/blob/master/ERLANG.md#steps-for-new-releases--version-bumps)

jfacorro: Yes, it works. I opened a [PR](https://github.com/tsloughter/rebar3_hex/pull/129) in `rebar3_hex` to fix it 🍾 .

elbrujohalcon: Lovely!! ❤️

PR #118, opened 2020-02-27
Allow {Module, Function} ignores in function_naming_convention/3

onno-vos-dev:
I'm slowly trying to modify as many rules as possible to accept a `{Module, Function}` or even `{Module, Function, Arity}` ignore. See: https://github.com/inaka/elvis/issues/488 for a discussion that I had with @elbrujohalcon regarding this. 

Rather than creating a big bang PR that fixes it in many places, I'd like to start small and fix one of the easiest ones to implement this for: function_naming_convention/3

I'm apparently able to update the Wiki (which seems a bit strange...?). Once approved and merged, I'll follow up with a PR in https://github.com/inaka/elvis/blob/master/rebar.config and edit the Wiki

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -147,8 +147,9 @@
                                       ignore => [atom()],
                                       max_length => integer()}.
 
+-type function_naming_convention_spec() :: module() | {module(), atom()}.
 -type function_naming_convention_config() :: #{regex => string(),
-                                               ignore => [module()]
+                                               ignore => [function_naming_convention_spec()]
                                               }.
 
 -spec function_naming_convention(elvis_config:config(),
@@ -157,27 +158,34 @@
     [elvis_result:item()].
 function_naming_convention(Config, Target, RuleConfig) ->
     Regex = maps:get(regex, RuleConfig, ".*"),
-    IgnoreModules = maps:get(ignore, RuleConfig, []),
-
+    Ignores = maps:get(ignore, RuleConfig, []),
     {Root, _} = elvis_file:parse_tree(Config, Target),
     ModuleName = elvis_code:module_name(Root),
-    case lists:member(ModuleName, IgnoreModules) of
-      false ->
-        FunctionNames = elvis_code:function_names(Root),
-        errors_for_function_names(Regex, FunctionNames);
-      true -> []
+    IgnoredFuns = lists:filtermap(
+                    fun({_ModuleName, Function}) -> {true, Function};
+                       (_) -> false
+                    end, Ignores),
+    case lists:member(ModuleName, Ignores) of
+      true -> [];
+        false ->
+        FunctionNames0 = elvis_code:function_names(Root),
+        FunctionNames = lists:filter(
+                          fun(FunctionNames) ->
+                                  not lists:member(FunctionNames, IgnoredFuns)
+                          end, FunctionNames0),
+        errors_for_function_names(Regex, FunctionNames)
     end.
 
 errors_for_function_names(_Regex, []) -> [];
-errors_for_function_names(Regex, [FunctionName | RemainingFuncNames]) ->
+errors_for_function_names(Regex, [FunctionName | Rem]) ->
     FunctionNameStr = atom_to_list(FunctionName),
     case re:run(FunctionNameStr, Regex) of
         nomatch ->
             Msg = ?FUNCTION_NAMING_CONVENTION_MSG,
             Info = [FunctionNameStr, Regex],
             Result = elvis_result:new(item, Msg, Info, 1),
-            [Result | errors_for_function_names(Regex, RemainingFuncNames)];
-        {match, _} -> errors_for_function_names(Regex, RemainingFuncNames)
+            [Result | errors_for_function_names(Regex, Rem)];
+        {match, _} -> errors_for_function_names(Regex, Rem)
     end.
 
 -type variable_naming_convention_config() :: #{regex => string(),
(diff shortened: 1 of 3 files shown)


src/elvis_style.erl
```diff
@@ -157,27 +158,34 @@
...
-        FunctionNames = elvis_code:function_names(Root),
-        errors_for_function_names(Regex, FunctionNames);
-      true -> []
+    IgnoredFuns = lists:filtermap(
+                    fun({_ModuleName, Function}) -> {true, Function};
+                       (_) -> false
+                    end, Ignores),
+    case lists:member(ModuleName, Ignores) of
+      true -> [];
+        false ->
```
elbrujohalcon: Indentation here is a bit ugly.

src/elvis_style.erl
```diff
@@ -157,27 +158,34 @@
...
+    Ignores = maps:get(ignore, RuleConfig, []),
     {Root, _} = elvis_file:parse_tree(Config, Target),
     ModuleName = elvis_code:module_name(Root),
-    case lists:member(ModuleName, IgnoreModules) of
-      false ->
-        FunctionNames = elvis_code:function_names(Root),
-        errors_for_function_names(Regex, FunctionNames);
-      true -> []
+    IgnoredFuns = lists:filtermap(
+                    fun({_ModuleName, Function}) -> {true, Function};
```
jfacorro: Isn't this just returning the whole list regardless of the module? If there are two modules with the same function name (e.g. `f:foo` and `g:foo`) and the ignore list has `[{g, foo}]`, won't this ignore the functions in modules `f` and `g`?

src/elvis_style.erl
```diff
@@ -157,27 +158,34 @@
...
+        FunctionNames = lists:filter(
+                          fun(FunctionNames) ->
+                                  not lists:member(FunctionNames, IgnoredFuns)
+                          end, FunctionNames0),
+        errors_for_function_names(Regex, FunctionNames)
     end.
 
 errors_for_function_names(_Regex, []) -> [];
-errors_for_function_names(Regex, [FunctionName | RemainingFuncNames]) ->
+errors_for_function_names(Regex, [FunctionName | Rem]) ->
```
jfacorro: Why this rename? 😜 Maybe we could leave a full word like `Remaining` instead of just `Rem`.

onno-vos-dev: Good catch, I initially passed an additional argument here which I later reverted. I'll revert this renaming

src/elvis_style.erl
onno-vos-dev: Good catch, I'll add a match for filtering out the functions for only this specific module

src/elvis_style.erl
onno-vos-dev: You're right. The default indentation for my Emacs is 2 spaces so I was fiddling around with getting the correct indentation to work :)

elbrujohalcon: We'll eventually add the `rebar3_format` plugin and happily run `rebar3 format` 🌟

onno-vos-dev: btw, that bad formatting was already there by the looks of it? (Unless I need more coffee and I'm blind...)

elbrujohalcon:
Don't forget to update [the wiki](https://github.com/inaka/elvis_core/wiki/Rules#function-naming-convention) accordingly.

onno-vos-dev:
I'll do :+1: There are some more things to fix there since some rules that allow ignores aren't listed there so I'll go through the thing properly :+1: (And bump elvis once elvis_core is tagged)

onno-vos-dev:
Could either of you create a new tag please?

onno-vos-dev:
Could either of you create a new tag please? 

Thanks for the merge!

jfacorro:
@onno-vos-dev https://github.com/inaka/elvis_core/pull/119

jfacorro:
@onno-vos-dev Release created [0.6.0](https://github.com/inaka/elvis_core/releases/tag/0.6.0).

onno-vos-dev:
Thank you @jfacorro I found a lot of inconsistencies in the Wiki. 

l'll try and do some form of PR by creating a Wiki on my github account and comment here. Am I just mis-reading things or could it be that the Wiki is (quite severally) outdated?

Line Length -> missing ignore
Nesting Level -> missing ignore
Module Naming Convention -> default regex seems wrong
Function Naming Convention -> missing ignore
Variable Naming Convention -> default regex seems wrong
Variable Naming Convention -> missing ignore
DRY -> missing ignore
Max Function Length -> IgnoreFunctions has wrong spec
No debug call -> default debug_functions is wrong
No deps master rebar -> rule doesn't seem to exist anymore?
Protocol for deps erlang.mk -> rule doesn't seem to exist anymore?
Protocol for deps rebar -> rule doesn't seem to exist anymore?
Old configuration format -> rule doesn't seem to exist anymore?
NO_SPACES_MSG -> not documented
no_seqbind-> not documented
no_useless_seqbind-> not documented

jfacorro:
The wiki page has been neglected for a long time. Maybe we should add the documentation for each rule as an edoc in each function of `elvis_style` which would at least be ketp in sync with the code.

onno-vos-dev:
I like that idea, alternatively, we make it part of the README.md. That way at least it's version controlled and easier to block a PR if docs aren't updated with it?

jfacorro:
Making it part of the `README.md` might make the file too big and discourage people to go through it. The reason why I was thinking about adding it to the edoc in each rule is because we could generate the edoc pages and publish them with each release.

onno-vos-dev:
Sounds good :+1: I'll move em over on the weekend and create a PR for it :+1:

PR #122, opened 2020-08-05
Allow use of rebar_raw_resource in rebar3 -oriented projects

paulo-ferraz-oliveira:
I have this entry in `rebar.config`'s `deps`:

```erlang
{dep, {raw, {git, "https://github.com/dep.git", {branch, "develop"}}}}
```

While analyzing `rebar.config`, `elvis` (via `elvis_core`) exits with:

```erlang
# rebar.config [FAIL]
Error: 'badarg' while applying rule 'protocol_for_deps_rebar'.
===> Linting failed
```

That entry respects the format of rebar3 plugin [rebar_raw_resource](https://github.com/basho/rebar_raw_resource).

### DIFF
--- a/src/elvis_project.erl
+++ b/src/elvis_project.erl
@@ -157,6 +157,8 @@ is_rebar_master_dep({_AppName, {_SCM, _Location, "master"}}) ->
     true;
 is_rebar_master_dep({_AppName, {_SCM, _Location, {branch, "master"}}}) ->
     true;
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
I might be missing a few changes... let me check.

paulo-ferraz-oliveira:
Many thanks.

PR #124, opened 2020-08-31
Prevent sub-folder files from being analyzed twice

paulo-ferraz-oliveira:
This is what is happening to me...
I have an `elvis.config` file such as
```erlang
[{elvis, [
    {config, [
        #{ dirs => ["src",
                    "src/sub"],
           filter => "*.erl",
           ruleset => erl_files
    ]},
    {verbose, true}
]}].
```
When analyzing file `file.erl` inside folder `src/sub` it gets analyzed twice, since `elvis_core`'s internals match it with both `src/` and `src/sub/`.
The "solution" found was to un-duplicate these occurrences.

### DIFF
--- a/src/elvis_file.erl
+++ b/src/elvis_file.erl
@@ -81,7 +81,7 @@ filter_files(Files, Dirs, Filter, IgnoreList) ->
     AppendFilter = fun(Dir) ->
                          case Dir of
                              "." -> Filter;
-                             Dir -> Dir ++ "/" ++ Filter
+                             Dir -> reduce_stars(Dir ++ "/" ++ Filter)
                          end
                    end,
     FullFilters = lists:map(AppendFilter, Dirs),
@@ -129,6 +129,9 @@ escape_all_chars(Glob) -> re:replace(Glob, ".", "[&]", [global]).
 replace_stars(Glob) -> re:replace(Glob, "[[][*][]]", ".*", [global]).
 
 replace_questions(Glob) -> re:replace(Glob, "[[][?][]]", ".", [global]).
+
+reduce_stars(DirAndFilter) ->
+    re:replace(DirAndFilter, "/\\*+/", "/", [global, {return, list}]).
 
 -spec find_encoding(Content::binary()) ->
   atom().
--- a/src/elvis_file.erl
+++ b/src/elvis_file.erl
@@ -103,7 +103,9 @@ filter_files(Files, Dirs, Filter, IgnoreList) ->
                 not lists:any(IsIgnored, IgnoreList)
         end,
     Found = lists:flatmap(FlatmapFun, Regexes),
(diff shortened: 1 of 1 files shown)


elbrujohalcon:
Does this fix inaka/elvis#508?

elbrujohalcon:
Related: https://github.com/inaka/elvis_core/commit/1d80b288763d0c5bb50e791ad511e04d34fe8eff#r36143723

elbrujohalcon (approved):

paulo-ferraz-oliveira:
Let me try to check that tonight. If it doesn't I might want to tackle that also.

paulo-ferraz-oliveira:
> Does this fix [inaka/elvis#508](https://github.com/inaka/elvis/issues/508)?

Let me try to check that tonight. If it doesn't I might want to tackle that also.

paulo-ferraz-oliveira:
Regarding `dirs`:
* `["src/**"]`: same result (i.e. not fixed),
* `["src", "src/**"]`: fixed,
* `["src", "src/*", "src/*/*"]`: fixed.

paulo-ferraz-oliveira:
I'm looking into why the issue occurs with `["src/**]`, but I need to understand initial requirements:

1. is declaring `["src"]` supposed to be recursive?
2. is declaring `["src/**"]` supposed to be recursive?

I feel there's a simpler approach to the problem (e.g. filesystem folder tree traversal...).

paulo-ferraz-oliveira:
> I'm looking into why the issue occurs with `["src/**]`, but I need to understand initial requirements:
> 
> 1. is declaring `["src"]` supposed to be recursive?
> 2. is declaring `["src/**"]` supposed to be recursive?
> 
> I feel there's a simpler approach to the problem (e.g. filesystem folder tree traversal...).

I think I found my answer here: https://erlang.org/doc/man/filelib.html#wildcard-1.

src/elvis_file.erl
```diff
@@ -128,6 +130,9 @@ replace_stars(Glob) -> re:replace(Glob, "[[][*][]]", ".*", [global]).
 
 replace_questions(Glob) -> re:replace(Glob, "[[][?][]]", ".", [global]).
 
+reduce_stars(DirAndFilter) ->
+    re:replace(DirAndFilter, "/\\*+/", "/", [global, {return, list}]).
```
paulo-ferraz-oliveira, line 134: This improves the analysis scope. Specifically, it deals with all the issues in https://github.com/inaka/elvis/issues/508.

elbrujohalcon:
Thank you, @paulo-ferraz-oliveira !!!

paulo-ferraz-oliveira:
> Thank you, @paulo-ferraz-oliveira !!!

Oh, this is nice 👍

PR #126, opened 2020-09-08
More visible defaults

paulo-ferraz-oliveira:
I tried to completely bridge the gap between the `elvis_rulesets` code, the `elvis_style` code and the [Rules Wiki](https://github.com/inaka/elvis_core/wiki).

### DIFF
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -6,18 +6,18 @@
 rules(erl_files) ->
   [ {elvis_style, line_length, #{limit => 100, skip_comments => false}}
   , {elvis_style, no_tabs}
-  , {elvis_style, no_trailing_whitespace}
+  , {elvis_style, no_trailing_whitespace, #{ignore_empty_lines => false}}
   , {elvis_style, macro_names}
   , {elvis_style, macro_module_names}
   , { elvis_style
     , operator_spaces
     , #{rules => [{right, ","}, {right, "++"}, {left, "++"}]}
     }
-  , {elvis_style, nesting_level, #{level => 4}}
-  , {elvis_style, god_modules, #{limit => 25}}
+  , {elvis_style, nesting_level, #{level => 4, ignore => []}}
+  , {elvis_style, god_modules, #{limit => 25, ignore => []}}
   , {elvis_style, no_if_expression}
-  , {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}}
-  , {elvis_style, used_ignored_variable}
+  , {elvis_style, invalid_dynamic_call, #{ignore => []}}
+  , {elvis_style, used_ignored_variable, #{ignore => []}}
   , {elvis_style, no_behavior_info}
   , { elvis_style
     , module_naming_convention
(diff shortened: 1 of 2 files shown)


src/elvis_rulesets.erl
```diff
@@ -6,36 +6,46 @@
...
   , { elvis_style
     , operator_spaces
     , #{rules => [{right, ","}, {right, "++"}, {left, "++"}]}
     }
-  , {elvis_style, nesting_level, #{level => 4}}
-  , {elvis_style, god_modules, #{limit => 25}}
+  , {elvis_style, nesting_level, #{level => 4, ignore => []}}
+  , {elvis_style, god_modules, #{limit => 25, ignore => []}}
   , {elvis_style, no_if_expression}
-  , {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}}
```
paulo-ferraz-oliveira, line 19: Eliminated reference to `elvis`.

src/elvis_rulesets.erl
```diff
@@ -6,36 +6,46 @@
...
     }
   , { elvis_style
     , function_naming_convention
-    , #{regex => "^([a-z][a-z0-9]*_?)*$"}
+    , #{regex => "^([a-z][a-z0-9]*_?)*$", ignore => []}
     }
   , {elvis_style, state_record_and_type}
   , {elvis_style, no_spec_with_records}
   , {elvis_style, dont_repeat_yourself, #{min_complexity => 10}}
-  , {elvis_style, no_debug_call, #{ignore => [elvis, elvis_utils]}}
```
paulo-ferraz-oliveira, line 33: Eliminated reference to `elvis` and `elvis_utils`.

src/elvis_rulesets.erl
```diff
@@ -6,36 +6,46 @@
...
+  , {elvis_style, max_module_length, #{max_length => 500, ignore => [], count_comments => false,
```
paulo-ferraz-oliveira, line 33: Added stuff that was left out.

elbrujohalcon: This was left out _on purpose_. The default ruleset doesn't define a max length for a module. If you want to define a max length for your modules, you can explicitly add it to the `rules` section of your `elvis.config`.

src/elvis_rulesets.erl
```diff
@@ -6,36 +6,46 @@
...
+                                       count_whitespace => false}}
+  , {elvis_style, no_call, #{}}
```
elbrujohalcon, line 35: This is not needed. It will just make every module run through this rule unaffected. I would leave it out and people who need this rule can add it to `rules` on their own.

src/elvis_rulesets.erl
```diff
@@ -6,36 +6,46 @@
...
+  , {elvis_style, no_debug_call, #{debug_functions => [{ct, pal}, {ct, print},
```
elbrujohalcon, line 36: It's fine to add this rule and also `no_common_caveats_call` but we'll need to remember to do a jump in (I think) _minor_ version when we do a release since this will probably change the results of running `elvis` for a lot of people.

elbrujohalcon (changes requested):
A few things I would not add to the ruleset.

src/elvis_rulesets.erl
paulo-ferraz-oliveira: Oh, I see. My bad. I hand't though about that, actually, since I thought the defaults (in the Wiki) were "the defaults". If you prefer I can revert the whole thing. I don't want to cause breaking changes just because of a consistency-related change.

elbrujohalcon: I don't feel strongly about it, since I basically use the defaults from the wiki in all my projects (and sometimes even stricter versions of them, including stuff like `proplists:get_value/2,3` in `no_call`).
@jfacorro: What do you think?

paulo-ferraz-oliveira: I usually start with the simplest `elvis.config`, then start adding restrictions I see fit. In any case, I remember having been warned for `no_debug_call` and I usually don't explicitly declare it. Must have missed something...

paulo-ferraz-oliveira:
I didn't think this through properly. I guess rules are added to the default when they make sense. In any case, I guess the Wiki should probably show this ("This rule is not enforced by default"), and the proposed "minimal `elvis.config`" tweaked, in that case.

elbrujohalcon:
> In any case, I guess the Wiki should probably show this ("This rule is not enforced by default"), and the proposed "minimal `elvis.config`" tweaked, in that case.
☝️  This. 100% this.

paulo-ferraz-oliveira:
> > In any case, I guess the Wiki should probably show this ("This rule is not enforced by default"), and the proposed "minimal `elvis.config`" tweaked, in that case.
> 
> ☝️ This. 100% this.

OK, I'll do this instead, and drop the PR.

PR #128, opened 2020-09-08
Make sure gen_statem follows the same rule as gen_server

paulo-ferraz-oliveira:


elbrujohalcon:
Travis is not happy, @paulo-ferraz-oliveira :P

paulo-ferraz-oliveira:
> Travis is not happy, @paulo-ferraz-oliveira :P

Let me check what it's complaining about. Tests pass locally, at least.

paulo-ferraz-oliveira:
Hm... I seem to have called `handle_event` `handle_call`. Tests don't fail, though, locally, so there's something I'm missing here.

paulo-ferraz-oliveira:
A bit outside the scope of this, I took the liberty to update Travis CI instructions, to deal with more Erlang/OTP versions.

paulo-ferraz-oliveira:
I'm now running the same as Travis CI runs (I added a Makefile and a `ct` target to it) and can see the issues show up. This is better than having to wait for GitHub to tell me where I got stuff wrong.

paulo-ferraz-oliveira:
> Travis is not happy, @paulo-ferraz-oliveira :P

Travis is no longer angry at me.

elbrujohalcon (approved):
This PR is fine, but I would like to eventually simplify our travis integration, to something like [this](https://github.com/AdRoll/spillway/blob/master/.travis.yml) since there is support for rebar3 on Travis now.

PR #129, opened 2020-09-08
Simplify by removal of seemingly unused rules no_seqbind and no_useless_seqbind

paulo-ferraz-oliveira:
They weren't documented, in any case, so whoever was using them was peaking into the code :-)


elbrujohalcon (approved):
I ❤️  negative PRs!!

PR #131, opened 2020-09-09
Add options `regex` and `ignore` to rule `macro_names`

paulo-ferraz-oliveira:
[Fix inaka/elvis#515]

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -6,7 +6,8 @@
      [#{dirs => ["src"],
         filter => "*.erl",
         rules => [{elvis_style, invalid_dynamic_call, #{ignore => [elvis_core]}},
-                  {elvis_style, dont_repeat_yourself, #{min_complexity => 20}}],
+                  {elvis_style, dont_repeat_yourself, #{min_complexity => 20}},
+                  {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}}],
         ruleset => erl_files
        },
       #{dirs => ["."],
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -40,6 +40,10 @@
 -define(INVALID_MACRO_NAME_MSG,
         "Invalid macro name ~s on line ~p. Use UPPER_CASE.").
 
+-define(INVALID_MACRO_NAME_REGEX_MSG,
+        "The macro named ~p on line ~p does not respect the format "
+        "defined by the regular expression '~p'.").
+
 -define(MACRO_AS_MODULE_NAME_MSG,
         "Don't use macros (like ~s on line ~p) as module names.").
 -define(MACRO_MODULE_NAMES_EXCEPTIONS,
@@ -243,13 +247,26 @@ no_trailing_whitespace(_Config, Target, RuleConfig) ->
     elvis_utils:check_lines(Src, fun check_no_trailing_whitespace/3,
                             RuleConfig).
 
+-type macro_names_config() :: #{regex => string(),
+                                ignore => [module()]
+                               }.
+
 -spec macro_names(elvis_config:config(),
                   elvis_file:file(),
-                  empty_rule_config()) ->
+                  macro_names_config()) ->
     [elvis_result:item()].
-macro_names(_Config, Target, _RuleConfig) ->
+macro_names(Config, Target, RuleConfig) ->
     {Src, _} = elvis_file:src(Target),
-    elvis_utils:check_lines(Src, fun check_macro_names/3, []).
+    IgnoreModules = maps:get(ignore, RuleConfig, []),
+    Regex = maps:find(regex, RuleConfig),
+    {Root, _File} = elvis_file:parse_tree(Config, Target),
+    ModuleName = elvis_code:module_name(Root),
+    case lists:member(ModuleName, IgnoreModules) of
+        false ->
+            elvis_utils:check_lines(Src, fun check_macro_names/3, Regex);
+        true->
+            []
+    end.
 
 -spec macro_module_names(elvis_config:config(),
                          elvis_file:file(),
@@ -860,14 +877,14 @@ check_no_trailing_whitespace(Line, Num, RuleConfig) ->
 
 %% Macro Names
 
--spec check_macro_names(binary(), integer(), [term()]) ->
+-spec check_macro_names(binary(), integer(), ConfigRegex :: {ok, string()} | error) ->
     no_result | {ok, elvis_result:item()}.
-check_macro_names(Line, Num, _Args) ->
-    {ok, Regex} = re:compile("^ *[-]define *[(]([^,(]+)"),
-    case re:run(Line, Regex, [{capture, all_but_first, list}]) of
+check_macro_names(Line, Num, ConfigRegex) ->
+    {ok, ReCompiled} = re:compile("^ *[-]define *[(] *([']([^']*)[']|[^,( ]+)"),
+    case re:run(Line, ReCompiled, [{capture, [1], list}]) of
         nomatch ->
             no_result;
-        {match, [MacroName]} ->
+        {match, [MacroName]} when ConfigRegex =:= error ->
             case string:to_upper(MacroName) of
                 MacroName ->
                     no_result;
@@ -875,8 +892,32 @@ check_macro_names(Line, Num, _Args) ->
                     Msg = ?INVALID_MACRO_NAME_MSG,
                     Result = elvis_result:new(item, Msg, [MacroName, Num], Num),
                     {ok, Result}
-            end
+            end;
+        {match, [MacroName]} ->
+            {ok, Regexp} = ConfigRegex,
+            Subject = string:strip(MacroName, both, $'),
+            {ok, RE} = re:compile(Regexp),
+            Options = [{capture, all, list}, global],
+            ReRunRes = re:run(Subject, RE, Options),
+            check_macro_name_regex(ReRunRes, {Subject, Num, Regexp})
     end.
+
+-spec check_macro_name_regex(ReRunRes :: {match, Captured :: [{integer(), integer()}]} | nomatch,
+                             {MacroName :: list(), LineNum :: integer(), Regexp :: string()}) ->
+    no_result | {ok, elvis_result:item()}.
+check_macro_name_regex({match, Captured} = _ReRunRes, {MacroName, LineNum, Regexp}) ->
+    MatchesCompletely = lists:flatten(Captured) =:= MacroName,
+    case MatchesCompletely of
+        true ->
+            no_result;
+        false ->
+            check_macro_name_regex(nomatch, {MacroName, LineNum, Regexp})
+    end;
+check_macro_name_regex(nomatch = _ReRunRes, {MacroName, LineNum, Regex}) ->
+    Msg = ?INVALID_MACRO_NAME_REGEX_MSG,
+    Info = [MacroName, LineNum, Regex],
+    Result = elvis_result:new(item, Msg, Info, LineNum),
+    {ok, Result}.
 
 %% Macro in Function Call as Module or Function Name
 
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -259,7 +259,8 @@ macro_names(Config, Target, RuleConfig) ->
     case lists:member(ModuleName, IgnoreModules) of
         false ->
             Regexp = maps:get(regex, RuleConfig, "^([A-Z][A-Z_0-9]+)$"),
-            MacroNodes = elvis_code:find(fun is_macro_define_node/1, Root, #{traverse => all, mode => node}),
+            MacroNodes = elvis_code:find(fun is_macro_define_node/1, Root,
+                                         #{traverse => all, mode => node}),
             check_macro_names(Regexp, MacroNodes, _ResultsIn = []);
         true->
             []
@@ -695,7 +696,8 @@ no_common_caveats_call(Config, Target, RuleConfig) ->
                   {timer, send_interval, 3},
                   {erlang, size, 1}
                  ],
-    no_call_common(Config, Target, RuleConfig, caveat_functions, DefaultFns, ?NO_COMMON_CAVEATS_CALL_MSG).
+    no_call_common(Config, Target, RuleConfig, caveat_functions, DefaultFns,
+                   ?NO_COMMON_CAVEATS_CALL_MSG).
 
 -spec node_line_limits(ktn_code:tree_node())->
     {Min :: integer(), Max :: integer()}.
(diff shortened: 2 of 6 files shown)


src/elvis_style.erl
```diff
@@ -860,24 +877,48 @@ check_no_trailing_whitespace(Line, Num, RuleConfig) ->
...
 %% Macro Names
 
--spec check_macro_names(binary(), integer(), [term()]) ->
+-spec check_macro_names(binary(), integer(), ConfigRegex :: {ok, string()} | error) ->
     no_result | {ok, elvis_result:item()}.
-check_macro_names(Line, Num, _Args) ->
-    {ok, Regex} = re:compile("^ *[-]define *[(]([^,(]+)"),
-    case re:run(Line, Regex, [{capture, all_but_first, list}]) of
+check_macro_names(Line, Num, ConfigRegex) ->
+    {ok, Regex} = re:compile("^ *[-]define *[(] *([']([^']*)[']|[^,( ]+)"),
```
paulo-ferraz-oliveira, line 883: I found this regexp to not cover stuff like `-define('A,a', 2).` which, although ugly, might show up in the wild.

Let me know if this change is OK. My goal here is to try and fetch:
1. either `'Macro'` (' followed by a potential repetition of non-' followed by '),
2. or `Macro` (a potential repetition of non ", or (" chars.

src/elvis_style.erl
```diff
@@ -860,24 +877,48 @@ check_no_trailing_whitespace(Line, Num, RuleConfig) ->
...
+    case re:run(Line, Regex, [{capture, [1], list}]) of
```
paulo-ferraz-oliveira, line 884: Added this because of the group that's created in the new match.

test/style_SUITE.erl
```diff
@@ -246,7 +246,13 @@ verify_macro_names_rule(_Config) ->
     File = "fail_macro_names.erl",
     {ok, Path} = elvis_test_utils:find_file(SrcDirs, File),
 
-    [_, _] = elvis_style:macro_names(ElvisConfig, Path, #{}).
+    [_, _] = elvis_style:macro_names(ElvisConfig, Path, #{}),
+
+    [_] = elvis_style:macro_names(ElvisConfig, Path, #{ regex => "^[A-Za-z_ ]+$" }),
```
paulo-ferraz-oliveira, line 251: Which a regex stuff that was failing (at least 1) is not failing any more.

test/style_SUITE.erl
```diff
@@ -246,7 +246,13 @@ verify_macro_names_rule(_Config) ->
...
+
+    [] = elvis_style:macro_names(ElvisConfig, Path, #{ regex => "^[A-Za-z_ \-]+$" }),
```
paulo-ferraz-oliveira, line 253: With an "improved" regexp nothing fails.

test/style_SUITE.erl
```diff
@@ -246,7 +246,13 @@ verify_macro_names_rule(_Config) ->
...
+
+    [_, _, _, _, _] = elvis_style:macro_names(ElvisConfig, Path, #{ regex => "^POTENTIAL_BAD-NAME$" }).
```
paulo-ferraz-oliveira, line 255: With a very targeted regexp almost everything fails.

paulo-ferraz-oliveira:
Let me know if I should add more tests or make them more readable.

Actually, now that I think about it I didn't add a test for `ignore`. I'ma do it.

paulo-ferraz-oliveira:
> Let me know if I should add more tests or make them more readable.
> 
> Actually, now that I think about it I didn't add a test for `ignore`. I'ma do it.

There you go: 6057b99.

paulo-ferraz-oliveira:
Added a new regression test for https://github.com/inaka/elvis/issues/515#issuecomment-689870216.

src/elvis_style.erl
elbrujohalcon: This is very much *intentional*. We do want to emit a warning if people use `A,a` as a macro name. That's the whole point :)

test/examples/fail_macro_names.erl
```diff
@@ -5,6 +5,8 @@
 -define(wtf_NAME,   "(yuno)").
 -define(GOOD_NAME(Arg), "(megusta)").
 -define(GOOD_NAME_TOO (Arg), "(megusta)").
+-define(  'POTENTIAL_BAD-NAME'  , nomegusta).
+-define('A,aZ', 2).
```
elbrujohalcon, line 9: These 2 should definitely be rejected by the rule (at least, with default configuration).

elbrujohalcon (changes requested):

src/elvis_style.erl
elbrujohalcon: In other words: Make the default regex as strict as possible.

elbrujohalcon: Yeah… probably, _the right solution:tm:_ …      would be to actually use the AST for this instead of parsing the code manually as a string.
In any case, this change is fine until that day…

elbrujohalcon (approved):

src/elvis_style.erl
```diff
@@ -243,13 +247,26 @@ no_trailing_whitespace(_Config, Target, RuleConfig) ->
...
                   elvis_file:file(),
-                  empty_rule_config()) ->
+                  macro_names_config()) ->
     [elvis_result:item()].
-macro_names(_Config, Target, _RuleConfig) ->
+macro_names(Config, Target, RuleConfig) ->
     {Src, _} = elvis_file:src(Target),
-    elvis_utils:check_lines(Src, fun check_macro_names/3, []).
+    IgnoreModules = maps:get(ignore, RuleConfig, []),
+    Regex = maps:find(regex, RuleConfig),
```
elbrujohalcon, line 261: What if we just use a default regex, like `[A-Z][A-Z_0-9]+`?

src/elvis_style.erl
paulo-ferraz-oliveira: I've never used the AST, but could try to check it out.

src/elvis_style.erl
paulo-ferraz-oliveira: That would break those cases where `to_upper` was valid (e.g. 'A-Z', even if ugly), but if it's something you prefer we can go with that.

elbrujohalcon: Yeah, that was the original intention anyway. Let's break ugly macro names! 💣

src/elvis_style.erl
elbrujohalcon: There are several other rules using it, like the one for atoms you created in the other PR ;)

paulo-ferraz-oliveira: Oh, you mean that `katana` stuff.

src/elvis_style.erl
paulo-ferraz-oliveira: Done in eca94aa.

src/elvis_style.erl
```diff
@@ -37,9 +37,6 @@
 -define(NO_TRAILING_WHITESPACE_MSG,
         "Line ~b has ~b trailing whitespace characters.").
 
--define(INVALID_MACRO_NAME_MSG,
```
paulo-ferraz-oliveira, line 40: This no longer makes sense, since we now have a regexp-based approach.

test/style_SUITE.erl
```diff
@@ -246,7 +246,7 @@ verify_macro_names_rule(_Config) ->
     File = "fail_macro_names.erl",
     {ok, Path} = elvis_test_utils:find_file(SrcDirs, File),
 
-    [_, _, _] = elvis_style:macro_names(ElvisConfig, Path, #{}),
+    [_, _, _, _] = elvis_style:macro_names(ElvisConfig, Path, #{}),
```
paulo-ferraz-oliveira, line 249: A new one started to fail, since now were not simply considering `to_upper`.

src/elvis_style.erl
paulo-ferraz-oliveira: Done in 73f125c. I'll open a new discussion there, since there are details I'm not comfortable with.

src/elvis_style.erl
```diff
@@ -874,30 +874,45 @@ check_no_trailing_whitespace(Line, Num, RuleConfig) ->
...
+                  Info = [MacroNameOriginal, Line, Regexp],
+                  Result = elvis_result:new(item, Msg, Info),
+                  ResultsIn ++ [Result];
+              {match, _Captured} ->
+                  ResultsIn
+          end,
+    check_macro_names(Regexp, RemainingMacroNodes, ResultsOut).
+
+is_macro_define_node(MaybeMacro) ->
+    case ktn_code:type(MaybeMacro) of
```
paulo-ferraz-oliveira, line 896: Isn't there a better way to do this? I looked at what `katana` produced and basically got stuff from there, but I feel there's a better way.

src/elvis_style.erl
```diff
@@ -874,30 +874,45 @@ check_no_trailing_whitespace(Line, Num, RuleConfig) ->
...
-    no_result;
-check_macro_name_regex(nomatch = _ReRunRes, {MacroName, LineNum, ConfigRegex}) ->
-    Msg = ?INVALID_MACRO_NAME_REGEX_MSG,
-    Info = [MacroName, LineNum, ConfigRegex],
-    Result = elvis_result:new(item, Msg, Info, LineNum),
-    {ok, Result}.
+macro_name_from_node(MacroNode) ->
+    MacroNodeValue = ktn_code:attr(value, MacroNode),
+    MacroAsAtom
+        = case lists:keyfind(_KeyVar = var, _N = 1, MacroNodeValue) of
```
paulo-ferraz-oliveira, line 906: Also here, I feel there must be a better way to do this. Initially, I targeted type `macro` but that actually shows up when a macro is used, not defined, which is what we're interested in.

I later used `'ATOM'` in the macro definition, which provoked an error, since
```erlang
-define(MACRO, 2).
```
produces different `katana` structures from
```erlang
-define('MACRO', 2).
```
(first one is a `var`, second one is an `atom`).

Also, I didn't want to deep-dive into `katana`, nor could I find up-to-date information on the API.

paulo-ferraz-oliveira:
Anyway, this is working (ups, tests are failing), and I'll want to `elvis` `elvis_core` since there's some stuff needing fixes there, too.

elvis.config
```diff
@@ -6,7 +6,8 @@
      [#{dirs => ["src"],
         filter => "*.erl",
         rules => [{elvis_style, invalid_dynamic_call, #{ignore => [elvis_core]}},
-                  {elvis_style, dont_repeat_yourself, #{min_complexity => 20}}],
+                  {elvis_style, dont_repeat_yourself, #{min_complexity => 20}},
+                  {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}}],
```
paulo-ferraz-oliveira, line 10: Not too sure about this one, but there be `io:format` inside.

paulo-ferraz-oliveira:
> I'll want to elvis-lint elvis_core

d50fbb2

elvis.config
elbrujohalcon: Yeah, `elvis` is a CLI… it *has* to have `io:format` somewhere.

src/elvis_style.erl
elbrujohalcon: I think this is _the right way_:tm:.
Macros are very hard to parse. I learned that while implementing the formatter 🤦

src/elvis_style.erl
elbrujohalcon: It's very very ugly… but I think this is the only way.
The alternative would be to use `ktn_dodger`, as we do in the formatter, but that comes with another huge pile of nightmares.

elbrujohalcon (approved):
I think this is good to go.

elbrujohalcon:
@paulo-ferraz-oliveira once you fix the tests, let me know and I can merge this.

elvis.config
paulo-ferraz-oliveira: If only there were an `allow_debug_call` :D to which we would add non-ignored functions :D

paulo-ferraz-oliveira:
What about the fact that `katana-code`'s `-spec(_).` is incomplete/wrong. Should we care about that (in the context of another repo, ofc)?

paulo-ferraz-oliveira:
> @paulo-ferraz-oliveira once you fix the tests, let me know and I can merge this.

Pushed c414b1f. I don't exactly know what `katana-code`'s main goals are, but the new addition of code (about getting a macro's name) could well live inside it, right? ... since we're messing, in `elvis_core`'s context with `katana-code`'s internal structures.

elbrujohalcon:
@jfacorro ☝️That's your territory :P

src/elvis_style.erl
```diff
@@ -860,24 +876,49 @@ check_no_trailing_whitespace(Line, Num, RuleConfig) ->
...
-        {match, [MacroName]} ->
-            case string:to_upper(MacroName) of
-                MacroName ->
-                    no_result;
-                _ ->
-                    Msg = ?INVALID_MACRO_NAME_MSG,
-                    Result = elvis_result:new(item, Msg, [MacroName, Num], Num),
-                    {ok, Result}
-            end
+check_macro_names(_Regexp, [] = _MacroNodes, ResultsIn) ->
```
jfacorro, line 879: Why the removal of the spec? 😄

jfacorro:
> I don't exactly know what katana-code's main goals are, but the new addition of code (about getting a macro's name) could well live inside it, right? ... since we're messing, in elvis_core's context with katana-code's internal structures.

I couldn't find in this PR's code where we are manipulating internal katana-code data structures. As far as I could see we are using `ktn_code:attr/2` for getting values.

src/elvis_style.erl
paulo-ferraz-oliveira: You're right, this shouldn't have happened. Since I rewrote the function I accidentally took it out. In my lib.s I'm usually as strict as `missing_spec_all`. Shall I re-add it?

paulo-ferraz-oliveira:
> > I don't exactly know what katana-code's main goals are, but the new addition of code (about getting a macro's name) could well live inside it, right? ... since we're messing, in elvis_core's context with katana-code's internal structures.
> 
> I couldn't find in this PR's code where we are manipulating internal katana-code data structures. As far as I could see we are using `ktn_code:attr/2` for getting values.

What I meant was "I'm handling a `katana-code` generated structure that can eventually change (https://github.com/inaka/elvis_core/pull/131/files#diff-9d4ce82551a31e44e1d37fb5c5966363R913) (and the code that does it, in the scope of `elvis` might surely be incomplete), especially since its `spec` specifies the return as `term()`..." 😄. I had proposed that this code be moved to `katana-code` since knowing about the internals shouldn't be required at this level, I guess. (I usually prefer opaque data structures that I "query" the producer for information about). No biggie, if we want it elsewhere, at least in the context of `elvis` it's now a function by itself.

PR #132, opened 2020-09-10
Add new rule `atom_naming_convention`

paulo-ferraz-oliveira:
[Fix inaka/elvis#516].

### DIFF
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -36,6 +36,7 @@ rules(erl_files) ->
     , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$", ignore => []}
     }
   , {elvis_style, no_nested_try_catch, #{ignore => []}}
+  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9_]+)$", ignore => []}}
   ];
 rules(makefiles) ->
   [ {elvis_project, no_deps_master_erlang_mk, #{ignore => []}}
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -36,7 +36,9 @@ rules(erl_files) ->
     , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$", ignore => []}
     }
   , {elvis_style, no_nested_try_catch, #{ignore => []}}
-  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$", ignore => []}}
+  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$",
+                                            enclosed_atoms => ".*",
+                                            ignore => []}}
   ];
 rules(makefiles) ->
   [ {elvis_project, no_deps_master_erlang_mk, #{ignore => []}}
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -36,7 +36,7 @@ rules(erl_files) ->
     , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$", ignore => []}
     }
   , {elvis_style, no_nested_try_catch, #{ignore => []}}
-  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9_]+)$", ignore => []}}
+  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$", ignore => []}}
   ];
 rules(makefiles) ->
   [ {elvis_project, no_deps_master_erlang_mk, #{ignore => []}}
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -763,7 +763,7 @@ check_atom_names(_Regex, _RegexEnclosed, [] = _AtomNodes, Acc) ->
     Acc;
 check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) ->
     AtomName0 = ktn_code:attr(text, AtomNode),
-    AtomName = string:strip(AtomName0, both, $'),
+    AtomName = string_strip_one(AtomName0, "'"),
     IsEnclosed = is_enclosed_atom(AtomName0, AtomName),
     RE = re_compile_for_atom_type(IsEnclosed, Regex, RegexEnclosed),
     AccOut
@@ -778,6 +778,11 @@ check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) -
                   AccIn
           end,
     check_atom_names(Regex, RegexEnclosed, RemainingAtomNodes, AccOut).
+
+string_strip_one(Str0, Char) ->
+    StrR = string:replace(Str0, Char, "", trailing),
+    StrL = string:replace(StrR, Char, "", leading),
+    StrL.
 
 re_compile_for_atom_type(false = _IsEnclosed, Regex, _RegexEnclosed) ->
     {ok, RE} = re:compile(Regex),
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -740,7 +740,7 @@ atom_naming_convention(Config, Target, RuleConfig) ->
     IgnoreModules = maps:get(ignore, RuleConfig, []),
     case lists:member(ModuleName, IgnoreModules) of
         false ->
-            Regex = maps:find(regex, RuleConfig),
+            Regex = maps:get(regex, RuleConfig, "^([a-z][a-z0-9]*_?)*(_SUITE)?$"),
             AtomNodes = elvis_code:find(fun is_atom_node/1, Root, #{traverse => all, mode => node}),
             check_atom_names(Regex, AtomNodes, []);
         true ->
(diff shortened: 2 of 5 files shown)


paulo-ferraz-oliveira:
Implementation should probably be tested in a few different projects before we can accept the default `regex`.

src/elvis_rulesets.erl
```diff
@@ -36,6 +36,7 @@ rules(erl_files) ->
     , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$", ignore => []}
     }
   , {elvis_style, no_nested_try_catch, #{ignore => []}}
+  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9_]+)$", ignore => []}}
```
paulo-ferraz-oliveira, line 39: I'm unsure about the default value. It works OK for most of the stuff I tested, but e.g. it doesn't when we call a `_SUITE` module (which is OK for rule `module_naming_convention`'s default).

src/elvis_style.erl
```diff
@@ -721,10 +726,55 @@ no_nested_try_catch(Config, Target, RuleConfig) ->
...
+                             atom_naming_convention_config()) ->
+    [elvis_result:item()].
+atom_naming_convention(Config, Target, RuleConfig) ->
+    {Root, _File} = elvis_file:parse_tree(Config, Target),
+    ModuleName = elvis_code:module_name(Root),
+    IgnoreModules = maps:get(ignore, RuleConfig, []),
+    case lists:member(ModuleName, IgnoreModules) of
+        false ->
+            Regex = maps:find(regex, RuleConfig),
+            AtomNodes = elvis_code:find(fun is_atom_node/1, Root, #{traverse => all, mode => node}),
```
paulo-ferraz-oliveira, line 744: I'm unsure about the `mode`. I think I tested with zipper and it worked OK too.

src/elvis_style.erl
```diff
@@ -721,10 +726,55 @@ no_nested_try_catch(Config, Target, RuleConfig) ->
...
 %% Private
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+check_atom_names(error = _Regex, _AtomNodes, _Acc) ->
+    [];
+check_atom_names({ok, _Regex}, [] = _AtomNodes, Acc) ->
+    Acc;
+check_atom_names({ok, Regex}, [AtomNode | RemainingAtomNodes], AccIn) ->
+    AtomName0 = ktn_code:attr(text, AtomNode),
+    AtomName = string:strip(AtomName0, both, $'),
```
paulo-ferraz-oliveira, line 760: I do this so that we may analyze the name itself (considering an atom with `'`) but it might be misleading regarding the regex we include in the config., which will not require `'`. Well, we _are_ talking about the "name".

src/elvis_style.erl
```diff
@@ -721,10 +726,55 @@ no_nested_try_catch(Config, Target, RuleConfig) ->
...
+    {ok, RE} = re:compile(Regex),
+    AccOut
+        = case re:run(_Subject = AtomName, RE) of
+              nomatch ->
+                  Msg = ?ATOM_NAMING_CONVENTION_MSG,
+                  {Line, _} = ktn_code:attr(location, AtomNode),
+                  Info = [AtomName0, Line, Regex],
```
paulo-ferraz-oliveira, line 767: ... though I still print it has found in the code.

src/elvis_rulesets.erl
elbrujohalcon: 🤔  The original idea was to prevent stuff in `camelCase`… so I would just add `(_SUITE)?` to this default, just in case, and be done with it.

elbrujohalcon: We can always adjust the default later if people complain about it too much.

src/elvis_style.erl
elbrujohalcon: I'm unsure as well, but if it works… it's probably fine.

src/elvis_style.erl
elbrujohalcon: I would've done the same… or maybe even ignore atoms that are enclosed in `'` altogether, considering that if you _had_ to use apostrophes, then your atom is not "normal" and should probably not be checked with the regex.

Maybe you can add a configuration option, like `enclosed_atoms => regex()`, that defaults to `.*`.
Or (without that parameter), check the whole thing with the regex, but add something like `|('.*')` to the default regex so that it immediately skips all enclosed atoms.

I personally favor the addition of `enclosed_atoms => same | regex()`, so that the user can specify that they want to use the same regex for all atoms regardless of enclosing. Not sure why someone would want that, instead of just removing the `'`s, but… 🤷

src/elvis_style.erl
elbrujohalcon: This is the right thing to do

elbrujohalcon (changes requested):

src/elvis_style.erl
paulo-ferraz-oliveira: Do you think it would be best to find somebody that was not unsure (i.e. "sure") :-) and teach us both something?

elbrujohalcon: I was actually waiting for @jfacorro to comment… but it's likely that he'll just say "it's fine as it is, folks!"

src/elvis_rulesets.erl
paulo-ferraz-oliveira: Done: 0f1a7f7

src/elvis_style.erl
paulo-ferraz-oliveira: This should do it: 56eb5f8.

src/elvis_style.erl
```diff
@@ -721,10 +726,72 @@ no_nested_try_catch(Config, Target, RuleConfig) ->
...
+    Regex;
+enclosed_atoms_regex_or_same(RegexEnclosed, _Regex) ->
+    RegexEnclosed.
+
+check_atom_names(_Regex, _RegexEnclosed, [] = _AtomNodes, Acc) ->
+    Acc;
+check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) ->
+    AtomName0 = ktn_code:attr(text, AtomNode),
+    AtomName = string:strip(AtomName0, both, $'),
+    IsEnclosed = is_enclosed_atom(AtomName0, AtomName),
```
elbrujohalcon, line 767: Does this work with `'\' this nasty atom\''`?
Maybe a more naïve, but still reasonable approach would be…

```erlang
IsEnclosed = AtomName0 =/= AtomName
```

…or maybe instead of `string:strip/3` you should use plain ol' pattern-matching… `[$'|RestOfTheAtomName]`…

elbrujohalcon (changes requested):

test/examples/fail_atom_naming_convention.erl
```diff
@@ -0,0 +1,9 @@
+-module(fail_atom_naming_convention).
+
+-export([for_test/0]).
+
+for_test() ->
+    this_is_not_an_OK_atom,
+    'and_neither-is_this',
```
elbrujohalcon, line 7: Please add a few atoms with backticks in them (i.e. `'\'', '\'\'', '\'startswithbacktick', 'backtick\'inside', 'backtick at the end\''`).

src/elvis_style.erl
paulo-ferraz-oliveira: > Does this work with `'\' this nasty atom\''`?

I can look closer into it (and add it to the tests). I don't see why it shouldn't. As long as `katana-code` is returning the atom as "\' this nasty atom\'", the comparison shouldn't fail.

paulo-ferraz-oliveira: Well, yeah, I've found an issue with the strip'ing. I'm going to use `string:replace` instead. I also found that the error message would always output `regex` and the non-conformant element, instead of `enclosed_atoms`, where it should. So that should be fixed also.

test/examples/fail_atom_naming_convention.erl
paulo-ferraz-oliveira: I'll add this and update the tests. Thanks.

src/elvis_style.erl
paulo-ferraz-oliveira: This should do the trick: db917ea. (all those "nasty" atoms were added to the test suite)

elbrujohalcon:
Travis is still unhappy :(

paulo-ferraz-oliveira:
> Travis is still unhappy :(

It doesn't like `string:replace/4` on OTP 19.3.

elbrujohalcon:
Then why not just using good ol' pattern-matching over lists?
```erlang
{IsEnclosed, AtomName} =
  case AtomName0 of
    [$'|Rest] ->
      [$'|Reversed] = lists:reverse(Rest),
      {true, lists:reverse(Reversed)};
    AtomName0 ->
      {false, AtomName0}
  end,
```

paulo-ferraz-oliveira:
> Then why not just using good ol' pattern-matching over lists?
> 
> ```erlang
> {IsEnclosed, AtomName} =
>   case AtomName0 of
>     [$'|Rest] ->
>       [$'|Reversed] = lists:reverse(Rest),
>       {true, lists:reverse(Reversed)};
>     AtomName0 ->
>       {false, AtomName0}
>   end,
> ```

I can use this, but I do see a benefit to using less code to achieve the same result.

PR #134, opened 2020-09-16
Complete removal of `seqbind` references

paulo-ferraz-oliveira:
I didn't think about looking for `seqbind` in the code base, at the time


elbrujohalcon (approved):

elbrujohalcon:
You need to update rebar.lock as well

paulo-ferraz-oliveira:
`seqbind` isn't there, but I can push an updated (as in `rebar 3.14.1`) file, nonetheless.

elbrujohalcon:
Oh, right… it was a test dep… 🤦 
It's fine to push a new rebar.lock, yes… but you also need to appease Travis.

paulo-ferraz-oliveira:
> Oh, right… it was a test dep… 🤦
> It's fine to push a new rebar.lock, yes… but you also need to appease Travis.

I'll look into why the tests are failing, since locally they're not (using the `Makefile` procedure`).

elbrujohalcon:
As a matter of fact, if you could simplify Travis integration to be something more akin to [this](https://github.com/AdRoll/spillway/blob/master/.travis.yml) (without the formatting), that would be lovely. We had too many hacks in our Travis integration back then because the support for Erlang was limited.

paulo-ferraz-oliveira:
What `email` would you use? Or shouldn't that option exist?

elbrujohalcon:
Remove that part.

paulo-ferraz-oliveira:
Should we also remove the `rebar3` executable (found at the root)?

elbrujohalcon:
Yes, remove that one, too.

elbrujohalcon (approved):

PR #135, opened 2020-09-17
Decrease chance of error and improve maintenance

paulo-ferraz-oliveira:
I found different defaults in `elvis_style` and `elvis_rulesets`, and in those cases the public Wiki won.

[Fix inaka/elvis#525].

### DIFF
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -4,42 +4,31 @@
 
 -spec rules(Group::atom()) -> list().
 rules(erl_files) ->
-  [ {elvis_style, line_length, #{limit => 100, skip_comments => false}}
-  , {elvis_style, no_tabs}
-  , {elvis_style, no_trailing_whitespace, #{ignore_empty_lines => false}}
-  , {elvis_style, macro_names, #{regex => "^([A-Z][A-Z_0-9]+)$", ignore => []}}
-  , {elvis_style, macro_module_names}
-  , { elvis_style
-    , operator_spaces
-    , #{rules => [{right, ","}, {right, "++"}, {left, "++"}]}
-    }
-  , {elvis_style, nesting_level, #{level => 4, ignore => []}}
-  , {elvis_style, god_modules, #{limit => 25, ignore => []}}
-  , {elvis_style, no_if_expression}
-  , {elvis_style, invalid_dynamic_call, #{ignore => []}}
-  , {elvis_style, used_ignored_variable, #{ignore => []}}
-  , {elvis_style, no_behavior_info}
-  , { elvis_style
-    , module_naming_convention
-    , #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$", ignore => []}
-    }
-  , { elvis_style
-    , function_naming_convention
-    , #{regex => "^([a-z][a-z0-9]*_?)*$", ignore => []}
-    }
-  , {elvis_style, state_record_and_type}
-  , {elvis_style, no_spec_with_records}
-  , {elvis_style, dont_repeat_yourself, #{min_complexity => 10}}
-  , {elvis_style, no_debug_call, #{ignore => []}}
-  , { elvis_style
-    , variable_naming_convention
-    , #{regex => "^_?([A-Z][0-9a-zA-Z]*)$", ignore => []}
-    }
-  , {elvis_style, no_nested_try_catch, #{ignore => []}}
-  , {elvis_style, atom_naming_convention, #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$",
-                                            enclosed_atoms => ".*",
-                                            ignore => []}}
-  ];
+    lists:map(
+        fun (Rule) ->
+            {elvis_style, Rule, elvis_style:default(Rule)}
+        end,
+        [ line_length
+        , no_tabs
+        , no_trailing_whitespace
+        , macro_names
+        , macro_module_names
+        , operator_spaces
+        , nesting_level
+        , god_modules
+        , no_if_expression
+        , invalid_dynamic_call
+        , used_ignored_variable
+        , no_behavior_info
+        , module_naming_convention
+        , function_naming_convention
+        , state_record_and_type
+        , no_spec_with_records
+        , dont_repeat_yourself
+        , no_debug_call
+        , variable_naming_convention
+        , no_nested_try_catch
+        , atom_naming_convention ]);
 rules(makefiles) ->
   [ {elvis_project, no_deps_master_erlang_mk, #{ignore => []}}
   , {elvis_project, protocol_for_deps_erlang_mk, #{ignore => []}}
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -31,13 +31,26 @@ rules(erl_files) ->
         , atom_naming_convention
         ]);
 rules(makefiles) ->
-  [ {elvis_project, no_deps_master_erlang_mk, #{ignore => []}}
-  , {elvis_project, protocol_for_deps_erlang_mk, #{ignore => []}}
-  ];
+    lists:map(
+        fun (Rule) ->
+            {elvis_project, Rule, elvis_project:default(Rule)}
+        end,
+        [ no_deps_master_erlang_mk
+        , protocol_for_deps_erlang_mk
+        ]);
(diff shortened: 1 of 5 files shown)


test/style_SUITE.erl
```diff
@@ -277,7 +277,7 @@ verify_operator_spaces(_Config) ->
     File = "fail_operator_spaces.erl",
     {ok, Path} = elvis_test_utils:find_file(SrcDirs, File),
 
-    [] = elvis_style:operator_spaces(ElvisConfig, Path, #{}),
+    [] = elvis_style:operator_spaces(ElvisConfig, Path, #{rules=>[]}),
```
paulo-ferraz-oliveira, line 280: Broken because you now have to be explicit about what you're overriding.

test/style_SUITE.erl
```diff
@@ -307,7 +307,7 @@ verify_operator_spaces_latin1(_Config) ->
     File = "fail_operator_spaces_latin1.erl",
     {ok, Path} = elvis_test_utils:find_file(SrcDirs, File),
 
-    [] = elvis_style:operator_spaces(ElvisConfig, Path, #{}),
+    [] = elvis_style:operator_spaces(ElvisConfig, Path, #{rules => []}),
```
paulo-ferraz-oliveira, line 310: Broken because you now have to be explicit about what you're overriding.

paulo-ferraz-oliveira:
This isn't yet complete, but if the general direction isn't this one, I'll want to adapt before I move to the other rulesets.

paulo-ferraz-oliveira:
Anyway, I had a local version that used macros for functions `default/2` and `option`. It simplified stuff, but may not be what you wanted.

Basically, where you have 
```erlang
option(OptionName, RuleConfig, Rule)
```
you'd have
```erlang
?OPT(OptionName)
```
with 
```erlang
-define(OTP(OptionName), (fun () -> maps:get(OptionName, RuleConfig, default(?FUNCTION_NAME, OptionName)) end))
```
(with the assumption that `RuleConfig` was defined in the call context - I know, "dirty").

paulo-ferraz-oliveira:
I then added a further change where `default(?FUNCTION_NAME, OptionName)` would also become
```erlang
-define(?DEF(OptionName), OptionName => default(?FUNCTION_NAME, OptionName)).
% so we could build the options maps as:
#{ ?DEF(limit) },
```

paulo-ferraz-oliveira:
Why these (local) changes?

I just feel that defining stuff as
```erlang
    #{ limit => default(god_modules, limit)
     , ignore => default(god_modules, ignore)
```
is more error-prone than (though the advantage is there's no "magic")
```erlang
    #{ ?DEF(limit)
     , ?DEF(ignore) }
```

and also I'd prefer to reduce stuff like
```erlang
    Regex = option(regex, RuleConfig, function_naming_convention),
    Ignores = option(ignore, RuleConfig, function_naming_convention),
```
to
```erlang
    Regex = ?OPT(regex),
    Ignores = ?OPT(ignore),
```
thus removing the need to keep using `RuleConfig` and the function name (or `?FUNCTION_NAME`).

paulo-ferraz-oliveira:
I pushed a preparation commit for those (macro-based) changes should we want them in `elvis_style`.

src/elvis_style.erl
```diff
@@ -129,6 +130,168 @@
         "Atom ~p on line ~p does not respect the format "
         "defined by the regular expression '~p'.").
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%% Default values
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+-spec default(Rule :: atom(), OptionName :: atom()) -> DefaultOptionValue :: term().
+default(line_length, limit) -> 100;
```
elbrujohalcon, line 138: Why do we need 2 different `default` functions? Can't these values just be embedded in `default/1`?

elbrujohalcon (changes requested):

elbrujohalcon:
Oh, and please move us to `meck 0.9.0` 🙏

elbrujohalcon:
Sorry, @paulo-ferraz-oliveira … but we have a rule [against macros](https://github.com/inaka/erlang_guidelines#no-macros).
I'm inclined to reluctantly allow `?FUNCTION_NAME`… although I would not use that myself. But that's like… the farthest I'm willing to go in that direction.

src/elvis_rulesets.erl
```diff
@@ -4,42 +4,31 @@
...
+        , no_behavior_info
+        , module_naming_convention
+        , function_naming_convention
+        , state_record_and_type
+        , no_spec_with_records
+        , dont_repeat_yourself
+        , no_debug_call
+        , variable_naming_convention
+        , no_nested_try_catch
+        , atom_naming_convention ]);
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'm getting used to it. It's not something I do in my everyday projects.

src/elvis_style.erl
```diff
@@ -1578,4 +1540,4 @@ check_nested_try_catchs(ResultFun, TryExp) ->
            Rule :: atom(),
            OptionValue :: term().
 option(OptionName, RuleConfig, Rule) ->
-    maps:get(OptionName, RuleConfig, default(Rule, OptionName)).
+    maps:get(OptionName, RuleConfig, maps:get(OptionName, default(Rule))).
```
elbrujohalcon, line 1549: To avoid the extra steps if they're not needed…
```suggestion
    case RuleConfig of
        #{OptionName := Value} -> Value;
        _ -> maps:get(OptionName, default(Rule))
    end.
```

even better if you can fit that into two function clause heads.

elbrujohalcon (changes requested):

src/elvis_rulesets.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: df1fea8

src/elvis_style.erl
paulo-ferraz-oliveira: They can: 146849a

src/elvis_style.erl
```diff
@@ -129,6 +130,168 @@
...
+default(no_debug_call, debug_functions) -> [ {ct, pal}
+                                           , {ct, print}
+                                           , {io, format, 1}
+                                           , {io, format, 2}];
+default(no_common_caveats_call, ignore) -> [];
+default(no_common_caveats_call, caveat_functions) -> [ {timer, send_after, 2}
+                                                     , {timer, send_after, 3}
+                                                     , {timer, send_interval, 2}
+                                                     , {timer, send_interval, 3}
+                                                     , {erlang, size, 1}].
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: df1fea8

paulo-ferraz-oliveira:
> Oh, and please move us to `meck 0.9.0` 🙏

I didn't want to change it, since it wasn't affecting the scope of my PR. But I'll do it, as per your request.

paulo-ferraz-oliveira:
> Sorry, @paulo-ferraz-oliveira … but we have a rule [against macros](https://github.com/inaka/erlang_guidelines#no-macros).
> I'm inclined to reluctantly allow `?FUNCTION_NAME`… although I would not use that myself. But that's like… the farthest I'm willing to go in that direction.

Hm... I guess it never made it to `elvis`. It is proposed [here](https://github.com/inaka/elvis/issues/119) though. We'll eventually get there :)

src/elvis_style.erl
paulo-ferraz-oliveira: I was going to propose that but then you said it wouldn't degrade performance :D.

paulo-ferraz-oliveira: d1a8d02

paulo-ferraz-oliveira:
@elbrujohalcon, don't merge just yet; there's stuff missing for `elvis_project`. (I just wanted to understand if I was going in the right direction)

elbrujohalcon (approved):
Is it ready now, @paulo-ferraz-oliveira ?

paulo-ferraz-oliveira:
It is.

PR #136, opened 2020-09-19
Allow module-level configuration via attribute `-elvis(_)`

paulo-ferraz-oliveira:
[Fix inaka/elvis#529]

I get the `-elvis(_).` attribute from the parse file tree, and I use it to override potential `elvis.config` -level rules.

What's missing, still?
1. ~tests,~
2. ~`{function(), arity()}` where we find `module`,~
3. ~`{function()}` where we find `module` (should this one exist?),~
~4. anything else?~
~**Edit**: 4.1. documentation.~ (check inaka/elvis#535)
~**Edit**: 4.2. using `elvis_config` `merge_rules/2` instead of my own version of it.~
~**Edit**: 4.3. remove key `module`~

**Edit**: it's ready for final approval and potential merge.

### DIFF
--- a/config/test.config
+++ b/config/test.config
@@ -28,21 +28,25 @@
                   },
                   {elvis_style, state_record_and_type},
                   {elvis_style, no_spec_with_records}
-                 ]
+                 ],
+        ruleset => erl_files
        },
       #{dirs => ["../../"],
         filter => "Makefile",
         rules => [{elvis_project, no_deps_master_erlang_mk, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_erlang_mk, #{ignore => []}}]
+                  {elvis_project, protocol_for_deps_erlang_mk, #{ignore => []}}],
+        ruleset => makefiles
        },
       #{dirs => ["../../"],
         filter => "rebar.config",
         rules => [{elvis_project, no_deps_master_rebar, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_rebar, #{ignore => []}}]
+                  {elvis_project, protocol_for_deps_rebar, #{ignore => []}}],
+        ruleset => rebar_config
        },
       #{dirs => ["../../"],
         filter => "elvis.config",
-        rules => [{elvis_project, old_configuration_format}]
+        rules => [{elvis_project, old_configuration_format}],
+        ruleset => elvis_config
        }
      ]
     },
--- a/config/test.config
+++ b/config/test.config
@@ -7,45 +7,21 @@
         filter => "**.erl",
         rules => [{elvis_style, line_length, #{limit => 80,
                                                skip_comments => false}},
-                  {elvis_style, no_tabs},
-                  {elvis_style, no_trailing_whitespace},
-                  {elvis_style, macro_names},
-                  {elvis_style, macro_module_names},
-                  {elvis_style, operator_spaces, #{rules => [{right, ","},
-                                                             {right, "++"},
-                                                             {left, "++"}]}},
                   {elvis_style, nesting_level, #{level => 3}},
-                  {elvis_style, god_modules, #{limit => 25}},
-                  {elvis_style, no_if_expression},
-                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}},
-                  {elvis_style, used_ignored_variable},
-                  {elvis_style, no_behavior_info},
-                  {
-                    elvis_style,
-                    module_naming_convention,
-                    #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$",
-                      ignore => []}
-                  },
-                  {elvis_style, state_record_and_type},
-                  {elvis_style, no_spec_with_records}
+                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}}
                  ],
         ruleset => erl_files
        },
       #{dirs => ["../../"],
         filter => "Makefile",
-        rules => [{elvis_project, no_deps_master_erlang_mk, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_erlang_mk, #{ignore => []}}],
         ruleset => makefiles
        },
       #{dirs => ["../../"],
         filter => "rebar.config",
-        rules => [{elvis_project, no_deps_master_rebar, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_rebar, #{ignore => []}}],
         ruleset => rebar_config
        },
       #{dirs => ["../../"],
         filter => "elvis.config",
-        rules => [{elvis_project, old_configuration_format}],
         ruleset => elvis_config
        }
      ]
--- a/elvis.config
+++ b/elvis.config
@@ -7,7 +7,8 @@
         filter => "*.erl",
         rules => [{elvis_style, invalid_dynamic_call, #{ignore => [elvis_core]}},
                   {elvis_style, dont_repeat_yourself, #{min_complexity => 20}},
-                  {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}}],
+                  {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}},
+                  {elvis_style, god_modules, #{ignore => [elvis_style]}}],
         ruleset => erl_files
        },
       #{dirs => ["."],
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -106,7 +106,7 @@ load_file_data(Config, File) ->
     end.
 
 %% @private
--spec main([]) -> ok.
+-spec main([]) -> ok | {fail, [elvis_result:file()]}.
 main([]) ->
     ok = application:load(elvis_core),
     rock(elvis_config:from_file("elvis.config")).
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -158,7 +158,7 @@ is_elvis_attr(Node) ->
 elvis_attr_rules([] = _ElvisAttrs) ->
     undefined;
 elvis_attr_rules(ElvisAttrs) ->
-    [ktn_code:attr(value, ElvisAttr) || ElvisAttr <- ElvisAttrs].
+    [Rule || ElvisAttr <- ElvisAttrs, Rule <- ktn_code:attr(value, ElvisAttr)].
 
 apply_rule({Module, Function}, {Result, Config, File}) ->
     apply_rule({Module, Function, #{}}, {Result, Config, File});
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -11,6 +11,9 @@
 %% for internal use only
 -export([do_rock/2]).
 
+%% for eating what we cook
+-export([main/1]).
+
 -type source_filename() :: nonempty_string().
 -type target() :: source_filename() | module().
 
@@ -101,6 +104,12 @@ load_file_data(Config, File) ->
             elvis_utils:error_prn(Msg, [Reason, Path]),
             File
     end.
+
+%% @private
+-spec main([]) -> ok.
+main([]) ->
+    ok = application:load(elvis_core),
+    rock(elvis_config:from_file("elvis.config")).
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%% Private
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -17,6 +17,10 @@
 -type source_filename() :: nonempty_string().
 -type target() :: source_filename() | module().
 
+-type rule() :: {Module :: module(), Function :: atom(), Options :: #{ atom() => term() }}
+              | {Module :: module(), Function :: atom()}.
+-export_type([rule/0]).
+
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%% Public API
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
@@ -85,14 +89,14 @@ do_parallel_rock(Config0) ->
                               [], [Config], Files, Parallel),
     elvis_result_status(Results).
 
--spec do_rock(elvis_file:file(), map() | [map()]) -> {ok, elvis_result:file()}.
+-spec do_rock(elvis_file:file(), elvis_config:config() | map()) -> {ok, elvis_result:file()}.
 do_rock(File, Config) ->
     LoadedFile = load_file_data(Config, File),
     Results = apply_rules(Config, LoadedFile),
     {ok, Results}.
 
 %% @private
--spec load_file_data(map() | [map()], elvis_file:file()) -> elvis_file:file().
+-spec load_file_data(elvis_config:config() | map(), elvis_file:file()) -> elvis_file:file().
 load_file_data(Config, File) ->
     Path = elvis_file:path(File),
     elvis_utils:info("Loading ~s", [Path]),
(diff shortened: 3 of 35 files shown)


src/elvis_core.erl
```diff
@@ -126,9 +126,51 @@ apply_rules_and_print(Config, File) ->
 apply_rules(Config, File) ->
     Rules = elvis_config:rules(Config),
     Acc = {[], Config, File},
-    {RulesResults, _, _} = lists:foldl(fun apply_rule/2, Acc, Rules),
+    {ParseTree, _} = elvis_file:parse_tree(Config, File),
+    {RulesResults, _, _} = lists:foldl(fun apply_rule/2, Acc,
+                                       merge_rules(Rules, {file, ParseTree})),
     elvis_result:new(file, File, RulesResults).
 
+merge_rules(ElvisConfigRules, {file, ParseTree}) ->
```
elbrujohalcon, line 134: `elvis_config` already has a `merge_rules/2` function that merges rules and understands about `disable` and all that. Can't we just export that function and somehow use it from there?

elbrujohalcon (commented):
I think we should somehow try to reuse the code in `elvis_config` to merge rules.
Besides that. How will the `elvis` attribute look like in files?

I imagine something like the following…

```erlang
-elvis([{elvis_style, state_record_and_type, disable}]).
-elvis([{elvis_style, line_length, #{limit => 150, skip_comments => false}},
            {elvis_style, nesting_level, #{level => 2}}]).
```

i.e. lists of rule exceptions just like the ones you can find on the `rules` attribute in `elvis.config`.

elbrujohalcon:
> I get the `-elvis(_).` attribute from the parse file tree; I use it to override potential `elvis.config` -level rules.
> Is this the expected behavior (module-level overrides project-level)?

Yes.
 
> 2. `{function(), arity()}` where we find `module`,
> 3. `{function()}` where we find `module` (should this one exist?),

I wasn't expecting `module` to be found anywhere. `-elvis` should only apply to the module that's being currently analyzed. `{function(), arity()}` and/or `function()` are things that I would gladly leave for a long distant future… If people really really get used to `-elvis` so much that they require us to add support for particular rules-by-function configuration, then we do it… but otherwise… I would ignore it for now.

> 4. anything else?

Proper documentation…

src/elvis_core.erl
paulo-ferraz-oliveira: Most surely. I didn't know about it, so might have reinvented the wheel here a bit. :-)

paulo-ferraz-oliveira:
The inspiration for `#{ module => Rules }` actually came from your original comment [here](https://github.com/inaka/elvis/issues/488#issuecomment-402118939), but it's easily removable.

elbrujohalcon:
> The inspiration for `#{ module => Rules }` actually came from your original comment [here](https://github.com/inaka/elvis/issues/488#issuecomment-402118939), but it's easily removable.

Right… My bad.
I went way over my head with that suggestion… At least for the first step, I would recommend keeping things simpler.

First, let's make this work in an easy and intuitive way… and then we can make them more complex if needed.

src/elvis_core.erl
paulo-ferraz-oliveira: 1ee5c3d

paulo-ferraz-oliveira:
I wrote an initial draft of the documentation for this [here](https://github.com/inaka/elvis/issues/535). Let me know.

I think we're only missing tests now.

src/elvis_core.erl
```diff
@@ -102,6 +105,12 @@ load_file_data(Config, File) ->
             File
     end.
 
+%% @private
+-spec main([]) -> ok | {fail, [elvis_result:file()]}.
```
paulo-ferraz-oliveira, line 109: This is a very basic attempt (though working) at consuming our own produce. Comment away.

src/elvis_style.erl
```diff
@@ -887,7 +887,10 @@ atom_naming_convention(Config, Target, RuleConfig) ->
         false ->
             Regex = option(regex, RuleConfig, atom_naming_convention),
             RegexEnclosed
-                = enclosed_atoms_regex_or_same(option(enclosed_atoms, RuleConfig, atom_naming_convention), Regex),
```
paulo-ferraz-oliveira, line 890: Because of `rock`'s results.

elvis.config
```diff
@@ -7,7 +7,8 @@
         filter => "*.erl",
         rules => [{elvis_style, invalid_dynamic_call, #{ignore => [elvis_core]}},
                   {elvis_style, dont_repeat_yourself, #{min_complexity => 20}},
-                  {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}}],
+                  {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}},
+                  {elvis_style, god_modules, #{ignore => [elvis_style]}}],
```
paulo-ferraz-oliveira, line 11: Because of `rock`'s results.

src/elvis_core.erl
```diff
@@ -126,9 +135,36 @@ apply_rules_and_print(Config, File) ->
...
+merge_rules(ElvisAttrRules, ElvisConfigRules) ->
+    elvis_config:merge_rules(ElvisAttrRules, ElvisConfigRules).
+
+is_elvis_attr(Node) ->
+    ktn_code:type(Node) =:= elvis.
+
+elvis_attr_rules([] = _ElvisAttrs) ->
+    undefined;
+elvis_attr_rules(ElvisAttrs) ->
+    lists:foldl(
```
elbrujohalcon, line 161: Can we use `lists:flatmap/2` or a list comprehension for this?

elbrujohalcon (approved):
I like this.

src/elvis_core.erl
paulo-ferraz-oliveira: Sure. Not used to `flatmap`, but can give it a go... Looking at the code, though, a list comprehension seems more appropriate.

paulo-ferraz-oliveira: 87573b3

src/elvis_core.erl
```diff
@@ -158,12 +158,7 @@ is_elvis_attr(Node) ->
 elvis_attr_rules([] = _ElvisAttrs) ->
     undefined;
 elvis_attr_rules(ElvisAttrs) ->
-    lists:foldl(
-        fun (ElvisAttr, ModuleLevelRules) ->
-            ModuleLevelRules ++ ktn_code:attr(value, ElvisAttr)
-        end,
-        [],
-        ElvisAttrs).
+    [ktn_code:attr(value, ElvisAttr) || ElvisAttr <- ElvisAttrs].
```
elbrujohalcon, line 161: Shouldn't this be a flat list instead?
```suggestion
    [Rule || ElvisAttr <- ElvisAttrs, Rule <- ktn_code:attr(value, ElvisAttr)].
```

elbrujohalcon (changes requested):

src/elvis_core.erl
elbrujohalcon: I guess the tests will catch these things…

src/elvis_core.erl
```diff
@@ -11,6 +11,9 @@
 %% for internal use only
 -export([do_rock/2]).
 
+%% for eating what we cook
```
elbrujohalcon, line 14:
```suggestion
%% for eating our own dogfood
```
[(It's a more common idiom in English)](https://en.wikipedia.org/wiki/Eating_your_own_dog_food)

paulo-ferraz-oliveira:
How about we add `rebar_cmd` and add make `rock` into an `elvis_core` command?

Something like:
```erlang
% rebar.config
{plugins, [rebar_cmd]}.
{commands, [{rock, "rebar3 escriptize && _build/default/bin/elvis_core"}]}.
{alias, [{rock, [{cmd, "rock"}]}]}.
```

_Rocking with Elvis_ would be possible via `rebar3 rock` (and would could also simplify `.travis.yml`).

src/elvis_core.erl
paulo-ferraz-oliveira: b83efdf

src/elvis_core.erl
paulo-ferraz-oliveira: Yup.

paulo-ferraz-oliveira:
Ah, wait, `"rebar3 escriptize && _build/default/bin/elvis_core"` wouldn't work in `rebar_cmd`: there's still [a feature that needs fixing 😄](https://github.com/gootik/rebar_cmd/issues/22).  But I can plug escriptize to compile and get it done.

elbrujohalcon:
> Ah, wait, `"rebar3 escriptize && _build/default/bin/elvis_core"` wouldn't work in `rebar_cmd`: there's still [a feature that needs fixing 😄](https://github.com/gootik/rebar_cmd/issues/22). But I can plug escriptize to compile and get it done.

Can't we… `erl -s elvis_core main` instead?

paulo-ferraz-oliveira:
https://github.com/inaka/elvis_core/pull/136/commits/d110f78a6f882bc8e4535be8fd141c508600efbc: what's happening here?

1. `rebar_cmd` is used for `rebar3 cmd <command>` -like execution,
2. an `alias` is created for us to do `rebar3 rock` instead of `rebar3 cmd rock`,
3. we're hooking on post-compile since `rebar_cmd` can't (as of this moment) understand `;` or `&&` properly, in shell commands ([it'll probably be taken care of, in the near future](https://github.com/gootik/rebar_cmd/issues/22)),
4. we ran our `escript` so we eat our own dogfood!

paulo-ferraz-oliveira:
> Can't we… erl -pa $(rebar3 path) -s elvis_core main instead?

We can, but it's not intuitive, is it? Let me know what you think about this latest commit.

elbrujohalcon (changes requested):
Cool! Only the tests are missing.

paulo-ferraz-oliveira:
```erlang
{ok, #{ rules := [ #{ items := [] } ]}} = elvis_core:do_rock(File, ElvisConfig).
```
seems to indicate "no items disrespect any of the rules".

On the other hand,
```erlang
{ok, #{ rules := [] }} = elvis_core:do_rock(File, ElvisConfig).
```
seems to indicate "no rules were present", or even "no items were analyzed on existing rules". This seems odd (but perhaps I'm mis-interpreting the interface), especially when rules are analyzed but respected.

config/test.config
```diff
@@ -28,21 +28,25 @@
                   },
                   {elvis_style, state_record_and_type},
                   {elvis_style, no_spec_with_records}
-                 ]
+                 ],
+        ruleset => erl_files
```
paulo-ferraz-oliveira, line 32: I added these to improve testing targets. In any case I'll refer to this further below.

src/elvis_core.erl
```diff
@@ -121,14 +134,32 @@ apply_rules_and_print(Config, File) ->
...
--spec apply_rules(map(), File::elvis_file:file()) ->
+-spec apply_rules(elvis_config:config() | map(), File::elvis_file:file()) ->
     elvis_result:file().
 apply_rules(Config, File) ->
     Rules = elvis_config:rules(Config),
     Acc = {[], Config, File},
-    {RulesResults, _, _} = lists:foldl(fun apply_rule/2, Acc, Rules),
+    {ParseTree, _} = elvis_file:parse_tree(Config, File),
+    {RulesResults, _, _} = lists:foldl(fun apply_rule/2, Acc,
+                                       merge_rules({file, ParseTree}, lists:flatten(Rules))),
```
paulo-ferraz-oliveira, line 144: I added this `lists:flatten` otherwise we'd be "merging" a list with a list of lists, and that lead to odd results. I can expand on it, if you feel uncomfortable with this.

test/elvis_test_utils.erl
```diff
@@ -1,13 +1,19 @@
...
         , find_file/2
         ]).
 
 -spec config() -> elvis_config:config().
 config() ->
   application:get_env(elvis_core, config, []).
 
+-spec config(Ruleset :: atom()) -> elvis_config:config().
+config(Ruleset) ->
+  RulesetCfgs = application:get_env(elvis_core, config, []),
```
paulo-ferraz-oliveira, line 14: For the improved testing I mentioned above.

paulo-ferraz-oliveira:
To create the tests I did the following:
1. picked all fail modules,
2. created the tests as I would see them fail,
3. "fixed" all the modules with `-elvis(_).`,
4. rinsed and repeated until all was Okay.

paulo-ferraz-oliveira:
Tests are failing remotely, not locally. I'm going to try and compile the same `rebar3` version (that Travis is using) to see if I can replicate it. If so, this means something changed and will probably degrade testing when we get to remote 3.14.1 (the one I'm using locally).

config/test.config
```diff
@@ -28,21 +28,25 @@
...
         filter => "rebar.config",
         rules => [{elvis_project, no_deps_master_rebar, #{ignore => []}},
-                  {elvis_project, protocol_for_deps_rebar, #{ignore => []}}]
+                  {elvis_project, protocol_for_deps_rebar, #{ignore => []}}],
+        ruleset => rebar_config
        },
       #{dirs => ["../../"],
         filter => "elvis.config",
-        rules => [{elvis_project, old_configuration_format}]
+        rules => [{elvis_project, old_configuration_format}],
```
elbrujohalcon, line 48: Isn't this rule already included in the ruleset?

src/elvis_core.erl
elbrujohalcon: No idea why `elvis_config:rules/1` returns a list of lists, but I trust you.

src/elvis_core.erl
```diff
@@ -121,14 +134,32 @@ apply_rules_and_print(Config, File) ->
...
     elvis_result:new(file, File, RulesResults).
 
+merge_rules({file, ParseTree}, ElvisConfigRules) ->
+    ElvisAttrs = elvis_code:find(fun is_elvis_attr/1, ParseTree,
+                                 #{ traverse => content, mode => node }),
+    ElvisAttrRules = elvis_attr_rules(ElvisAttrs),
+    merge_rules(ElvisAttrRules, ElvisConfigRules);
+merge_rules(ElvisAttrRules, ElvisConfigRules) ->
+    elvis_config:merge_rules(ElvisAttrRules, ElvisConfigRules).
```
elbrujohalcon, line 153:
```suggestion
    elvis_config:merge_rules(ElvisAttrRules, ElvisConfigRules).
```
Can't we just avoid the intermediate step? Or is there some sort of recursion I'm not seeing?

test/elvis_test_utils.erl
```diff
@@ -1,13 +1,19 @@
...
+  [Cfg || Cfg <- RulesetCfgs, maps:find(ruleset, Cfg) =:= {ok, Ruleset}].
```
elbrujohalcon, line 15:
```suggestion
  [Cfg || #{ruleset := R} = Cfg <- RulesetCfgs, R =:= Ruleset].
```
I think this should be equivalent to your expression and I think it's a bit clearer.

elbrujohalcon (changes requested):
A few tiny stylish things only.
I ❤️  the new way of testing… it looks very similar to what we have in `rebar3_format`.

src/elvis_core.erl
paulo-ferraz-oliveira: Well, because `rules` is a list of maps 😄 and the maps get turned into lists (thus list of lists). Don't trust me, but the tests. If possible we should get a third opinion, no?

config/test.config
paulo-ferraz-oliveira: I only added the ruleset element to those maps. I didn't touch anything else, since I didn't "need" to. Should I remove the "by default" elements?

src/elvis_core.erl
paulo-ferraz-oliveira: Probably, there's no recursion you're not seeing 😄. The original code [had more function clauses](https://github.com/inaka/elvis_core/pull/136/commits/98a896f4f8797bbf020141b9adbc30eb1015edc3#diff-0e321e7ef4fb27c145572ae07f7ebeacR134), I kept removing them, and simplifying code (as I tested) and got to this. Let me check.

paulo-ferraz-oliveira: I wonder if this is one of those cases the pre-compiler tools optimizes. 🤔

paulo-ferraz-oliveira: dbc0357

paulo-ferraz-oliveira:
**Unrelated**: I increasingly feel the need to implement `ignore` as a generic option for all rules. It'll probably be the one I'll pick next.

config/test.config
elbrujohalcon: yes, please.

paulo-ferraz-oliveira: f269a08

elbrujohalcon (approved):

paulo-ferraz-oliveira:
❤️ thanks much.

PR #137, opened 2020-10-02
Feature option ignore in all rules

paulo-ferraz-oliveira:
[Fix inaka/elvis#507]
[Fix inaka/elvis#488]

As per previous discussions, this initial implementation step can be seen as a request for comments. I will also comment the code on some decisions and what I'd like to improve and/or require help.

Possible documentation (also up for review) is found at inaka/elvis#540.


paulo-ferraz-oliveira:
So, my naive approach to this problem was:
1. remove `ignore` from the default options list of every rule (the less the merrier) since it'll be handled as a special case,
2. create type `ignorable()` to hold the information at the center of this change,
3. strip all rules of the currently implemented behavior of handling `ignore` on a per rule basis,
4. dig deeper into where we can best apply the filter so changes after that are minimal (ended up being `elvis_file: resolve_parse_tree`).

**Note**: only one test had to be touched, because one of the options disappeared.

src/elvis_file.erl
```diff
@@ -111,11 +121,32 @@ filter_files(Files, Dirs, Filter, IgnoreList) ->
 %% Private
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
--spec resolve_parse_tree(string(), binary()) ->
+-spec resolve_parse_tree(boolean(), string(), binary(), module(), list()) ->
     undefined | ktn_code:tree_node().
-resolve_parse_tree(".erl", Content) ->
-    ktn_code:parse_tree(Content);
-resolve_parse_tree( _, _) ->
+resolve_parse_tree(false = _ModIsIgnored, ".erl", Content, Mod, Ignore) ->
```
paulo-ferraz-oliveira, line 126: This seemed like the appropriate place to centralize the filtering capabilities, since I didn't want to have to keep calling a generic function `is_ignored(...)`, inside every rule, with which I'd generate a new case clause (as was).

src/elvis_file.erl
```diff
@@ -111,11 +121,32 @@ filter_files(Files, Dirs, Filter, IgnoreList) ->
...
+                                       }
+                           }) ->
+                             not(lists:member({Mod, FunName}, Ignore)
+                                 orelse lists:member({Mod, FunName, FunArity}, Ignore));
+                         (_) -> true
+        end,
+        TreeContent),
+    Tree#{ content => FilteredTreeContent };
+resolve_parse_tree(true = _ModIsIgnored, ".erl", _Content, Mod, _Ignore) ->
+    #{ type => root
```
paulo-ferraz-oliveira, line 142: Hurrah for non-opaque structures. In any case, I did try with `undefined`, `ignored`, ... and tweaking code affected by this change, but the way it is stuff is simply too tight. Does this seem completely inadequate?

src/elvis_style.erl
```diff
@@ -604,7 +542,7 @@ module_naming_convention(Config, Target, RuleConfig) ->
     Regex = option(regex, RuleConfig, module_naming_convention),
     IgnoreModules = option(ignore, RuleConfig, module_naming_convention),
```
paulo-ferraz-oliveira, line 543: This is the only `ignore` I left in `elvis_style` and for a somewhat good reason: it doesn't parse the tree to apply a rule (only to get a module name). Let me know if there is a better alternative.

src/elvis_file.erl
elbrujohalcon: I like your approach. 👍

src/elvis_file.erl
elbrujohalcon: So… you're returning an empty AST here so that the rules have nothing to review? It's clever, but please add a comment explaining the reasoning.

elbrujohalcon (commented):
I would've naively added the code to ignore modules to `elvis_core:apply_rule/2` instead of adding it to every rule, but I understand why you did it this way (i.e. to support specific function ignores).
Nevertheless, maybe you want to consider moving the ignoring code for modules to `elvis_core:apply_rule/2`, and then filtering the `ignore` list in the options to remove the module names before passing it in to the application of every rule.

src/elvis_file.erl
paulo-ferraz-oliveira: I'll add the comment with the reasoning, but I disagree it's clever 😄; I find it rather dirty, but it serves the purpose at hand (at least until `ktn` decides to change the structure - in which case many other things in `elvis_core` would also have to be reviewed).

paulo-ferraz-oliveira:
> I would've naively added the code to ignore modules to `elvis_core:apply_rule/2` instead of adding it to every rule, but I understand why you did it this way (i.e. to support specific function ignores).
> Nevertheless, maybe you want to consider moving the ignoring code for modules to `elvis_core:apply_rule/2`, and then filtering the `ignore` list in the options to remove the module names before passing it in to the application of every rule.

Sure, I can give this a try. Thanks.

src/elvis_file.erl
paulo-ferraz-oliveira: a52fc4e

src/elvis_style.erl
paulo-ferraz-oliveira: @elbrujohalcon: I see that `line_length`, `no_spaces`, `no_tabs`, and `no_trailing_whitespace` are also not "under" my change, since they're (I consider them, at least) more style oriented, in a "project"-oriented analysis.

Should I:
1. not care that `ignore` doesn't work for these, as is now (in the PR)?
2. make `ignore` work for this on a module-level, as probably another comment will take care of?
3. move these to `{elvis_project, _, _}` since they seem more high-level than other rules?

src/elvis_core.erl
```diff
@@ -13,6 +13,7 @@
 
 %% for eating our own dogfood
 -export([main/1]).
+-export([apply_rule/2]).
```
paulo-ferraz-oliveira, line 16: Exposed, since now tests are using it because ignore is no longer at the rule execution -level, but higher up.

elbrujohalcon: Does it work if it's placed between `ifdef`, i.e.…

```erlang
-ifdef(TEST).
-export [apply_rule/2].
-endif.
```

src/elvis_style.erl
elbrujohalcon: I, as a user, would expect `ignore` directives to be respected for those rules, too.
It would not surprise me to find people wanting to ignore the `no_trailing_whitespace` rule for certain auto-generated modules, for instance.

elbrujohalcon: I understand that it's not easy to implement that since those rules don't parse the tree and, in general, when we apply them we don't really know where we're standing (i.e. which function are we in).
I think a somewhat reasonable approach would be to…

1. Move all these rules to a different module (`elvis_text_style`, maybe?).
2. Wherever we said that _every rule has an `ignore => ignorable()` config_, amend the text to correctly state that `{M, F} and {M, F, A}` are only valid for  _rules **in `elvis_style`** and not in other rule modules_.

…or maybe…

2. Split the config params into `ignore => [module()], ignore_functions => [ignorable()]` and make this second one `elvis_style`-exclusive.

I think that would be safe and it would also prevent people from assuming that other modules (outside `elvis_style`) will support `ignore`.

elbrujohalcon (changes requested):

src/elvis_core.erl
paulo-ferraz-oliveira: It should.

paulo-ferraz-oliveira: 711b970

paulo-ferraz-oliveira:
> Move all these rules to a different module (elvis_text_style, maybe?).

Just noting this IS a breaking change. But it's the one I'll go with.

elbrujohalcon:
> > Move all these rules to a different module (elvis_text_style, maybe?).
> 
> Just noting this IS a breaking change. But it's the one I'll go with.

Yes, it is. It will require a nice jump in minor version. Maybe mayor version, too? 🤔  SemVer is tricky…

paulo-ferraz-oliveira:
> > > Move all these rules to a different module (elvis_text_style, maybe?).
> > 
> > 
> > Just noting this IS a breaking change. But it's the one I'll go with.
> 
> Yes, it is. It will require a nice jump in minor version. Maybe mayor version, too? 🤔 SemVer is tricky…

It's not that tricky :D

Major = interface broke
Minor = new feature added (compatible with previous version)
Patch = bug fix (compatible with previous version)

You initially need to decide what is part of your application's interface, though. (e.g. "internal" functions might be exposed in Erlang modules, but not be part of the interface).

In `elvis_core`'s case, for example:
1. the API is part the interface,
2. the `elvis.config` file is part of the interface, and, to a certain extent,
3. the output is part of the interface.

You can also look at it this way: if somebody imports my lib. will they have to change anything to make it work properly with this new version? If so, then the interface was broken; major bump.

elbrujohalcon:
Cool… so a major bump should be.
OTOH we can do it more slowly… we can keep the functions on `elvis_style`, with something like…

```erlang
-deprecated({no_tabs, 3, next_major_release}).
-spec no_tabs(elvis_config:config(), elvis_file:file(), empty_rule_config()) -> [elvis_result:item()].
no_tabs(Config, Target, RuleConfig) ->
  elvis_text_style(Config, Target, RuleConfig).
```

Then we remove it on the next major release.
I'm not sure if it's worth the hassle, but it may reduce the number of angry users if there is any.

paulo-ferraz-oliveira:
> if there is any.

Users? Or angry users? 🤣 

I can break the interface, no problem. It's just a question of moving your function name, in any case. (what I usually do is have a `MIGRATION.md` file with instructions on how to migrate when breaking changes occur)

elbrujohalcon:
> > if there is any.
> 
> Users? Or angry users? 🤣
> 

Yes. :trollface: 

> I can break the interface, no problem. It's just a question of moving your function name, in any case. (what I usually do is have a `MIGRATION.md` file with instructions on how to migrate when breaking changes occur)

`MIGRATION.md` is a great idea!!
Let's break the interface, add `MIGRATION.md`, and move to `1.0.0` for once! It's, in fact, something we should've done a long long time ago anyway.

MIGRATION.md
```diff
@@ -0,0 +1,22 @@
...
+since these are incremental.
+
+This file's format is influenced by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
+[Make a README](https://www.makeareadme.com/).
+
+## Going from `0.7.0` to `1.0.0`
+
+### Update
+
+- your `ignore` option (from rule `max_function_length`) to also contain the content of
```
elbrujohalcon, line 15:
```suggestion
- your `ignore` option (e.g. from rule `max_function_length`) to also contain the content of
```

The rule was just an example, right?

MIGRATION.md
```diff
@@ -0,0 +1,22 @@
...
+`ignore_functions`
+- your rule config. files (e.g. `elvis.config`), by moving `elvis_style` to `elvis_text_style` for
```
elbrujohalcon, line 17:
```suggestion
- your rule config files (e.g. `elvis.config`), by moving `elvis_style` to `elvis_text_style` for
```

MIGRATION.md
elbrujohalcon, line 15:
```suggestion
- your `ignore` option from rule `max_function_length` to also contain the content of
```

elbrujohalcon (changes requested):
Just some tiny tweaks to text.

MIGRATION.md
paulo-ferraz-oliveira: I'm going to use "configuration" then. "config"'s not a word, which is why I used the . for abbreviation.

paulo-ferraz-oliveira: 8afb76f

MIGRATION.md
paulo-ferraz-oliveira: 8afb76f

MIGRATION.md
paulo-ferraz-oliveira: I also update 0.7.0 to 0.x and 1.0.0 to 1.x, since we don't actually know the departure and arrival version for the consumer, so it's safer to say "if you're bumping into 1.x, from 0.x", since 2.x will surely involve a new `MIGRATION.md` round of updates.

elbrujohalcon: I like your ideas and your strictness :P

paulo-ferraz-oliveira:
@elbrujohalcon, anything missing?

PR #139, opened 2020-11-20
Accept British spelling for behaviour.

NAR:
Erlang accepts the `-behaviour` spelling too, so when checking for OTP behaviours, check this spelling.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1102,7 +1102,11 @@ is_otp_module(Root) ->
                              gen_statem,
                              supervisor_bridge
                             ]),
-    IsBehaviorAttr = fun(Node) -> behavior == ktn_code:type(Node) end,
+    IsBehaviorAttr =
+        fun(Node) ->
+            behavior == ktn_code:type(Node) orelse
+            behaviour == ktn_code:type(Node)
+        end,
     case elvis_code:find(IsBehaviorAttr, Root) of
         [] ->
             false;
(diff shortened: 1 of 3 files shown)


src/elvis_style.erl
```diff
@@ -1102,7 +1102,11 @@ is_otp_module(Root) ->
                              gen_statem,
                              supervisor_bridge
                             ]),
-    IsBehaviorAttr = fun(Node) -> behavior == ktn_code:type(Node) end,
```
paulo-ferraz-oliveira, line 1105: I'd like @jfacorro's and/or @elbrujohalcon's feedback here. I believe this change should be done at the `ktn_code` level. The **type** should be independent of the spelling (lest other stuff that depends on `ktn_code` suffer from the same and we keep adding this workaround over and over). What say you?

paulo-ferraz-oliveira: :( neither `behaviour` nor `behavior` are "expressed" at the `ktn_code` level.

elbrujohalcon: It goes way deeper… I think it's fine to be able to distinguish it. We might even have a rule against one or the other spelling one day…

elbrujohalcon (approved):

elbrujohalcon:
Thanks, @NAR :)

src/elvis_style.erl
paulo-ferraz-oliveira: What I was referring to is that `:type` [already specifies many things](https://github.com/inaka/katana-code/blob/master/src/ktn_code.erl#L43) (e.g. `root`), but not `behavior` or `behaviour`.

elbrujohalcon: Oh! Nice catch, @paulo-ferraz-oliveira! Do you mind opening a ticket in that project to solve that?

paulo-ferraz-oliveira: Not at all.

PR #141, opened 2020-11-25
Allow for custom rules in elvis configuration

cnasten:
If using multiple RuleGroups with custom made rules, instead of having
to duplicate the rules support a custom rules section of the configuration
that can be referenced in the rules argument:

```erlang
[
 {
  elvis,
  [
   {rules,
    #{project => [{elvis_style, no_tabs}]}
   },
   {config,
    #{rules => {custom_rules, project}}
   }
  ]
 }
].
```

### DIFF
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -49,6 +49,8 @@ from_file(Path) ->
 -spec load(atom(), term()) -> config().
 load(Key, AppConfig) ->
     ElvisConfig = proplists:get_value(Key, AppConfig, []),
+    Rulesets = proplists:get_value(rulesets, ElvisConfig, #{}),
+    register_rulesets(Rulesets),
     Config =  proplists:get_value(config, ElvisConfig, []),
     ensure_config_list(Config).
 
@@ -56,6 +58,11 @@ ensure_config_list(Config) when is_map(Config) ->
     [Config];
 ensure_config_list(Config) ->
     Config.
+
+register_rulesets(Rulesets) ->
+    lists:foreach(fun({Name, Rules}) ->
+                          elvis_rulesets:register_ruleset(Name, Rules)
+                  end, maps:to_list(Rulesets)).
 
 -spec validate(Config::config()) -> ok.
 validate([]) ->
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -50,7 +50,7 @@ from_file(Path) ->
 load(Key, AppConfig) ->
     ElvisConfig = proplists:get_value(Key, AppConfig, []),
     Rulesets = proplists:get_value(rulesets, ElvisConfig, #{}),
-    register_rulesets(Rulesets),
+    elvis_rulesets:set_rulesets(Rulesets),
     Config =  proplists:get_value(config, ElvisConfig, []),
     ensure_config_list(Config).
 
@@ -58,11 +58,6 @@ ensure_config_list(Config) when is_map(Config) ->
     [Config];
 ensure_config_list(Config) ->
     Config.
-
-register_rulesets(Rulesets) ->
-    lists:foreach(fun({Name, Rules}) ->
-                          elvis_rulesets:register_ruleset(Name, Rules)
-                  end, maps:to_list(Rulesets)).
 
 -spec validate(Config::config()) -> ok.
 validate([]) ->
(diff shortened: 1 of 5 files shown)


config/elvis-test-custom-rules.config
```diff
@@ -0,0 +1,15 @@
+[
+ {
+  elvis,
+  [
+   {rules,
```
elbrujohalcon, line 5: I would call these `rulesets`.

config/elvis-test-custom-rules.config
```diff
@@ -0,0 +1,15 @@
...
+    #{project => [{elvis_style, no_tabs}]}
+   },
+   {config,
+    [#{dirs => ["."],
+       filter => "*.erl",
+       rules => {custom_rules, project}}]
```
elbrujohalcon, line 11: I would not use `rules` for this. I would instead use the already existing `ruleset` key.

elbrujohalcon: And I would not require the tuple, the atom alone should do (as long as it doesn't collide with the existing rulesets. Or we can even choose to allow users to override them.

elbrujohalcon (changes requested):
Please consider using the existing rulesets infrastructure for this as much as possible.
Thanks.

config/elvis-test-custom-rules.config
cnasten: 👍

config/elvis-test-custom-rules.config
cnasten: What do you think about the updated version? More to your liking or had you something else in mind? :)

src/elvis_rulesets.erl
```diff
@@ -1,6 +1,14 @@
 -module(elvis_rulesets).
 
--export([rules/1]).
+-export([rules/1,
+         register_ruleset/2]).
+
+-define(RULESET_TABLE, elvis_custom_rulesets).
```
elbrujohalcon, line 6: Don't use a macro, just use the atom everywhere. It should be fine.

elbrujohalcon: Or alternatively, you can use `?MODULE` as the table name.

src/elvis_rulesets.erl
```diff
@@ -53,4 +61,17 @@ rules(elvis_config) ->
         end,
         [ old_configuration_format
         ]);
-rules(_Group) -> [].
+rules(Group) ->
+    try ets:lookup_element(?RULESET_TABLE, Group, 2) of
+        Rules -> Rules
+    catch
+        error:badarg -> []
+    end.
```
elbrujohalcon, line 69:
```suggestion
    try ets:lookup_element(?RULESET_TABLE, Group, 2)
    catch
        error:badarg -> []
    end.
```

src/elvis_config.erl
```diff
@@ -57,6 +59,11 @@ ensure_config_list(Config) when is_map(Config) ->
 ensure_config_list(Config) ->
     Config.
 
+register_rulesets(Rulesets) ->
+    lists:foreach(fun({Name, Rules}) ->
```
elbrujohalcon, line 63: Should we drop and recreate the table here (with something like `elvis_rulesets:clean()` or something)?
Just in case someone runs `elvis_core:rock(…)` multiple times with different configurations on each.

elbrujohalcon (changes requested):
Just a few mostly cosmetic changes.

elbrujohalcon (approved):

elbrujohalcon:
**Thank you, @cnasten!**

Do you mind adding documentation about this new feature to [the wiki](https://github.com/inaka/elvis_core/wiki/Rules).

It would be awesome if you can actually add a whole new section about _Rulesets_, briefly explaining that there are some predefined ones but you can also define your own ones… etc…

paulo-ferraz-oliveira:
@elbrujohalcon, I don't see mention of custom rules, but only custom rulesets. Do we actually support custom rules (as in module+function+def, and centralised execution)?

elbrujohalcon:
Short answer, since I'm on the phone: Yes, you add whatever you want to the `rules` list. You can put your own rules there.

paulo-ferraz-oliveira:
Ok, seems undocumented, though, but I'm probably missing something.

paulo-ferraz-oliveira:
Sure, but since I'm trying to validate the config. I'm sure validating that at least custom rules are 2-tuple/3-tuple and refer to existing module:function would be appreciated. And that's what I'm aiming for.

PR #169, opened 2021-01-12
Issue warnings on empty folders

### LINKED ISSUE #146: Warn on empty analysis
When executing analysis on a given folder reference (origin: `elvis.config` > (inexisting)`dirs => ["text/examples"]` instead of (existing)`dirs => ["test/examples"]`), if no elements are found in a given examined path, a warning should be output, so we know we're dealing with an exception. (it also helps keep our `elvis.config` as trimmed as possible).

paulo-ferraz-oliveira:
Closes #146.

Since this is a warning, not a hard error, and there's no output (except for the visible message), I couldn't think of a nice enough test. Suggestions?

**Note**: I add the possibility to write warnings, in magenta, since I want to distinguish these from errors.

### DIFF
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -158,6 +158,14 @@ resolve_files(RuleGroup, Files) ->
     Dirs = dirs(RuleGroup),
     Ignore = ignore(RuleGroup),
     FilteredFiles = elvis_file:filter_files(Files, Dirs, Filter, Ignore),
+    _ = case FilteredFiles of
+            [] ->
+                Error = elvis_result:new(warn, "Searching for files in ~p yielded none. "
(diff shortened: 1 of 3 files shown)


elbrujohalcon:
I don't think we have tests in place for any of these things (I mean… just with a quick grep… I couldn't find any test for error printing either). Maybe we test these things indirectly from `elvis`? 🤔 

I'm fine if you don't want to spend more time trying too hard to test this. If you're already done with this, I'll go ahead and merge it as-is.

paulo-ferraz-oliveira:
I'd appreciate it if you'd at least try it out and tell me if the UX is good. Other than that, I'm OK for merge.

PR #175, opened 2021-02-08
Fix unwarranted match as found by OTP 24

paulo-ferraz-oliveira:
As introduced by https://github.com/erlang/otp/pull/2995. Fun times!

(also, I wrote the original code, and it would be strange for me to not have done that "match" in error)

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -879,7 +879,7 @@ macro_as_atom({var, _Text, MacroAsAtom}, _Types, _MacroNodeValue) ->
     MacroAsAtom;
 macro_as_atom({atom, _Text, MacroAsAtom}, _Types, _MacroNodeValue) ->
     MacroAsAtom;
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
A release of this inside `rebar3_lint` would be nice. Shall I? (I'm not actually sure about the whole routine you apply before a release).

elbrujohalcon:
Awesome findings!! Thanks, @richcarl :)

elbrujohalcon:
> A release of this inside `rebar3_lint` would be nice. Shall I? (I'm not actually sure about the whole routine you apply before a release).

We first need a release of `elvis_core`… Then you can send a PR against `rebar3_lint` and then… you're an admin there… you can release it at your own will.

elbrujohalcon:
https://hex.pm/packages/elvis_core/1.1.1

paulo-ferraz-oliveira:
Thanks.

Also, yeah, this new warning is nice.

PR #176, opened 2021-02-08
Ease consumption of warning

paulo-ferraz-oliveira:
Found this while pulling 1.1.1 into `rebar3_lint`. Since this is a minor detail I guess we can leave it here, in `master`, until we need a new release.

Also, mind you that this message might need to be further improved, in the future, if an even more complicated setup (`elvis.config`) is done.


elbrujohalcon:
Yeah, let's just keep it in master.

elbrujohalcon (approved):

elbrujohalcon (approved):

PR #179, opened 2021-03-10
Fix analysis over umbrella apps

### LINKED ISSUE #178: Analysis not working over umbrella apps
As per https://github.com/project-fifo/rebar3_lint/issues/26#issue-826719550.

paulo-ferraz-oliveira:
Closes #178.

- [x] make sure tests are passing in Windows
- [x] make sure to add tests for updated functionality (namely we should test folder structure as per umbrella apps)

---
**Note**: this includes a small patch to run the tests on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,29 +30,3 @@ jobs:
       - run: |
           rebar3 escriptize
           _build/default/bin/elvis_core
-  ci-windows:
-    name: >
-      Run Windows-based checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
-    runs-on: ${{matrix.os}}
-    strategy:
-      matrix:
-        otp_vsn: [23.2]
-        os: [windows-latest]
-    steps:
-    - uses: actions/checkout@v2
-    - uses: gleam-lang/setup-erlang@v1.1.2
-      with:
-        otp-version: ${{matrix.otp_vsn}}
-      id: install_erlang
-    - run: wget https://s3.amazonaws.com/rebar3/rebar3 -OutFile rebar3
-      shell: powershell
-    - run: |
-        & "${{steps.install_erlang.outputs.erlpath}}\bin\escript.exe" rebar3 dialyzer
-    - run: |
-        & "${{steps.install_erlang.outputs.erlpath}}\bin\escript.exe" rebar3 ct
-    - run: |
-        & "${{steps.install_erlang.outputs.erlpath}}\bin\escript.exe" rebar3 cover
-    - run: |
-        & "${{steps.install_erlang.outputs.erlpath}}\bin\escript.exe" rebar3 escriptize
-    - run: |
-        & "_build\default\bin\elvis_core"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,15 +9,16 @@ on:
       - master
 jobs:
   ci:
-    name: Run checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
+    name: >
+      Run Linux-based checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
         otp_vsn: [19.3.6.13, 20.3.8.26, 21.3.8.17, 22.3.4.9, 23.1.5]
-        os: [ubuntu-latest]
+        os: [ubuntu-18.04]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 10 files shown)


src/elvis_file.erl
```diff
@@ -86,37 +86,42 @@ find_files(Dirs, Pattern) ->
...
+                                   DirOrFile =:= Path
+                               end,
+                               ExpandedFilter)
+                 end,
+                 Files).
+
 %% @doc Filter files based on the glob provided.
 -spec filter_files([file()], [string()], string(), [string()]) -> [file()].
 filter_files(Files, Dirs, Filter, IgnoreList) ->
-    AppendFilter = fun(Dir) ->
```
paulo-ferraz-oliveira, line 92: There's a fix here, and a bit of refactoring.

The refactoring is basically to try and improve readability (no more functions declared as variables, for example).

The fix is to have analysis work over umbrella projects, where input `dirs` might be `apps/**/src`.

Considerations on the algo:
1. we still create a list of filters (`dir`+`filter`) that we expand with `wildcard`, (`ExpandedFilters`)
2. we go over a list of previously found files to check if they "fall" in the filters, (`Found`)
3. we remove the `ignore` element from the previous list (function result).

We do this without resorting to regex, which was not only confusing, as-was, but error prone (`**` not accepted).

Now, the code for `Found` seems oddly similar to that of function `find_files/2` further above, so some more refactoring might be in order.

.github/workflows/ci.yml
```diff
@@ -29,3 +30,29 @@ jobs:
       - run: |
           rebar3 escriptize
           _build/default/bin/elvis_core
+  ci-windows:
```
paulo-ferraz-oliveira, line 33: Let's hope I can get this working without many issues

.github/workflows/ci.yml
```diff
@@ -9,15 +9,16 @@ on:
...
   ci:
-    name: Run checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
+    name: >
+      Run Linux-based checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
         otp_vsn: [19.3.6.13, 20.3.8.26, 21.3.8.17, 22.3.4.9, 23.1.5]
-        os: [ubuntu-latest]
+        os: [ubuntu-18.04]
```
paulo-ferraz-oliveira, line 18: This break due to wx -related stuff being renamed.

elbrujohalcon:
I like it… let's see if the windows-ci starts working at some point.

src/elvis_file.erl
```diff
@@ -86,37 +86,46 @@ find_files(Dirs, Pattern) ->
...
+    Filter;
+dir_to(Filter, Dir) ->
+    filename:join(Dir, Filter).
+
+file_in(ExpandedFilter, Files) ->
+    lists:filter(fun (_File = #{ path := Path }) ->
+                     lists:any(fun (DirOrFile) ->
+                                   DirOrFile =:= Path
+                               end,
+                               ExpandedFilter)
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: It was soooo late in the night. (also, I don't know why I was under the impression `lists:member` did a full list scan, which would make no sense, but still...).

paulo-ferraz-oliveira:
@elbrujohalcon, I moved the tasks from the issue here (check the description) so there's visibility. I still want to had tests lest we end up in the same situation.

paulo-ferraz-oliveira:
@elbrujohalcon: pushed some tests. I'm not sure we need something more elaborate. It walks an expected (and advertised, in `rebar3_lint` folder structure) using regex.
Should we probably throw in a `ignore` test?

paulo-ferraz-oliveira:
I segregated the Windows CI thing to an Issue, lest we're blocked because of it. I'd wrongly assumed stuff was already working and would have only to deal with the container stuff, but it seems not.

paulo-ferraz-oliveira:
@joaohf, do you want to try these changes out to see if they fix your specific issue?

joaohf:
@paulo-ferraz-oliveira sure, I'll try today.

elbrujohalcon (approved):
I like how this looks.

joaohf:
Alright, the fix works for me. Really good. Thank you guys.

paulo-ferraz-oliveira:
@joaohf, is it acceptable to rely on `master` for the time being, before we cut a release? I don't expect this to break for anyone relying on `master`, but the next few days will tell 😄

joaohf:
Sure, don't worry. I can rely on master until there.

PR #181, opened 2021-03-10
Fix a crash happening when a macro is also an atom (as a function)

paulo-ferraz-oliveira:
@elbrujohalcon: remember how you love macros? 😄 

This was crashing (prior to this pull request)...

```
The macro named "debug" on line 
```


elbrujohalcon:
Have I already manifested my profound dislike for macros? 😡

paulo-ferraz-oliveira:
Never 😄

PR #184, opened 2021-03-13
Prevent crash for operators being analyzed at the left-most line position

### LINKED ISSUE #183: -> at the beginning of the line makes the linter crash
The following code:

```
-spec wait_for_fun(term(), non_neg_integer(), non_neg_integer())
-> {ok, any()} | ok | timeout.
```

Causes the linter to fail with:

```
# apps/els_dap/test/els_dap_test_utils.erl [FAIL]
Error: 'function_clause' while applying rule 'operator_spaces'.
```

Moving the `->` away from the beginning of the line solves the issue.

paulo-ferraz-oliveira:
Closes #183.

cc: @robertoaloi


elbrujohalcon (approved):
Great

PR #185, opened 2021-03-15
Further adapt to OTP 24 while bumping some dep.s

paulo-ferraz-oliveira:
I'll still test this as a `_checkout` in `elvis` and `rebar3_lint`, but this is a move forward.


paulo-ferraz-oliveira:
Still working out some warnings on OTP 24.

paulo-ferraz-oliveira:
So here's a difference:
* OTP 23
```erlang
#{ attrs => #{
       location => {23, 5},
       name => "module_name",
       text => "?module_name"
   },
   content => [],
   type => macro }
```
* OTP 24
```erlang
#{ attrs => #{
       location => {23, 5},
       text => "?"
   },
   content => [],
   node_attrs => #{
       function=> #{
           attrs => #{
               location => {23, 5},
               text => "?"
           },
           node_attrs => #{
               function => #{
                   attrs => #{
                       location => {23, 18},
                       name => "function_name",
                       text => "?function_name"
                   },
                   content => [],
                   type => macro
               },
               module => #{
                   attrs => #{
                       location => {23, 5},
                       name => "module_name",
                       text => "?module_name"
                   },
                   content => [],
                   type => macro
               }
           },
           type => remote
       }
   },
   type => call
 }
```

(that's always fun)

paulo-ferraz-oliveira:
@elbrujohalcon, @jfacorro ^ does that seem more `elvis_core`-related or `katana-code`-related?

paulo-ferraz-oliveira:
I'm thinking more `elvis_core`, since this is interpreted different: `?module_name:?function_name()`.
In OTP 23 it's a macro followed by stuff, in OTP 24 it's a function.

elvis.config
```diff
@@ -18,10 +18,6 @@
                   {elvis_style, atom_naming_convention, #{ignore => [elvis_task]}}],
         ruleset => beam_files
        },
-      #{dirs => ["."],
```
paulo-ferraz-oliveira, line 21: Unused, so not need for the lib. itself. (also, `elvis_core` warns about this 😄).

rebar.config
```diff
@@ -23,11 +23,11 @@
...
     {extra_src_dirs, ["test/examples"]},
-    {deps, [ {mixer,       "1.0.1", {pkg, inaka_mixer}}
-           , {meck,        "0.9.0"}
-           , {xref_runner, "1.1.1"}
+    {deps, [ {mixer,       "1.1.1", {pkg, inaka_mixer}}
+           , {meck,        "0.9.2"}
+           , {xref_runner, "1.2.0"}
     ]},
-    {erl_opts, [ nowarn_missing_spec ]},
+    {erl_opts, [ nowarn_missing_spec, nowarn_export_all ]},
```
paulo-ferraz-oliveira, line 30: This avoid warnings, but in reality fixes no real issue apart from that.

test/style_SUITE.erl
```diff
@@ -163,7 +163,7 @@ verify_function_naming_convention(Config) ->
     PathFail = "fail_function_naming_convention." ++ Ext,
 
     [_CamelCaseError, _ALL_CAPSError, _InitialCapError,
-     _HyphenError, _PredError, _EmailError] =
```
paulo-ferraz-oliveira, line 166: This is OTP 24 complaining, and it's right 😄.

src/elvis_style.erl
```diff
@@ -953,6 +953,8 @@ has_remote_call_parent(Zipper) ->
     case ktn_code:type(Node) of
         remote ->
             true;
+        call ->
```
paulo-ferraz-oliveira, line 956: I started this with a `has_remote_call_parent(zipper:down(Zipper))` implementation, but was afraid we could fall into infinite recursion `up`/`down`/`up`/`down`/`up`/`down`/... (this fixes the tests for OTP 24 and also runs on previous versions, but I'm not exactly 💯 on this change).

test/style_SUITE.erl
elbrujohalcon: I think the original idea was not for those two errors to match, just to ignore both.
In other words, I would've replaced the other one by `_EmailError2` or something like that.

elbrujohalcon (approved):
No idea about the `zipper` parts… but the PR is fine in general.

test/style_SUITE.erl
paulo-ferraz-oliveira: Sure, I can do that (which is why I think the warning is very nice), but it's working in any case 😄

paulo-ferraz-oliveira: 23709ab

Issue #172, opened 2021-01-28
Consider more default operator spaces


paulo-ferraz-oliveira:
(don't want to get into an argument about what is an operator or not 😄)

Should we consider defaults for these?

```erlang
=
+
-
*
/
%
<
=<
>
>=
==
=:=
/=
=/=
--
=>
:=
<-
<=
||
|
::
->
```

elbrujohalcon:
I like it! I would include all of them in the default, except for `)` and `(`. Those sometimes may require spaces around them, yet sometimes not (I think).

paulo-ferraz-oliveira:
Cool. I've update the previous list by removing `(` and `)`. It's always a choice to add them in the future, I guess (either as default or on the consumer).

paulo-ferraz-oliveira:
Linked to #234, it seems.

elbrujohalcon:
Also, `%` is not an operator.

paulo-ferraz-oliveira:
> Also, `%` is not an operator.

I wanted to make sure you were paying attention 🤣 (seriously, though, I don't know where that one came from).

PR #187, opened 2021-03-21
allow checking spaces around | and ||

z8674558:


elbrujohalcon (approved):
Niiice!

paulo-ferraz-oliveira:
We should have added tests for when those operators are at line-start and at line-end. It was what was missing before with `->` and was left undetected. I'll probably do a more comprehensive tests once I get around to https://github.com/inaka/elvis_core/issues/172. 👍

PR #195, opened 2021-05-14
Increase consumer confidence

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
@elbrujohalcon, I don't exactly understand why CI's stuck, but I think I've seen this before and you might have to tweak some settings or something.

elbrujohalcon:
Yeah… I had to remove branch protections since that part of the UI is now broken in github 😢

PR #198, opened 2021-07-06
elvis.config options for application:get_env/2,3

### LINKED ISSUE #197: Fetch more options from elvis.config + rebar.config
At the moment, `elvis_core` depends on `application:get_env` which is not very usable for the consumer, especially if his/her project is not bound to a release.

The purpose of this task is to, while keeping backward compatibility, fetch further options from `elvis.config` + `rebar.config` for usability.

**Note**: search the code for `application:`.

paulo-ferraz-oliveira:
Closes #197.

~Opening as draft so we can discuss further.~

I committed in three different moments so the refactoring is easier to follow. Can squash before merge...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: erlef/setup-beam@v1
         with:
           otp-version: ${{matrix.otp_vsn}}
-          rebar3-version: '3.14'
+          rebar3-version: '3.15'
       - run: rebar3 dialyzer
       - run: rebar3 ct
       - run: rebar3 cover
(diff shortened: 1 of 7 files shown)


elbrujohalcon:
Looks good so far…

paulo-ferraz-oliveira:
Moving away from draft. Whatever happens now is "needs a fix".

elbrujohalcon (changes requested):
We just need to add a small note to the configuration section of the README stating which options can be set on the application environment. Other than that, I would merge this.

paulo-ferraz-oliveira:
I've pushed that update. I'm gonna do a new pull request to brush up the README (it needs some ❤️).

PR #200, opened 2021-07-25
Allow for analysis under rebar3_hank

paulo-ferraz-oliveira:
(also add xref to the analysis)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        otp_vsn: [20, 21, 22, 23, 24]
+        otp_vsn: [21, 22, 23, 24]
         os: [ubuntu-latest]
     steps:
       - uses: actions/checkout@v2
@@ -22,6 +22,8 @@ jobs:
         with:
           otp-version: ${{matrix.otp_vsn}}
           rebar3-version: '3.15'
+      - run: rebar3 hank
+      - run: rebar3 xref
       - run: rebar3 dialyzer
       - run: rebar3 ct
       - run: rebar3 cover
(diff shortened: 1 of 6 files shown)


src/elvis_project.erl
```diff
@@ -25,6 +25,14 @@
         "Please check Elvis's GitHub repository to find out what the "
         "new format is.").
 
+-hank([{unnecessary_function_arguments, [
+            old_configuration_format/3,
```
paulo-ferraz-oliveira, line 29: This is part of a non-declared behaviour. It would be great to have the dynamic callback declarations here 😄

.github/workflows/ci.yml
```diff
@@ -14,14 +14,16 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        otp_vsn: [20, 21, 22, 23, 24]
```
paulo-ferraz-oliveira, line 17: `rebar3_hank` is 21+

src/elvis_project.erl
elbrujohalcon: A comment stating just that above this line would be useful.

elbrujohalcon (changes requested):

src/elvis_project.erl
paulo-ferraz-oliveira: Sure, I'll add it later.

paulo-ferraz-oliveira: Handled with

```erlang
% These are part of a non-declared "behaviour"
% The reason why we don't try to handle them with different arity is
%  that arguments are ignored in different positions (1 and 3) so that'd
%  probably be messier than to ignore the warning
```

Also, I rebased on top of `master`.

elbrujohalcon (approved):

PR #201, opened 2021-07-26
Fix function/arity filtering

paulo-ferraz-oliveira:


src/elvis_core.erl
```diff
@@ -192,7 +192,8 @@ apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
             case lists:member(AnalyzedModule, Ignores) of
                 false ->
                     FilteredConfigMap
-                        = ConfigMap#{ ignore => lists:delete(AnalyzedModule, Ignores) },
+                        = maps:merge(ConfigMap#{ ignore => lists:delete(AnalyzedModule, Ignores) },
```
paulo-ferraz-oliveira, line 195: Module-level filtering wins over config-level filtering.

src/elvis_file.erl
```diff
@@ -54,8 +54,10 @@ parse_tree(Config, Target) ->
 %% @doc Add the root node of the parse tree to the file data, with filtering.
 -spec parse_tree(elvis_config:configs() | elvis_config:config(), file(), elvis_core:rule_config())
     -> {ktn_code:tree_node(), file()}.
-parse_tree(_Config, File = #{parse_tree := ParseTree}, _RuleConfig) ->
-    {ParseTree, File};
+parse_tree(_Config, File = #{parse_tree := ParseTree0}, RuleConfig) ->
+    Ignore = maps:get(ignore, RuleConfig, []),
```
paulo-ferraz-oliveira, line 58: This is basically what was missing... the change to `elvis_core` makes sure we merge all the `ignore`s, as expected, and the change to this makes sure we cover more ground.

elbrujohalcon (approved):

PR #202, opened 2021-07-26
Remove error-prone _meta_SUITE while creating alias _test_

paulo-ferraz-oliveira:


rebar.config
```diff
@@ -23,8 +23,7 @@
 {profiles, [
   {test, [
     {extra_src_dirs, ["test/examples"]},
-    {deps, [ {mixer,       "1.1.1", {pkg, inaka_mixer}}
-           , {meck,        "0.9.2"}
+    {deps, [ {meck,        "0.9.2"}
            , {xref_runner, "1.2.0"}
```
elbrujohalcon, line 27: Do we still need `xref_runner`?

elbrujohalcon (changes requested):

rebar.config
paulo-ferraz-oliveira: No. It's since been removed.

PR #203, opened 2021-07-26
Don't crash upon trying to handle unicode that Erlang accepts

paulo-ferraz-oliveira:
`inaka/mixer` started complaining about some stuff when I added `rebar3_lint` to it, so I dug deeper and tried to come up with a test that shows (hopefully all bases covered) where Unicode handling might've failed in the past (now fixed).

We're not interested in pretty-printing, in this case, we're just interested in not crashing, since these are hopefully exceptions to the rule in anyone's code.


paulo-ferraz-oliveira:
I'm rebasing on top of `unfeature/meta-suite` to not have to deal with that stuff now. Can rebase later, if you wish.

elbrujohalcon (approved):

PR #209, opened 2021-09-16
Fix #208: New rule for numeric format

### LINKED ISSUE #208: [New Rule] Disallow numbers with underscores in them
## Rule Idea
I would like to have a rule that emits a warning if anybody uses underscores in a number, like `1_000_000`.

## Reasoning
I'm aware that this is a radical point of view, but the goal is to be able to consistently grep the code when needed. If I want to find where a certain `123384` value is specified in the code, I don't want to have to grep for `1_?2_?3_?3_?8_?4` instead of `123384`.

## Options
Convenient options in this rule would be:
* Allow the user to define a value above which underscores are _required_, so that one can specify that numbers below `10000` don't use `_`, but numbers above it should always use it.
* Allow the user to define a regex (like `\d{1,3}(_\d\d\d)+` or something) to validate the number format.

elbrujohalcon:
This PR adds a new rule that by default does basically nothing, but it allows devs to specify things like…
> we don't want numbers with underscores in them.

The rule allows users to specify a global regex for numbers and/or one for integers and one for floats.

#### TODO
- [ ] Adjust the wiki once the PR is merged.


elbrujohalcon:
Oh, crap… OTP22… I guess we'll need to consider removing support for it if we're going to merge this PR. Let's think about it for a bit.

facundoolano (approved):

elbrujohalcon:
Following @inaka's tradition… once we implemented something that's not supported in OTP22, we just remove support for it.
To be precise: This library still works on OTP22, it's just that it can't be _tested_ in those versions.

PR #212, opened 2021-10-03
Fix 210: New rule for behavior vs behaviour.

jackyhui96:


src/elvis_style.erl
```diff
@@ -235,6 +240,10 @@ default(numeric_format) ->
      , float_regex => same
      }.
 
+default(incorrect_behaviour_spelling) ->
+    #{ spelling_type => american
```
elbrujohalcon, line 244: Instead of… or on top of `american / british`… I would like to accept the actual words.
In other words, I would like to use this as the configuration for the rule…
```erlang
{elvis_style, behaviour_spelling, #{spelling_type => behaviour}}
```

src/elvis_style.erl
```diff
@@ -765,6 +774,29 @@ numeric_format(Config, Target, RuleConfig) ->
                          IntNodes,
                          check_numeric_format(FloatRegex, FloatNodes, [])).
 
+-spec incorrect_behaviour_spelling(elvis_config:config(),
+                                   elvis_file:file(),
+                                   empty_rule_config()) ->
+    [elvis_result:item()].
+incorrect_behaviour_spelling(Config, Target, RuleConfig) ->
```
elbrujohalcon, line 781: Please rename the rule as `behaviour_spelling`. We try our best to use _positive_ names for the rules :)

elbrujohalcon (changes requested):
Thank you very much, @jackyhui96 !!
Two minimal adjustments and it will be ready to be merged :)

src/elvis_style.erl
```diff
@@ -765,6 +774,29 @@ numeric_format(Config, Target, RuleConfig) ->
...
+            case {NodeType, SpellingType} of
+                {behavior, british}   -> true;
+                {behaviour, american} -> true;
+                _                     -> false
+            end
+        end,
+    case elvis_code:find(Predicate, Root) of
+        [] -> [];
+        InconsistentBehaviorNodes ->
+            ResultFun = result_node_line_fun(?INCORRECT_BEHAVIOuR_SPELLING(SpellingType)),
```
elbrujohalcon, line 796:
```suggestion
            ResultFun = result_node_line_fun(?INCORRECT_BEHAVIOUR_SPELLING(SpellingType)),
```

jackyhui96:
Will work on the changes now and push up, could you label it with hacktoberfest so it counts towards it please? :) Was meant to add this to your original issue which has the label but not sure on the correct process to do that

src/elvis_style.erl
```diff
@@ -126,6 +127,10 @@
         "Number ~p on line ~p does not respect the format "
         "defined by the regular expression '~p'.").
 
+-define(BEHAVIOUR_SPELLING(SpellingType),
+        "The behavior/behaviour in line ~p is misspelt, please use the "
+        ++ atom_to_list(SpellingType) ++ " spelling.").
```
elbrujohalcon, line 132:
```suggestion
-define(BEHAVIOUR_SPELLING(SpellingType),
        "The behavior/behaviour in line ~p is misspelt, please use the "
        "~p spelling.").
```
And then just pass the atom in when building the message.

src/elvis_style.erl
```diff
@@ -233,6 +238,10 @@ default(numeric_format) ->
     #{ regex => ".*"
      , int_regex => same
      , float_regex => same
+     };
+
+default(behaviour_spelling) ->
+    #{ spelling_type => behavior
```
elbrujohalcon, line 244:
```suggestion
    #{ spelling_type => behaviour
```
I checked the source code for OTP and [`behaviour`](https://github.com/erlang/otp/search?q=behaviour) is much more popular than [`behavior`](https://github.com/erlang/otp/search?q=behavior). I think that the one with the `u` should be the default.

elbrujohalcon: Also in general…
* [~167K uses of `behaviour`](https://github.com/search?q=%22-behaviour%22&type=Code&l=Erlang).
* [~8K uses of `behavior`](https://github.com/search?q=%22-behavior%22&type=Code&l=Erlang).

elbrujohalcon: Even your rule is called `behavioUr_spelling` :P

src/elvis_style.erl
```diff
@@ -765,6 +774,29 @@ numeric_format(Config, Target, RuleConfig) ->
...
+    SpellingType = option(spelling_type, RuleConfig, behaviour_spelling),
+    Root = get_root(Config, Target, RuleConfig),
+    Predicate =
+        fun(Node) ->
+            NodeType = ktn_code:type(Node),
+            case {NodeType, SpellingType} of
+                {behavior, behaviour}   -> true;
+                {behaviour, behavior} -> true;
+                _                     -> false
+            end
```
elbrujohalcon, line 791:
```suggestion
            NodeType /= SpellingType
```

elbrujohalcon (changes requested):
I updated the repo tags (It had `hacktober` instead of `hacktoberfest` 🤦🏻 ).
I left a few suggestions to improve the code, too.
Besides them, now that we no longer support `american`/`british` as spelling types, I would be tempted to rename `spelling_type` as `proper_spelling`, or just `spelling`. What do you think?

src/elvis_style.erl
jackyhui96: I still need to check the Node is a type of behaviour, so I've used the below. Any objections?
`lists:member(NodeType, [behaviour, behavior]) andalso NodeType /= Spelling`

elbrujohalcon: It's brilliant! 🧐

elbrujohalcon (changes requested):
This is great, @jackyhui96! Thank you very much!
The only missing things are to add the new rule to the list of default rules for `erl_files` and `beam_files` rulesets and to update [the wiki](https://github.com/inaka/elvis_core/wiki/Rules).

jackyhui96:
> This is great, @jackyhui96! Thank you very much! The only missing things are to add the new rule to the list of default rules for `erl_files` and `beam_files` [rulesets](https://github.com/inaka/elvis_core/blob/master/src/elvis_rulesets.erl) and to update [the wiki](https://github.com/inaka/elvis_core/wiki/Rules).

No worries @elbrujohalcon, happy to help and contribute! I've updated the [the wiki](https://github.com/inaka/elvis_core/wiki/Rules) and pushed the default rule changes now

elbrujohalcon:
Woah! Now we found other tests are broken… 🤦🏻

elbrujohalcon (approved):

PR #214, opened 2021-10-13
Fix #213: Handle no-text numbers in numeric_format

### LINKED ISSUE #213: badarg when applying numeric_format
When applying the new `numeric_format` rule to a file like the one below, it fails with `badarg`:

```erlang
-module(bin).

-type bad() :: <<_:8>>.

-export_type([bad/0]).
```

That's because of the `8` that can be found there and, according to the parser, it has no associated `text` property.

elbrujohalcon:


paulo-ferraz-oliveira:
That's odd. I saw a recent version being released, though. It already contains the numeric stuff fix, right?

PR #215, opened 2021-10-14
Exit with `1` if `escriptize` (from CI) fails

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
The first commit should put the error into evidence. Further commits will be attempts to fix the issue: https://github.com/inaka/elvis_core/runs/3896309155?check_suite_focus=true#step:7:9

PR #219, opened 2022-01-06
Add new rule no_macros (defaults to disabled)

### LINKED ISSUE #160: Complain about preprocessor macro use.
A rule to complain the use of any non-built-in macro, allowing for exceptions via a whitelist. 

e.g.

``` erlang
{elvis_style, no_macros, ["EXCEPTION_1", "EXCEPTION_2"]}
```

paulo-ferraz-oliveira:
Closes #160.

I'll leave some comments in the code regarding decisions or the lack thereof.

We're not testing that the default value of `disabled` (or "omitted") yields no bad results, but experimenting in a different project shows the rule is not applied.

- [x] should I consider not adding `{elvis_style, no_macros}` to `test.config` and revert the test suite changes I did to provoke the issue? Or create a new test?
- [x] potentially act upon https://github.com/inaka/elvis_core/pull/219#discussion_r779789068
- [x] potentially act upon https://github.com/inaka/elvis_core/pull/219#discussion_r779790669
- [ ] make sure the rule's names and options' names are Ok.

### DIFF
--- a/config/test.config
+++ b/config/test.config
@@ -8,7 +8,8 @@
         rules => [{elvis_text_style, line_length, #{limit => 80,
                                                     skip_comments => false}},
                   {elvis_style, nesting_level, #{level => 3}},
-                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}}
+                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}},
+                  {elvis_style, no_macros}
                  ],
         ruleset => erl_files
        },
@@ -17,7 +18,8 @@
         rules => [{elvis_text_style, line_length, #{limit => 80,
                                                     skip_comments => false}},
                   {elvis_style, nesting_level, #{level => 3}},
-                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}}
+                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}},
+                  {elvis_style, no_macros}
                  ],
         ruleset => beam_files
        },
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -246,6 +246,7 @@ merge_rules(UserRules, DefaultRules) ->
         lists:filter(
             fun({_FileName, _RuleName, OverrideOptions}) ->
                 disable /= OverrideOptions;
+               ({_FileName, _RuleName}) -> true; % not disabled
                (_) -> false end,
             UserRules
         ),
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -185,8 +185,6 @@ elvis_attr_rules(ElvisAttrs) ->
            Result :: {Results, ElvisCfg, File}.
 apply_rule({Module, Function}, {Result, Config, File}) ->
     apply_rule({Module, Function, #{}}, {Result, Config, File});
-apply_rule({_Module, _Function, disable}, {Result, Config, File}) ->
-    {Result, Config, File};
 apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
     ConfigMap =
         try
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -185,6 +185,8 @@ elvis_attr_rules(ElvisAttrs) ->
            Result :: {Results, ElvisCfg, File}.
 apply_rule({Module, Function}, {Result, Config, File}) ->
     apply_rule({Module, Function, #{}}, {Result, Config, File});
+apply_rule({_Module, _Function, disable}, {Result, Config, File}) ->
+    {Result, Config, File};
 apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
     ConfigMap =
         try
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -147,6 +147,9 @@ default(macro_names) ->
 default(macro_module_names) ->
     #{};
 
+default(no_macros) ->
+    #{};
+
 default(operator_spaces) ->
     #{ rules => [ {right, ","}
                 , {right, "++"}
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -147,9 +147,6 @@ default(macro_names) ->
 default(macro_module_names) ->
     #{};
 
-default(no_macros) ->
-    disable;
-
 default(operator_spaces) ->
     #{ rules => [ {right, ","}
                 , {right, "++"}
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -6,6 +6,7 @@
          variable_naming_convention/3,
          macro_names/3,
          macro_module_names/3,
+         no_macros/3,
          operator_spaces/3,
          nesting_level/3,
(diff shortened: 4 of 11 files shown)


config/test.config
```diff
@@ -8,7 +8,8 @@
         rules => [{elvis_text_style, line_length, #{limit => 80,
                                                     skip_comments => false}},
                   {elvis_style, nesting_level, #{level => 3}},
-                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}}
+                  {elvis_style, invalid_dynamic_call, #{ignore => [elvis]}},
+                  {elvis_style, no_macros}
```
paulo-ferraz-oliveira, line 12: This effectively activates the check for all our tests (which causes required changes to the test suites, as seen below).

src/elvis_config.erl
```diff
@@ -246,6 +246,7 @@ merge_rules(UserRules, DefaultRules) ->
         lists:filter(
             fun({_FileName, _RuleName, OverrideOptions}) ->
                 disable /= OverrideOptions;
+               ({_FileName, _RuleName}) -> true; % not disabled
```
paulo-ferraz-oliveira, line 249: We were not considering the fact that a rule with no options doesn't necessarily mean `disabled`: hopefully this will surface positives in places where such overrides are being used, though not common (since we had no "disabled by default" rules up until now).

src/elvis_core.erl
```diff
@@ -185,6 +185,8 @@ elvis_attr_rules(ElvisAttrs) ->
            Result :: {Results, ElvisCfg, File}.
 apply_rule({Module, Function}, {Result, Config, File}) ->
     apply_rule({Module, Function, #{}}, {Result, Config, File});
+apply_rule({_Module, _Function, disable}, {Result, Config, File}) ->
```
paulo-ferraz-oliveira, line 188: Seems odd to put this here, but required, since when the rule is not overridden it is considered disable (up until now all our rules were either enabled or had an options map, by default).

src/elvis_style.erl
```diff
@@ -44,6 +45,23 @@
 -define(MACRO_AS_FUNCTION_NAME_MSG,
             "Don't use macros (like ~s on line ~p) as function names.").
 
+-define(NO_MACROS_MSG,
+            "Unexpected macro (~p) used on line ~p.").
+
+-define(PREDEF_MACROS, % from unexported eep:predef_macros/1
```
paulo-ferraz-oliveira, line 51: Should we act differently here? I don't think the list will change that much and the first one to find a false positive gets to come and PR 😄 (we could also fetch this list dynamically, but 1. extra overhead, 2. we'd have to use a parse transform to avoid that overhead). I'm open to suggestions.

src/elvis_style.erl
```diff
@@ -329,6 +350,40 @@ macro_module_names(Config, Target, RuleConfig) ->
...
+
+    lists:foldl(
+        fun (MacroNode, Acc) ->
+            Macro = list_to_atom(ktn_code:attr(name, MacroNode)),
+            case lists:member(Macro, AllowedMacros) of
+                true ->
+                    Acc;
+                false ->
+                    {Line, _Col} = ktn_code:attr(location, MacroNode),
+                    [elvis_result:new(item, ?NO_MACROS_MSG, [Macro, Line]) | Acc]
```
paulo-ferraz-oliveira, line 377: Results in reverse order to avoid `++` but I'm not sure this is interesting. (I do fix my files from top to bottom, though, since that way I can copy paste the issues to a text file and fixing up won't change line numbers - thus avoiding unnecessary re-runs).

test/examples/fail_used_ignored_variable.erl
```diff
@@ -6,6 +6,8 @@
         , macro_with_underscore/0
         ]).
 
+-elvis([{elvis_style, no_macros, #{allow => ['__']}}]).
```
paulo-ferraz-oliveira, line 9: Best macro ever! Very descriptive 😄

test/style_SUITE.erl
```diff
@@ -312,6 +314,22 @@ verify_macro_module_names(Config) ->
...
 
+-spec verify_no_macros(config()) -> any().
+verify_no_macros(Config) ->
+    Ext = proplists:get_value(test_file_ext, Config, "erl"),
+
+    PathFail = "fail_no_macros." ++ Ext,
+    FailRes = elvis_core_apply_rule(Config, elvis_style, no_macros, #{}, PathFail),
+    case Ext of
+        "beam" ->
+            [] = FailRes; % no macros on BEAM files
```
paulo-ferraz-oliveira, line 325: This also prompted me to review the Rules (they've been updated targeting 1.4.0). Check the diff: https://github.com/inaka/elvis_core/wiki/Rules/_compare/fb9c3a3b655faf66fd37f1d1ef125ed1ded9c592...4bb05900cbb419eee2b01a772ec966258ddcb0cb

paulo-ferraz-oliveira:
Tests passing. All yours!

src/elvis_style.erl
elbrujohalcon: It's extra-funny to implement a new rule to avoid macros using yet another macro to keep the list of macros :P
Maybe put this in a function instead?

src/elvis_style.erl
elbrujohalcon: I don't really care either way.

test/examples/fail_used_ignored_variable.erl
elbrujohalcon: 🪄

test/style_SUITE.erl
elbrujohalcon: 👌🏻

src/elvis_style.erl
```diff
@@ -143,6 +161,9 @@ default(macro_names) ->
 default(macro_module_names) ->
     #{};
 
+default(no_macros) ->
+    disable;
```
elbrujohalcon, line 165: I would not have a rule disabled by default. Leave it with options but _outside_ of any rulesets. If users want to use it, they can include it in the `rules` part, which is what they must do in this situation anyway, right?

elbrujohalcon (changes requested):
I like this new rule, but instead of making it disabled by default (which, to me, makes little sense), I would just leave it out of the `rulesets`.

src/elvis_style.erl
paulo-ferraz-oliveira: I will. I just thought it would be unnecessary overhead, since the macro definition is pre-processed.

paulo-ferraz-oliveira: 1cd6197

src/elvis_style.erl
paulo-ferraz-oliveira: 6f88302

paulo-ferraz-oliveira:
Are the rule's names and options' names Ok?

elbrujohalcon:
> Are the rule's names and options' names Ok?

Yes.

elbrujohalcon (approved):

paulo-ferraz-oliveira:
Wonderful. Thanks for merging.

PR #222, opened 2022-02-21
Make state_record_and_type not default

### LINKED ISSUE #171: Remove state_record_and_type from erl_files
This is an outdated rule.
The concept is not bad (_When implementing an OTP behavior, define some type to hold its state_), but the actual implementation of the rule requires the type to be called `state/0` and that's too strict. For instance, we have implementations of `gen_statem` where it makes much more sense to call that type `data/0` or `state_data/0`.

We should either allow the record name to be matched against a more flexible regular expression or just remove the rule from `erl_files` altogether.

paulo-ferraz-oliveira:
(not only is this rule's value somewhat limited, but there's, at the moment, no easy way to track the behaviours for which we want state to be named as such - and it's a decision probably best left to the developer)

This closes #171.

### DIFF
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -30,7 +30,6 @@ rules(erl_files) ->
         , {elvis_style, no_behavior_info}
         , {elvis_style, module_naming_convention}
         , {elvis_style, function_naming_convention}
-        , {elvis_style, state_record_and_type}
         , {elvis_style, no_spec_with_records}
         , {elvis_style, dont_repeat_yourself}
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Hm... `windows-latest` (from CI) will probably fail due to https://github.com/erlef/setup-beam/issues/90. I'm sorting out the merging of that.

paulo-ferraz-oliveira:
Hm... it seems either `windows-latest` was never Windows'22 or it got reverted recently. In any case, "All checks have passed". ✅

PR #223, opened 2022-03-03
Add proper line numbers to result items.

ztion:
This change will add line numbers to some result items which previously always set line number to 0. This will make it more useful for ErlangLS, as the diagnostics found can be shown on the real line in the editor of the user.


elbrujohalcon (approved):

elbrujohalcon:
@ztion I guess you'll need a new release of `elvis_core` published to hex.pm, right?

ztion:
@elbrujohalcon Hi! Yes, that would be nice!

elbrujohalcon:
You got it! https://hex.pm/packages/elvis_core/1.3.2

PR #226, opened 2022-04-20
Add new rule: no_throw

### LINKED ISSUE #224: New Rule: Avoid throw
Stemming from [the conversation on the Erlang Forums](https://erlangforums.com/t/should-we-eventually-improve-catch-throw-semantics/1210/6):

> IMO, it is a really bad style, especially in the context of a functional language. It is returning something through side effects. That may have its uses to [quickly break out of deep recursion](https://learnyousomeerlang.com/errors-and-exceptions#try-a-try-in-a-tree), but they are rare. I think I never used throw even once. In any case, I think it should not be made “socially acceptable” to return stuff via throw, much less encouraged.
(said @Maria-12648430)

> I definitely agree. I prefer to use tail-recursive functions instead of non-local returns unless that turns out to be too cumbersome. That is, if the code becomes convoluted written in a tail-recursive way, I will write it in a body-recursive way and use throw to break out of the recursion. I have typically used that in complicated optimization passes in the compiler; that is, the code will do the optimization in recursive functions, but if anything turns up that makes the optimization impossible I will throw a `not_possible` exception to break out of the recursion.
(said @bjorng)

paulo-ferraz-oliveira:
Closes #224.

**Edit**:

- the [Wiki](https://github.com/inaka/elvis_core/wiki/Rules#no-throw) is updated and proposes version 1.4.0 for this rule to be included
- I'm not sure about the rule name
- I'm not sure if the rule should be enabled by default

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ The following is a checklist you can follow when implementing a new Elvis rule:
 
 - [ ] Write both test cases that fail and that pass the rule.
 - [ ] Edit the [Rules](https://github.com/inaka/elvis/wiki/Rules) wiki page, by adding the rule to the list.
-- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/main/config/elvis.config) file.
+- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/HEAD/config/elvis.config) file.
 
 ## Questions?
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ The following is a checklist you can follow when implementing a new Elvis rule:
 
 - [ ] Write both test cases that fail and that pass the rule.
 - [ ] Edit the [Rules](https://github.com/inaka/elvis/wiki/Rules) wiki page, by adding the rule to the list.
-- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/master/config/elvis.config) file.
+- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/main/config/elvis.config) file.
 
 ## Questions?
 
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ The default value for `parallel` is `1`.
 ## Configuration examples
 
 You can find examples for configuration files in this project's
-[`config`](https://github.com/inaka/elvis_core/tree/master/config) directory.
+[`config`](https://github.com/inaka/elvis_core/tree/main/config) directory.
 
 ## Application environment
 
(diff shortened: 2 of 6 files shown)


src/elvis_style.erl
```diff
@@ -1437,6 +1464,17 @@ call_mfa(Call) ->
     A = length(ktn_code:content(Call)),
     {M, F, A}.
 
+is_call(_Node, []) ->
+    false;
+is_call(Node, [MFA | Tail]) ->
+    is_call(Node, MFA) orelse is_call(Node, Tail);
```
paulo-ferraz-oliveira, line 1470: Does this feel odd?

test/examples/fail_no_throw.erl
```diff
@@ -0,0 +1,19 @@
+-module(fail_no_throw).
```
paulo-ferraz-oliveira, line 1: Is this enough?

paulo-ferraz-oliveira:
I moved some stuff from `master` to `main` since the GHA were blocked. I created a task to deal with the rest later.

CONTRIBUTING.md
```diff
@@ -23,7 +23,7 @@ The following is a checklist you can follow when implementing a new Elvis rule:
 
 - [ ] Write both test cases that fail and that pass the rule.
 - [ ] Edit the [Rules](https://github.com/inaka/elvis/wiki/Rules) wiki page, by adding the rule to the list.
-- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/master/config/elvis.config) file.
+- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/main/config/elvis.config) file.
```
elbrujohalcon, line 26:
```suggestion
- [ ] When the issue has the label `default rule`, add the rule to [Default Rules in Elvis Webhook](https://github.com/inaka/elvis/wiki/Default-Rules-in-Elvis-Webhook) and in the [`config/elvis.config`](https://github.com/inaka/elvis/blob/HEAD/config/elvis.config) file.
```
To be ready for the next few trends in brain-naming fashion.

README.md
```diff
@@ -222,7 +222,7 @@ The default value for `parallel` is `1`.
 ## Configuration examples
 
 You can find examples for configuration files in this project's
-[`config`](https://github.com/inaka/elvis_core/tree/master/config) directory.
+[`config`](https://github.com/inaka/elvis_core/tree/main/config) directory.
```
elbrujohalcon, line 225:
```suggestion
[`config`](https://github.com/inaka/elvis_core/tree/HEAD/config) directory.
```

src/elvis_style.erl
```diff
@@ -127,6 +128,9 @@
         "Atom ~p on line ~p does not respect the format "
         "defined by the regular expression '~p'.").
 
+-define(NO_THROW_MSG,
+        "The call to throw/1 on line ~p is unexpected").
```
elbrujohalcon, line 132:
```suggestion
        "Usage of throw/1 on line ~p is not recommended").
```

src/elvis_style.erl
elbrujohalcon: Yeah, I would've used `lists:any(fun is_call/1, Expressions)` or something like that.

test/examples/fail_no_throw.erl
elbrujohalcon: Yes.

elbrujohalcon (changes requested):
A few minor recommendations.

src/elvis_style.erl
paulo-ferraz-oliveira: That's what I was looking for. At 3 AM the brain was mostly shut down 😄

elbrujohalcon (approved):

elbrujohalcon:
Oh, wait… I merged and then I realized that this rule was not added to the `erl_files` / `beam_files` rulesets… and I think it should be included there. Sorry.

paulo-ferraz-oliveira:
No prob. I did ask about that, but you're just too quick 😄 I'll pull request a new one.

PR #230, opened 2022-04-20
Make no_throw default

paulo-ferraz-oliveira:


elbrujohalcon:
Looks like we did have some throws here and there… 🤷🏻

elbrujohalcon (approved):

Issue #229, opened 2022-04-20
Apply a formatter to the code


elbrujohalcon:
Add `rebar3_format` and apply it to the code in this repo.
Also add it to CI pipelines.

PR #231, opened 2022-04-24
Add new rule no_space

paulo-ferraz-oliveira:
**Edit**: the [Wiki](https://github.com/inaka/elvis_core/wiki/Rules#no-space) already expects this from 1.4.0 on. If we decide to make it default I'm sure it can be considered a breaking change (for those relying on `main`).

### DIFF
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -179,7 +179,7 @@ print_node(Node) ->
 -spec print_node(ktn_code:tree_node(), integer()) -> ok.
 print_node(Node = #{type := Type}, CurrentLevel) ->
     Type = ktn_code:type(Node),
-    Indentation = lists:duplicate(CurrentLevel * 4, $ ),
+    Indentation = lists:duplicate(CurrentLevel * 4, $\s),
     Content = ktn_code:content(Node),
 
     ok = elvis_utils:info("~s - [~p] ~p~n", [Indentation, CurrentLevel, Type]),
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -50,9 +50,9 @@
 -define(NO_MACROS_MSG,
             "Unexpected macro (~p) used on line ~p.").
 
--define(MISSING_SPACE_MSG, "Missing space ~s ~p on line ~p").
+-define(MISSING_SPACE_MSG, "Missing space to the ~s of ~p on line ~p").
 
--define(UNEXPECTED_SPACE_MSG, "Unexpected space ~s ~p on line ~p").
+-define(UNEXPECTED_SPACE_MSG, "Unexpected space to the ~s of ~p on line ~p").
 
 -define(NESTING_LEVEL_MSG,
         "The expression on line ~p and column ~p is nested "
@@ -167,6 +167,7 @@ default(operator_spaces) ->
 default(no_space) ->
     #{ rules => [ {right, "("}
                 , {left, ")"}
+                , {left, ","}
                 ]
      };
 
@@ -424,7 +425,7 @@ operator_spaces(Config, Target, RuleConfig) ->
     AllNodes = OpNodes ++ PunctuationTokens,
 
     FlatMap = fun(Rule) ->
-                  check_spaces(Lines, AllNodes, Rule, Encoding, should_have)
+                  check_spaces(Lines, AllNodes, Rule, Encoding, {should_have, []})
               end,
     lists:flatmap(FlatMap, Rules).
 
@@ -448,8 +449,23 @@ no_space(Config, Target, RuleConfig) ->
     TextNodes = lists:filter(fun is_text_node/1, Tokens),
     {Src, #{encoding := Encoding}} = elvis_file:src(Target),
     Lines = elvis_utils:split_all_lines(Src),
+    AllSpaceUntilText
+        = [{Text,
+            re:compile("^[ ]+" ++ re:replace(
+                                      Text,
+                                      "(\\.|\\[|\\]|\\^|\\$|\\+|\\*|\\?|\\{|\\}|\\(|\\)|\\||\\\\)",
+                                      "\\\\\\1",
+                                      [{return, list}, global]
+                                  )
+            )} || {left, Text} <- Rules],
     FlatMap = fun(Rule) ->
-                  check_spaces(Lines, TextNodes, Rule, Encoding, should_not_have)
+                  check_spaces(
+                      Lines,
+                      TextNodes,
+                      Rule,
+                      Encoding,
+                      {should_not_have, AllSpaceUntilText}
+                  )
               end,
     lists:flatmap(FlatMap, Rules).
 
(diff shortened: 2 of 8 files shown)


src/elvis_style.erl
```diff
@@ -1144,26 +1176,34 @@ has_remote_call_parent(Zipper) ->
             has_remote_call_parent(zipper:up(Zipper))
     end.
 
-%% Operator Spaces
--spec check_operator_spaces(Lines :: [binary()],
-                            OperatorNodes :: [ktn_code:tree_node()],
-                            Rule :: {right | left, string()},
-                            Encoding :: latin1 | utf8) ->
+%% Operator (and Text) Spaces
```
paulo-ferraz-oliveira, line 1179: Most name spaces were done to make this function more generic

test/examples/fail_no_space.erl
```diff
@@ -0,0 +1,111 @@
+%% Initial comment
```
paulo-ferraz-oliveira, line 1: There's a couple of extra spaces to make the test fail (which means it passes).

elbrujohalcon (approved):

elbrujohalcon:
I would consider this a default, yes. I'm not an expert on SemVer, but being a breaking change, is `1.4.0` fine?

paulo-ferraz-oliveira:
When your change/update means the consumer has to (potentially) update code in order for stuff to continue working, that's a major bump. In this case, for example, this new rule might break some CI out there. I don't mind making it default, in any case.

paulo-ferraz-oliveira:
Also, since I believe we'll want this to be `rebar3_format` compatible, in the future (given #229), what other elements, like `{` should we add as a default?

paulo-ferraz-oliveira:
And (since we're breaking interface) on a related change we could introduce spaces around `=:=`, `==`, `=/=`, `/=`, `=`, and many other operators. Does this make sense?

paulo-ferraz-oliveira:
(thinking on it, I believe any default might be Ok, the list will grow as more and more people use if and expect other defaults, maybe 🤷)

paulo-ferraz-oliveira:
I've pushed the "make it default."

paulo-ferraz-oliveira:
The "making it default" showed two new interesting results.

1. I hadn't contemplated (in the tests), lines like `      )`,
2. I hadn't contemplated `$ ` (space character) which I dislike, since it's "invisible". I will be replacing that with `$\s`.

elbrujohalcon:
Yeah, `$ ` is a bad thing. I would also like Elvis to tell me not to use it (i.e. to replace it by `$\s`.

elbrujohalcon:
About item 1 in your list… would the idea be to ignore the rule if `)` is the first (non-space) character in the line?

elbrujohalcon:
> Also, since I believe we'll want this to be `rebar3_format` compatible, in the future (given #229), what other elements, like `{` should we add as a default?

We don't _necessarily_ want Elvis to be compatible with `rebar3_format` (although that would be nice), #229 is just to format the code in this repo. That's all.

elbrujohalcon:
> And (since we're breaking interface) on a related change we could introduce spaces around `=:=`, `==`, `=/=`, `/=`, `=`, and many other operators. Does this make sense?

This makes sense to me, indeed. But this can happen in a different PR.

paulo-ferraz-oliveira:
> https://github.com/inaka/elvis_core/issues/229 is just to format the code in this repo. That's all.

I know, but this repo. is under `elvis_core` validations, too 😄 In any case, I won't worry about that for now. Come to think of it, when you have a formatter and a linter combined, the linter worrying about text style is less relevant. (I think we've talked about this in the past, and either you wrote something somewhere or it was just an idea - vague recollection)

paulo-ferraz-oliveira:
> I would also like Elvis to tell me not to use it (i.e. to replace it by $\s.

I'll add an exception to this in the test case (and hopefully it suffices to add it to the default rules).

elbrujohalcon:
> > #229 is just to format the code in this repo. That's all.
> 
> I know, but this repo. is under `elvis_core` validations, too 😄 In any case, I won't worry about that for now. Come to think of it, when you have a formatter and a linter combined, the linter worrying about text style is less relevant. (I think we've talked about this in the past, and either you wrote something somewhere or it was just an idea - vague recollection)

I did wrote an article in my blog… but I'm lazy and I don't want to find the link now :P

paulo-ferraz-oliveira:
I've updated the code to deal with the "whitespace until char, when left" type of situation.

`$ ` will be handled elsewhere, since this is not a text character.

I'll push the changes later, since more tests are now failing for this being a default 😄

src/elvis_code.erl
```diff
@@ -179,7 +179,7 @@ print_node(Node) ->
 -spec print_node(ktn_code:tree_node(), integer()) -> ok.
 print_node(Node = #{type := Type}, CurrentLevel) ->
     Type = ktn_code:type(Node),
-    Indentation = lists:duplicate(CurrentLevel * 4, $ ),
+    Indentation = lists:duplicate(CurrentLevel * 4, $\s),
```
paulo-ferraz-oliveira, line 182: As per new rule `no_space` being executed on the lib.'s code.

src/elvis_style.erl
```diff
@@ -167,6 +167,7 @@ default(operator_spaces) ->
 default(no_space) ->
     #{ rules => [ {right, "("}
                 , {left, ")"}
+                , {left, ","}
```
paulo-ferraz-oliveira, line 170: As per `operator_spaces`' "mirror" option.

src/elvis_style.erl
```diff
@@ -448,8 +449,23 @@ no_space(Config, Target, RuleConfig) ->
     TextNodes = lists:filter(fun is_text_node/1, Tokens),
     {Src, #{encoding := Encoding}} = elvis_file:src(Target),
     Lines = elvis_utils:split_all_lines(Src),
+    AllSpaceUntilText
+        = [{Text,
+            re:compile("^[ ]+" ++ re:replace(
+                                      Text,
+                                      "(\\.|\\[|\\]|\\^|\\$|\\+|\\*|\\?|\\{|\\}|\\(|\\)|\\||\\\\)",
```
paulo-ferraz-oliveira, line 456: These are the special characters for regex that we might want to transform from consumer input.

test/examples/fail_no_space.erl
```diff
@@ -67,6 +69,7 @@ function8(
 
 function9(
 ) ->
+    _ = $ ,
```
paulo-ferraz-oliveira, line 72: Will eventually provoke a badmatch, in a future rule implementation, for now its just a "placeholder".

test/examples/pass_operator_spaces_elvis_attr.erl
```diff
@@ -6,7 +6,7 @@
 
 -elvis([{elvis_style , operator_spaces , #{rules => [{right , "++"} ,
                                                      {left , ","}]}}]).
-
+-elvis([{elvis_style , no_space , #{rules => [{right , "("} , {left , ")"}]}}]).
```
paulo-ferraz-oliveira, line 9: All the spaces before `,` are because of the rule being executed: `operator_spaces`.

paulo-ferraz-oliveira:
I know you've approved it, but recent changes brought a bit more code, so an extra pair of eyes (no hurry) would help.

elbrujohalcon (approved):
At some point, we'll have to review the test files (all of them, not just yours), so that their text actually reflects what we're trying to test with them. And you might want to adjust your comments considering that now you know that `$ ` is not a space _after_ `$`, but the space character.
None of those things matter that much to me, so I'll approve this PR.
Merge it at will.

paulo-ferraz-oliveira:
Updated as per last comment. Merging...

PR #233, opened 2022-04-26
Add new rule no_block_expressions

### LINKED ISSUE #147: New rule: avoid blocks expressions
Definition: [Block Expressions](https://erlang.org/doc/reference_manual/expressions.html#block-expressions).

**Reasoning**: Use functions… with names… don’t be shy! (by @elbrujohalcon)

Also, we should support `ignore` as we do for other rules (let's just hope `ignore` is already generalized once we get here 😉).

paulo-ferraz-oliveira:
Closes #147.


paulo-ferraz-oliveira:
Hurrah self-use: https://github.com/inaka/elvis_core/pull/233/checks#step:7:12

elbrujohalcon (approved):

PR #236, opened 2022-05-04
Improve debug

### LINKED ISSUE #182: Ease debug
Recently, @robertoaloi reported an issue where he was getting a `function_clause`:

```
# apps/els_dap/test/els_dap_general_provider_SUITE.erl [FAIL]
Error: 'function_clause' while applying rule 'operator_spaces'.
```

This exception is being eaten up internally. We should figure out how to either throw it or use `rebar3`'s logging facilities (if any) to ease this process.

paulo-ferraz-oliveira:
Exceptions aren't supposed to occur, so instead of just hiding the stacktrace we display it in all its glory...

Closes #182.

### DIFF
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -212,9 +212,9 @@ apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
                     elvis_result:new(rule, Function, [])
             end
         catch
-            _:Reason ->
-                Msg = "'~p' while applying rule '~p'.",
-                elvis_result:new(error, Msg, [Reason, Function])
+            _:Reason:Stacktrace ->
+                Msg = "'~p' while applying rule '~p': ~p.",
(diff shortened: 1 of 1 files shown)


src/elvis_core.erl
```diff
@@ -212,9 +212,9 @@ apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
                     elvis_result:new(rule, Function, [])
             end
         catch
-            _:Reason ->
-                Msg = "'~p' while applying rule '~p'.",
-                elvis_result:new(error, Msg, [Reason, Function])
+            _:Reason:Stacktrace ->
```
paulo-ferraz-oliveira, line 215: It's simple, I know, but I'm wondering if we needed to hide the stacktrace for any reason...

elbrujohalcon: Probably to be compatible with OTP22 without using a macro

elbrujohalcon (approved):

PR #238, opened 2022-05-26
Fixes #177 by rewriting macro_module_names rule

### LINKED ISSUE #177: macro_module_names not failiing on OTP 24
Still haven't looked into this in detail, but I get 

```
# src/mustache.erl [FAIL]
  - macro_module_names
    - Don't use macros (like MUSTACHE_CTX on line 95) as module names.
```

in OTP 19, but not OTP 24.

elbrujohalcon:
Fixes #177 by rewriting macro_module_names rule


paulo-ferraz-oliveira (approved):
Seems reasonable to me...

PR #242, opened 2022-06-02
Fix #241: Get Elvis ready for OTP25

### LINKED ISSUE #241: Make sure that Elvis runs on OTP25
(What the title says)

elbrujohalcon:
Fix #241: Get Elvis ready for OTP25

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,5 +50,8 @@ jobs:
         if: ${{ matrix.os == 'windows-latest' }}
         run: |
           set ERL_FLAGS="-enable-feature all"
-          rebar3 do compile, format --verify, test, escriptize
+          rebar3 compile
+          rebar3 format --verify
+          rebar3 test
+          rebar3 escriptize
           _build/default/bin/elvis_core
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        otp_vsn: [23.3, 24.3]
+        otp_vsn: [23.3, 24.3, 25.0]
```
paulo-ferraz-oliveira, line 16: If you prefer you can also use 25, and `setup-beam` will always choose the latest available.

elbrujohalcon: I prefer to know which minor version we're using.

test/examples/fail_no_nested_try_catch.erl
```diff
@@ -1,6 +1,6 @@
 -module(fail_no_nested_try_catch).
 
--ignore_xref({maybe, throw, 1}).
```
paulo-ferraz-oliveira, line 3: Hahaha. In any case, do we need tests for `maybe`, specifically?

paulo-ferraz-oliveira (approved):

PR #244, opened 2022-06-09
Add new rule: no_author

### LINKED ISSUE #149: New Rule: Don't use -author
For some/most of my projects, nobody cares who _authored_ a module.
Also generally, the `author` attribute usually only represents the first developer that created the module, which may as well had left the project a long long time ago.
It sometimes contains an email in atom format, sometimes in a string, sometimes it's just a name…

In general, this is a very misused attribute.

So, I'm proposing a new rule to warn against the usage of such attribute.
To be clear: it should *not* be part of the `erl_files` ruleset.

elbrujohalcon:
Fixes #149.

The wiki was already updated: https://github.com/inaka/elvis_core/wiki/Rules#no-author

### DIFF
--- a/src/elvis_rulesets.erl
+++ b/src/elvis_rulesets.erl
@@ -36,6 +36,7 @@ rules(erl_files) ->
                {elvis_style, no_nested_try_catch},
                {elvis_style, atom_naming_convention},
                {elvis_style, no_throw},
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira (approved):
This look good.

PR #251, opened 2022-06-16
Fix #227: Adapt master to main everywhere

### LINKED ISSUE #227: Adapt master > main
We should probably (since `main` is now in many places) have new / copied rules for `main`, where `master` is present.

elbrujohalcon:
@paulo-ferraz-oliveira is this the right approach?
Should I instead had…

1. Created new rules for `main` but keep the ones for `master`?
2. Added a `branches` option to `no_deps_master_*` rules and default to `[master, main]`?
3. Like item 2 but also renaming the rules to `no_deps_default_branch_*`?
4. Just be super-strict and forbid dependencies using branches, period? i.e. Always use a tag or commit ref.
5. Remove the rules altogether?


paulo-ferraz-oliveira:
I kinda like that this rule is in place (not the naming, but the verification).

`elvis_core` is already opinionated, though, but not allowing `master` (meaning a dep. would need to be changed) seems a little farfetched to me.

I'd go with option 2., for clarity, or 4, for brevity. Options can always be disabled (and these ones pertain to a single file, so...).

elbrujohalcon:
@paulo-ferraz-oliveira I was about to take the 2nd approach, but at some point handling `erlang.mk` deps consistently became too complex (because there is no real distinction between branches, tags, and commits there) and so I took the brutal approach.
These are the changes this PR introduces as of today:
* We no longer support the analysis of `Makefile`. That is, I removed the `makefiles` ruleset together with all the `erlang_mk` rules.
* `no_deps_master_rebar` is replaced by `no_branch_deps` which will issue a warning if you use a branch for a dependency in your `rebar.config`. It will recommend you to use a `tag` or a `ref`.
* `protocol_for_deps_rebar` is now called `protocol_for_deps`.
* `git_for_deps_rebar` which is actually an alias for `protocol_for_deps_rebar` is now called `git_for_deps`.

If you approve these changes, I'll update the wiki accordingly.

paulo-ferraz-oliveira:
> We no longer support the analysis of Makefile. That is, I removed the makefiles ruleset together with all the erlang_mk rules.

I'm Ok with this. But since this implies a removal whoever uses a new `elvis_core` version will have an exception to deal with. I consider this a breaking change (no problem, just bump the major).

> no_deps_master_rebar is replaced by no_branch_deps which will issue a warning if you use a branch for a dependency in your rebar.config. It will recommend you to use a tag or a ref.

I'm Ok with this; and it's also a breaking change.

> protocol_for_deps_rebar is now called protocol_for_deps.

I'm Ok with this; and it's also a breaking change.

> git_for_deps_rebar which is actually an alias for protocol_for_deps_rebar is now called git_for_deps.

I'm Ok with this; and it's also a breaking change.

paulo-ferraz-oliveira (approved):
This is Ok.

Do consider a major version bump when releasing the lib. next, though (check Conversation).

elbrujohalcon:
Major version bump, it is!

elbrujohalcon:
Wiki updated: https://github.com/inaka/elvis_core/wiki/Rules#no-deps-master-rebar

Issue #167, opened 2014-12-02
Add code suggestions to comments


elbrujohalcon:
When possible, add code suggestions to the warnings.
As an example,
Instead of…

> - macro_names
>   - Invalid macro name the_macro on line 11. Use UPPER_CASE.

…say…

> - macro_names
>   - Invalid macro name the_macro on line 11. Use UPPER_CASE.
>   - QuickFix: replace it by `THE_MACRO`

paulo-ferraz-oliveira:
I started to look at this, but not all, I guess, require a "quick fix", especially since it might not be so quick. How do you propose a fix for `line_length`, for example? "Use a formatter?" 😄 I'll propose a few (when I have some time to code) and maybe we can go from there.

elbrujohalcon:
Yeah… I agree that some rules don't lend themselves easily to _Quick Fixes_. So, maybe… we can say "There is no quick fix here" for those.

PR #253, opened 2022-06-23
Move docs from wiki to code

### LINKED ISSUE #252: Move docs from wiki to code
## Is your feature request related to a problem? Please describe.
Wikis are apart from code so each pull request for a new or updated rule (which are most of the PRs, anyway) requires an extra update somewhere else.
It would be better if what's now on the wiki would be in the README or in a `doc` folder (maybe with one md document per rule?).

I would like to move the current documentation from the wiki to somewhere in the code.

None.

## Additional Context
As a bonus: I would like that to be uploaded to hexdocs when the project is published.

paulo-ferraz-oliveira:
Closes #252 and touches https://github.com/inaka/elvis_core/issues/167#issuecomment-702371801.

After this we can probably hide the Wiki from the project.


elbrujohalcon (approved):
I like it as it is, but I was wondering if it would made our life easier in the future to have per-module subfolders, like `elvis_style/dont_repeat_yourself.md` instead of just `dont_repeat_yourself.md` 🤔

paulo-ferraz-oliveira:
I can do that, no prob. I believe the module name should be in the context of the execution, already.

elbrujohalcon:
Yeah… and… in _theory_ we can totally have `elvis_style:no_tabs` and `elvis_project:no_tabs`. I can't think of any good reason to do that… but, it's in fact possible.

paulo-ferraz-oliveira:
Yeah, the module is out of context, but it doesn't seem hard to get. I'll try to do this at lunch, today.

PR #255, opened 2022-07-21
Fixes #157: New Rule: No successive maps

### LINKED ISSUE #157: Don't use map()map() syntax
The idea behind this rule comes from [this email](http://erlang.org/pipermail/erlang-questions/2017-April/092112.html) by @kvakvs.
Basically, the goal is to emit a warning if someone _forgets a comma_ and writes something like:

```erlang
[#{a => 1}
 #{b => 2}]
```

It will also warn on things like ``#{a =>1}#{b => 2}``, but I think it's worth it since that expression is easily rewritable as ``#{a => 1, b => 2}``.

elbrujohalcon:
Fixes #157: New Rule: No successive maps

I didn't get to implement the parentheses exception that @kvakvs mentions in the ticket.
I guess it can be implemented in a non-trivial way, but it was too complex for me.
So, for now… no successive maps are allowed, regardless of the parenthesizing.


kvakvs (approved):
👌

PR #259, opened 2022-10-21
New Rule: Consistent Generic Type

### LINKED ISSUE #258: New rule: term_or_any
### Name

term_or_any

### Brief Description

As per https://github.com/erlang/otp/issues/6289 we could imagine `elvis_core` having an opinion on which one is more reasonable: `any()` or `term()`. Personally, I prefer `term()` since it relates directly to the documentation. `any()` _seems_ too generic (I know, highly debatable). In any case, we could also allow the rule to (by default) allow term and any, and only propose one or the other based on consumption preference.

### Reasoning

Check link in previous section.

### Refactoring Proposal

None, at the moment.

jackyhui96:
Closes #258


paulo-ferraz-oliveira:
@jackyhui96, if you change your description to e.g. `Closes #258`, it automatically links to that issue.

RULES.md
```diff
@@ -46,6 +46,7 @@ identified with `(since ...)` for convenience purposes.
 - [Numeric Format](doc_rules/elvis_style/numeric_format.md)
 - [Operator Spaces](doc_rules/elvis_style/operator_spaces.md)
 - [State Record and Type](doc_rules/elvis_style/state_record_and_type.md)
+- [Types Term vs Any](doc_rules/elvis_style/types_term_or_any.md)
```
paulo-ferraz-oliveira, line 49:
```suggestion
- [Types `term()` or `any()`](doc_rules/elvis_style/types_term_or_any.md)
```

doc_rules/elvis_style/types_term_or_any.md
```diff
@@ -0,0 +1,16 @@
+# Types Term vs Any
```
paulo-ferraz-oliveira, line 1:
```suggestion
# Types `term()` or `any()`
```

doc_rules/elvis_style/types_term_or_any.md
```diff
@@ -0,0 +1,16 @@
...
+
+Use `term()` or `any()` consistenttly for types in specs.
+
+> Works on `.beam` file? Yes.
+
+## Options
+
+- `preference :: term | any`.
+  - default: `no_preference`.
```
paulo-ferraz-oliveira, line 10: I don't think we need a default, in this case, since `elvis_core` will not be opinionated in this sense. The default is to not include the rule... But if we do end up with a default I'd prefer `none` or `undefined`.

src/elvis_style.erl
```diff
@@ -99,6 +99,9 @@
 -define(ALWAYS_SHORTCIRCUIT_MSG,
         "Non-shortcircuiting operator (~p) found in line ~p. "
         "It's recommended to use ~p, instead.").
+-define(TYPES_TERM_OR_ANY,
+        "The type in line ~p is not preferred, please use the "
```
paulo-ferraz-oliveira, line 103: I'd go with something like "Found ~p on line ~p, but your preferred type is ~p".

doc_rules/elvis_style/types_term_or_any.md
```diff
@@ -0,0 +1,16 @@
+# Types Term vs Any
+
+Use `term()` or `any()` consistenttly for types in specs.
+
+> Works on `.beam` file? Yes.
+
+## Options
+
+- `preference :: term | any`.
```
paulo-ferraz-oliveira, line 9: I'd go with a shorter `prefer` name for the option.

src/elvis_style.erl
```diff
@@ -897,6 +902,37 @@ behaviour_spelling(Config, Target, RuleConfig) ->
...
+-spec types_term_or_any(elvis_config:config(), elvis_file:file(), empty_rule_config()) ->
+                           [elvis_result:item()].
+types_term_or_any(Config, Target, RuleConfig) ->
+    TypePreference = option(preference, RuleConfig, types_term_or_any),
+    case TypePreference of
+        no_preference ->
+            [];
+        _ ->
+            Root = get_root(Config, Target, RuleConfig),
+            Predicate =
```
paulo-ferraz-oliveira, line 914: We should probably put this predicate in its own private function, just to ease the code flow.

src/elvis_style.erl
```diff
@@ -897,6 +902,37 @@ behaviour_spelling(Config, Target, RuleConfig) ->
...
+                   NodeType == type
+                   andalso lists:member(NodeName, [term, any])
+                   andalso NodeName /= TypePreference
+                end,
+            case elvis_code:find(Predicate, Root, #{traverse => all, mode => node}) of
+                [] ->
+                    [];
+                InconsistentTypeNodes ->
+                    ResultFun =
+                        fun(Node) ->
```
paulo-ferraz-oliveira, line 927: For the same reason above, we should probably separate this anon. function to its own private definition.

paulo-ferraz-oliveira (commented):
I ask for a few tweaks but I like the overall direction and implementation.

src/elvis_style.erl
```diff
@@ -99,9 +99,7 @@
 -define(ALWAYS_SHORTCIRCUIT_MSG,
         "Non-shortcircuiting operator (~p) found in line ~p. "
         "It's recommended to use ~p, instead.").
--define(TYPES_TERM_OR_ANY,
-        "The type in line ~p is not preferred, please use the "
-        "~p type.").
+-define(TYPES_TERM_OR_ANY, "Found ~p on line ~p, but your preferred type is ~p.").
```
paulo-ferraz-oliveira, line 102:
```suggestion
-define(TYPES_TERM_OR_ANY, "Found usage of type ~p() on line ~p, but your preferred type is ~p().").
```

paulo-ferraz-oliveira: You can commit directly from GitHub, too, if you want, by using button "Commit suggestion" above.

paulo-ferraz-oliveira:
I'm Ok with all the recent changes. I'll approve, assuming you'll eventually accept the only new suggestion I've added. I'll also add @elbrujohalcon as a reviewer so he can give this a look. Many thanks.

paulo-ferraz-oliveira (approved):

doc_rules/elvis_style/types_term_or_any.md
```diff
@@ -0,0 +1,15 @@
+# Types `term()` or `any()`
+
+Use `term()` or `any()` consistently for types in specs.
+
+> Works on `.beam` file? Yes.
+
+## Options
+
+- `prefer :: term | any`.
```
elbrujohalcon, line 9: We should specify which one is the default.

src/elvis_style.erl
```diff
@@ -99,6 +99,8 @@
 -define(ALWAYS_SHORTCIRCUIT_MSG,
         "Non-shortcircuiting operator (~p) found in line ~p. "
         "It's recommended to use ~p, instead.").
+-define(TYPES_TERM_OR_ANY,
+        "Found usage of type ~p() on line ~p, but your preferred type is ~p().").
```
elbrujohalcon, line 103:
```suggestion
        "Found usage of type ~p/0 on line ~p. Please use ~p/0, instead.").
```

RULES.md
```diff
@@ -46,6 +46,7 @@ identified with `(since ...)` for convenience purposes.
 - [Numeric Format](doc_rules/elvis_style/numeric_format.md)
 - [Operator Spaces](doc_rules/elvis_style/operator_spaces.md)
 - [State Record and Type](doc_rules/elvis_style/state_record_and_type.md)
+- [Types `term()` or `any()`](doc_rules/elvis_style/types_term_or_any.md)
```
elbrujohalcon, line 49:
```suggestion
- [Consistent Generic Type](doc_rules/elvis_style/consistent_generic_type.md)
```
☝🏻 I'm not sold on this title/name either, but… it would allow us to include `_` in the list if we ever want to also consider that a _generic_ type. I know it's not exactly the same… but in most scenarios, it serves the same purpose as `term()` or `any()`.

elbrujohalcon: In other words, I would like the name of this rule not to list all the available options it has.

doc_rules/elvis_style/types_term_or_any.md
elbrujohalcon: Also… can we call this parameter something like `preferred_type` or just `type`?

elbrujohalcon (changes requested):
I like the idea.
I think this rule is very similar to [Behaviour Spelling](https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/behaviour_spelling.md) and so I would expect its implementation, documentation, tests, etc. to be more similar to it..
That's the goal of all the suggestions I made.

Besides that, can you add more tests to showcase the different places where the type can be, like…

```erlang
% Type definitions when this is alone or combined
-type my_type() :: any().
-type combined() :: term() | my_type().

% Callback definitions (with or without when)
-callback my_callback(term()) -> any().
-callback my_callback(X) -> X when X :: term().

% Specs with when
-spec my_spec(X) -> X when X :: any().

% Record definitions
-record(my_record, {t :: term(), a :: any()}).
```

…and can you add tests where this should not be validated, like these…

```erlang
% A parametric type called any
-type any(Thing) :: Thing.

% A function called any
-spec any() -> term().
any() -> any.

% A function that calls the function called any
my_function() -> any().
```

doc_rules/elvis_style/types_term_or_any.md
elbrujohalcon: I disagree. I think the rule needs a default, so it can be added without specifying options, but the default should be either `term` or `any`.

src/elvis_style.erl
elbrujohalcon: HA! I just recommended changing it back to "Please use…" which would be more similar to other messages we have.

doc_rules/elvis_style/types_term_or_any.md
elbrujohalcon: …aaand I think the option names should be nouns. What you're choosing is the _preferred type_… so, in my mind, the option should be called `preferred_type` or just `type`, just like the one for _Behaviour Spelling_ is called `spelling`.

elbrujohalcon:
Sorry for the back and forth, @jackyhui96 … Multiple maintainers, multiple views on the project… 🤷🏻 
If you want, before considering my suggestions, wait for @paulo-ferraz-oliveira to reply to my comments so we can agree on what our preference is.

doc_rules/elvis_style/types_term_or_any.md
paulo-ferraz-oliveira: :) It was previously named `preference` and I asked to change to `prefer` (as a verb).

There was also a default (`no_preference`) which I asked to remove, because I'm not sure `elvis_core` should implement this as a default rule (i.e. active from the next release, as it'll force many to change - thus it's considered a breaking change).

src/elvis_style.erl
paulo-ferraz-oliveira: Yeah, I'd asked to change this, too. I believe "but your preferred type is" is more explicit, since it was the one actually chosen is the options (if we decide to keep the "no default" behaviour).

RULES.md
paulo-ferraz-oliveira: I guess this came from the issue title, but I'm Ok with your new suggestion.

src/elvis_style.erl
paulo-ferraz-oliveira: Does this mean you'd prefer to have a default and the rule be active from the next release?

paulo-ferraz-oliveira:
In any case, the only meaningful request to change I see is the fact that the rule should not be restricted to `term()` or `any()`, everything else are details, for me, that I have no strong opinions on:
1. whether the rule is active by default or not (in which case it would need a default),
2. whatever the output message is, as long as it's easy to act on.

I agree with adding more tests.

src/elvis_style.erl
elbrujohalcon: I wouldn't add it to the ruleset… but I would have a default value anyway.

elbrujohalcon:
> 1. whether the rule is active by default or not (in which case it would need a default),
I agree not to activate the rule by default (i.e. not to put it in the default ruleset `erl_files`) but, at the same time, it can totally have a default so users can add it just like…
```erlang
rules => [{elvis_style, whatever_name_it_ends_up_with, #{}}]
```
…without the need to specify a preferred value explicitly.

> 2. whatever the output message is, as long as it's easy to act on.
Yeah… easy to act on is fine, but consistent with other messages is important to me, too.

> I agree with adding more tests.
🤝

src/elvis_style.erl
paulo-ferraz-oliveira: Make sense. I should've noticed we have that for all rules. I imagine you're referring to activating it without options and letting `elvis_core` do its thing...

elbrujohalcon: Yeah… Exactly. My goal is for people to be able to _just add the rule_ without explicitly choosing a preferred type. Elvis should not fail in that situation, it should just either consider `term()` or `any()` or `_`… the preferred type.

elbrujohalcon:
Alright @jackyhui96 … You have green light to move forward with the changes. Sorry again for the back-and-forth.

doc_rules/elvis_style/types_term_or_any.md
paulo-ferraz-oliveira: I hadn't seen this or other "already resolved" conversations yesterday. GitHub is funny in the way it shows stuff.

doc_rules/elvis_style/types_term_or_any.md
paulo-ferraz-oliveira: Sure. I guess it would be nice to find time to write down some of these orientations so it's easier to review pull requests (and for people implementing them)... but... time!

paulo-ferraz-oliveira:
@elbrujohalcon, is the rule to be renamed as `consistent_generic_type`?

paulo-ferraz-oliveira:
I'm summarising the requested changes below:

- [ ] add a default (named `preferred_type`) to the rule (and update the doc. accordingly): the value is `term`
- [ ] change message to `Found usage of type ~p/0 on line ~p. Please use ~p/0, instead.`
- [ ] add more tests as per https://github.com/inaka/elvis_core/pull/259#pullrequestreview-1152596276
- [ ] doc. should state rule as "Consistent Generic Type": the file should be named `consistent_generic_type.md` and the rule `consistent_generic_type` for use

@elbrujohalcon, let me know if this is it, basically, otherwise I'll update this comment so it's less confusing.

elbrujohalcon:
comment updated, @paulo-ferraz-oliveira ;)

jackyhui96:
Currently it looks like the rule doesn’t apply to callbacks

elbrujohalcon:
@jackyhui96 you mean that `elvis_code:find/3` doesn't traverse callbacks?
If that's so, it's a bug and it's worth reporting.

elbrujohalcon (approved):

elbrujohalcon:
@paulo-ferraz-oliveira I already _approved_ the PR. I leave it up to you to merge it, if you're comfortable with it :)

jackyhui96:
> @jackyhui96 you mean that `elvis_code:find/3` doesn't traverse callbacks? If that's so, it's a bug and it's worth reporting it as such.
> 
> Or maybe elements within callbacks are not categorized as types… 🤔 … If that's so, same thing: it merits a bug report.

Yeah it doesn't categorise them as types 😞 this is what the nodes look like for a callback:
```
#{attrs => #{location => {13,1},  text => "-"},           type => '-'},
#{attrs => #{location => {13,2},  text => "callback"},    type => atom, value => callback},
#{attrs => #{location => {13,11}, text => "my_callback"}, type => atom, value => my_callback},
#{attrs => #{location => {13,22}, text => "("},           type => '('},
#{attrs => #{location => {13,23}, text => "term"},        type => atom, value => term},
#{attrs => #{location => {13,27}, text => "("},           type => '('},
#{attrs => #{location => {13,28}, text => ")"},           type => ')'},
#{attrs => #{location => {13,29}, text => ")"},           type => ')'},
#{attrs => #{location => {13,31}, text => "->"},          type => '->'},
#{attrs => #{location => {13,34}, text => "term"},        type => atom, value => term},
#{attrs => #{location => {13,38}, text => "("},           type => '('},
#{attrs => #{location => {13,39}, text => ")"},           type => ')'},
#{attrs => #{location => {13,40}, text => ".\n"},         type => dot},
```

elbrujohalcon:
Oooooh… that's bad. Do you mind filing a bug for that, @jackyhui96 ?

jackyhui96:
@elbrujohalcon sure thing, is this a bug with elvis or with `ktn_code`?

doc_rules/elvis_style/consistent_generic_type.md
```diff
@@ -0,0 +1,16 @@
+# Consistent Generic Type
+
+Use `term()` or `any()` consistently for types in specs.
+
+> Works on `.beam` file? Yes.
+
+## Options
+
+- `preferred_type :: term | any`.
+- default: `term`.
```
paulo-ferraz-oliveira, line 10:
```suggestion
  - default: `term`.
```

src/elvis_style.erl
```diff
@@ -99,8 +99,8 @@
 -define(ALWAYS_SHORTCIRCUIT_MSG,
         "Non-shortcircuiting operator (~p) found in line ~p. "
         "It's recommended to use ~p, instead.").
--define(TYPES_TERM_OR_ANY,
-        "Found usage of type ~p() on line ~p, but your preferred type is ~p().").
+-define(CONSISTENT_GENERIC_TYPE,
+        "Found usage of type ~p/0 on line ~p. Please use ~p/0, instead.").
```
paulo-ferraz-oliveira, line 103: @elbrujohalcon, is `/0` really more readable than `()` here? Because the consumer will not find `/0`, but they will find `()`.

paulo-ferraz-oliveira:
@elbrujohalcon, I left a minor comment (hopefully the last 😄).

src/elvis_style.erl
paulo-ferraz-oliveira: Solved over Slack... we'll keep as-is.

paulo-ferraz-oliveira:
@jackyhui96, I'm merging... many thanks for this.

PR #262, opened 2023-02-03
Fix macro name resolution (`call` gets highest precedence)

filmor:
The `call` node has to be used with highest precedence when looking for
the macro name, as otherwise `-define(CALL(A), some_atom).` will pick
`some_atom` as the macro name instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,14 +14,14 @@ jobs:
     strategy:
       matrix:
         otp_vsn: [23, 24, 25]
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-20.04, windows-latest]
     steps:
       - uses: actions/checkout@v3
       - uses: erlef/setup-beam@v1
         id: setup-beam
         with:
           otp-version: ${{matrix.otp_vsn}}
-          rebar3-version: '3.18'
+          rebar3-version: '3'
       - name: Restore _build
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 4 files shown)


elbrujohalcon (approved):
Great catch!

elbrujohalcon:
We need to use Ubuntu-20.04 instead of Ubuntu-latest for CI.
Do you mind including that change in this PR, @filmor ?

paulo-ferraz-oliveira:
> We need to use Ubuntu-20.04 instead of Ubuntu-latest for CI. Do you mind including that change in this PR, @filmor ?

You can also wait for the other PR to be merged, and then rebase this one, @filmor.

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         otp_vsn: [23.3, 24.3, 25.0]
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-20.04, windows-latest]
```
paulo-ferraz-oliveira, line 17: I don't mind this change here, but it would be better suite on the [other pull request](https://github.com/inaka/elvis_core/pull/263), on top of which this one could build, with a `git rebase`.

filmor: Sure, will do that.

filmor:
You'll either need to adjust the expected status or I'll put slightly more work into this (tomorrow :)) to make the `matrix.os` part static.

.github/workflows/ci.yml
paulo-ferraz-oliveira: Thanks...

PR #263, opened 2023-02-03
Tweak CI target versions

filmor:


elbrujohalcon (commented):
@paulo-ferraz-oliveira you are the expert in this area. What do you think?

paulo-ferraz-oliveira:
The change seems reasonable. In practice, what it means is that we'll let `setup-beam` choose the "latest" version. @filmor, could you also change `rebar3-version` to `3`, in that case? Thanks.

paulo-ferraz-oliveira:
OTP 23 is not available on `ubuntu-latest`, we need to either drop it (less desirable) or change to `ubuntu-20.04`. @filmor, would you be so kind to change that too? Thanks.

elbrujohalcon (approved):

Issue #268, opened 2023-02-27
New Rule: No specs in header files


elbrujohalcon:
### Brief Description

Just like #162 but for specs.

### Should be on by default?

**YES**

### Reasoning

Same as #162.

### Refactoring Proposal

Move the specs to where they belong.

### Examples

```erlang
%% bad.hrl
-spec a(function) -> _.

-define(A_MACRO, here).
```

```erlang
%% good.hrl
-define(A_MACRO, here).
```

Issue #162, opened 2014-08-07
Rule idea: no `-type`s in .hrl files


erszcz:
Defining types in public header files (especially those intended for inclusion via `-include_lib()`) might lead to type name clashes between projects and even modules of a single big project.
Instead, types should be defined in modules which they correspond to (with `-export_type()`) and this way take advantage of the namespacing offered by module names.
In other words, "no type definitions in header files" rule means that we will always need to use `some_mod:some_type()` unless referring to a type from the same module it's defined in.

paulo-ferraz-oliveira:
Another good one.

elbrujohalcon:
We should split `hrl_files` from `erl_files` ruleset… Which is something we should've done already a long time ago anyway :)

paulo-ferraz-oliveira:
There's not many `hrl`-specific rules, at the moment, are there? In any case, I agree, it's a good suggestion. It can be breaking, however, but we can always add a warning and keep current behaviour and break when we have enough stuff to go to 2.x

PR #269, opened 2023-02-27
New ruleset: hrl_files

elbrujohalcon:
As a first step towards #162 and #268, I'm creating the `hrl_files` ruleset here.
I basically copied everything from `erl_files` and removed the ones that didn't seem to make sense on hrl files.
To be fair, I would've kept all rules because _dirty people_ can do _dirty things_ on hrl files, but… well… we can improve that list over time if we want to, anyway.


test/elvis_SUITE.erl
```diff
@@ -14,14 +14,10 @@
...
          rock_with_rule_groups/1, rock_this_skipping_files/1, rock_this_not_skipping_files/1,
-         rock_with_umbrella_apps/1, custom_ruleset/1, throw_configuration/1,
+         rock_with_umbrella_apps/1, custom_ruleset/1, hrl_ruleset/1, throw_configuration/1,
          find_file_and_check_src/1, find_file_with_ignore/1, invalid_file/1, to_string/1,
          chunk_fold/1]).
 
-                                 %% Rocking
-
-         %% Util & Config
-
```
elbrujohalcon, line 24: Leftovers from when we formatted the code using the formatter 🤷🏻

paulo-ferraz-oliveira (approved):
Would you create an issue so that the Wiki's updated later according to these changes?

elbrujohalcon:
I'll update the wiki right now, don't worry.

elbrujohalcon:
The wiki was updated, but I wonder if we shouldn't just close the wiki in favor of RULES.md

PR #273, opened 2023-02-28
Fix #265: New Rule: export used types

### LINKED ISSUE #265: New Rule: Export All Used Types
### Brief Description

The idea is to emit a warning every time a spec for an exported function uses a _local_ type that's not exported.

### Should be on by default?

**YES**

### Reasoning

If you put a type in a spec of an exported function, that means that people writing modules that call that function should know what that type is and they may as well pass instances of that type around before calling your function. So, to add their own specs, and also to keep dialyzer messages cleaner, they should put `your_mod:your_type()` in their specs, instead of the actual definition of the type.
To be able to do that, the type must be exported.

### Refactoring Proposal

Add the non-exported types to an `-export_type(…).` attribute.

### Examples

```erlang
-module bad.

-type my_type() :: my | type.

-export [my_fun/1].

-spec my_fun(my_type()) -> my_type().
my_fun(my) -> type;
my_fun(type) -> my.
```

```erlang
-module good.

-type my_type() :: my | type.
-export_type [my_type/0].

-export [my_fun/1].

-spec my_fun(my_type()) -> my_type().
my_fun(my) -> type;
my_fun(type) -> my.
```

maco:


RULES.md
```diff
@@ -52,6 +52,7 @@ identified with `(since ...)` for convenience purposes.
 - [Consistent Generic Type](doc_rules/elvis_style/consistent_generic_type.md)
 - [Used Ignored Variable](doc_rules/elvis_style/used_ignored_variable.md)
 - [Variable Naming Convention](doc_rules/elvis_style/variable_naming_convention.md)
+- [Export Used Type](doc_rules/elvis_style/export_used_types.md)
```
elbrujohalcon, line 55:
```suggestion
- [Export Used Types](doc_rules/elvis_style/export_used_types.md)
```

src/elvis_core.erl
```diff
@@ -9,6 +9,8 @@
 %% for eating our own dogfood
 -export([main/1]).
 
+-export_type([target/0]).
```
elbrujohalcon, line 12: NIIIIIIIICE!!!

src/elvis_rulesets.erl
```diff
@@ -69,7 +69,8 @@ rules(erl_files) ->
                {elvis_style, no_author},
                {elvis_style, no_catch_expressions},
                {elvis_style, numeric_format},
-               {elvis_style, behaviour_spelling}]);
+               {elvis_style, behaviour_spelling},
+               {elvis_style, export_used_types}]);
```
elbrujohalcon, line 73: Can you also add it to `beam_files`, please?

doc_rules/elvis_style/export_used_types.md
```diff
@@ -0,0 +1,17 @@
+# Export Used Types
+
+(since [2.1.0](https://github.com/inaka/elvis_core/releases/tag/2.1.0))
+
+Exporting a function without exporting the types it depends on can result in
+reimplementing the types in every module that uses those functions. To avoid
+this, when a function is exported, its types should be too.
+
```
elbrujohalcon, line 8: Around here there should be something that says "Works on beam files? Yes"

elbrujohalcon (changes requested):
Great job! The only thing missing, from my perspective, is to make the rule run on beam files and test that.

src/elvis_style.erl
```diff
@@ -104,6 +113,8 @@
         "It's recommended to use ~p, instead.").
 -define(CONSISTENT_GENERIC_TYPE,
         "Found usage of type ~p/0 on line ~p. Please use ~p/0, instead.").
+-define(EXPORT_USED_TYPES,
+        "Type ~p/~p, defined on line ~p, is used by an exported function but not exported itself").
```
elbrujohalcon, line 117:
```suggestion
-define(EXPORT_USED_TYPES_MSG,
        "Type ~p/~p, defined on line ~p, is used by an exported function but not exported itself").
```

paulo-ferraz-oliveira:
Should we mention, somewhere, that exported types can also be `opaque`, as a general guidance?

src/elvis_core.erl
paulo-ferraz-oliveira: We could probably make this opaque (?)

src/elvis_style.erl
paulo-ferraz-oliveira: Some elements above it are missing that `_MSG` too 😢

src/elvis_core.erl
elbrujohalcon: It can't be opaque… people must know how to _build_ an instance of `target/0`. If we make it opaque we will need to create a function like `new_target(source_filename() | module()) -> target().`.
To me it's an overkill and it's also super confusing.

erszcz:
@paulo-ferraz-oliveira 

> Should we mention, somewhere, that exported types can also be `opaque`, as a general guidance?

I'm not sure of the intention of the question as a bystander, but FWIW opaque types have to be exported to be useful at all. In the local module they're just types (i.e. you can pattern match and construct an opaque in the module that defines it without constraints), they only have special treatment when referred to as external types, but for that to be possible, they have to be exported.

paulo-ferraz-oliveira:
> but FWIW opaque types have to be exported to be useful at all

I know 😄, there's even a compiler warning if I'm not mistaken.

What I'm hinting at is that this rule will make consumers just go "Oh, I need to add an export_type here" and do it without thinking about the consequences of potentially exposing a non-opaque. If it was they're intention before, sure, the type should already be opaque and exposed; if it wasn't, now that it's being exposed there are consequences to it.

src/elvis_core.erl
paulo-ferraz-oliveira: In this case you're right. Since I had no context I (wrongly) assumed this was a return type, and it's an input type. Agreed on the non-required/confusing use of `new_target(...)`.

doc_rules/elvis_style/export_used_types.md
```diff
@@ -0,0 +1,17 @@
+# Export Used Types
+
+(since [2.1.0](https://github.com/inaka/elvis_core/releases/tag/2.1.0))
```
elbrujohalcon, line 3:
```suggestion
(since [3.0.0](https://github.com/inaka/elvis_core/releases/tag/3.0.0))
```

elbrujohalcon (changes requested):

src/elvis_style.erl
```diff
@@ -1005,6 +1017,53 @@ always_shortcircuit(Config, Target, RuleConfig) ->
...
+                        (_, Acc) ->
+                            Acc
+                    end,
+                    #{},
+                    AllTypes),
+
+    % Report
+    lists:map(fun({Name, Arity} = Info) ->
+                 Line = maps:get(Info, LineNumbers, unknown),
+                 elvis_result:new(item, ?EXPORT_USED_TYPES, [Name, Arity, Line], Line)
```
maco, line 1063:
```suggestion
                 elvis_result:new(item, ?EXPORT_USED_TYPES_MSG, [Name, Arity, Line], Line)
```

because if it's changed up top…

elbrujohalcon (approved):

elbrujohalcon:
Once we have a new release, we should run it on OTP itself.
This rule in particular encapsulates something that [people were asking for](https://erlangforums.com/t/it-would-be-nice-if-more-otp-modules-included-export-type/2152) a while back.

paulo-ferraz-oliveira:
[Here's a pointer](https://github.com/erlang/otp/pull/2794#issuecomment-708379870) from the OTP team and exporting types: "Our policy is to be restrictive with exporting types. Only input and return types should be necessary to export." which correlates with the rule.

PR #274, opened 2023-02-28
Fix #266: New Rule: Consistent Variable Casing

### LINKED ISSUE #266: Consistent Variable Casing
### Brief Description

Use the same _casing_ for all variables with the same name within a module. E.g., emit a warning when Elvis finds both `Userid` and `UserId`or `UserID` in a module.

### Should be on by default?

**YES**

### Reasoning

If two variables have the same name, it's very likely that they refer to the same concept and, therefore, it would make the lives of code readers easier if they share the exact same name, including casing and all.

### Refactoring Proposal

Rename one or more of the variables.

### Examples

```erlang
-module bad.

-export [my_fun/1, my_fun/2].

my_fun(MyVar) -> my_fun(MyVar, too).

my_fun(Myvar, too) -> {Myvar, too}.
```

```erlang
-module good.

-export [my_fun/1, my_fun/2].

my_fun(MyVar) -> my_fun(MyVar, too).

my_fun(MyVar, too) -> {MyVar, too}.
```

elbrujohalcon:


RULES.md
```diff
@@ -52,6 +52,7 @@ identified with `(since ...)` for convenience purposes.
 - [Consistent Generic Type](doc_rules/elvis_style/consistent_generic_type.md)
 - [Used Ignored Variable](doc_rules/elvis_style/used_ignored_variable.md)
 - [Variable Naming Convention](doc_rules/elvis_style/variable_naming_convention.md)
+- [Consistent Variable Casing](doc_rules/elvis_style/consistent_variable_casing.md)
```
paulo-ferraz-oliveira, line 55: Could these be sorted alphabetically?

paulo-ferraz-oliveira:
I like this rule.

RULES.md
elbrujohalcon: Well… they were not sorted already… Do you mind writing a different ticket to sort them out?

paulo-ferraz-oliveira (approved):

RULES.md
elbrujohalcon: https://github.com/inaka/elvis_core/issues/276

paulo-ferraz-oliveira: Yeah, they were only for a short time, after I created that file :)

PR #292, opened 2023-03-01
Fix #286: New functions for max arity

### LINKED ISSUE #286: Max Function Arity
### Brief Description

Warn if a function has too many arguments.

### Should be on by default?

**YES**

### Options

-  `max_arity :: pos_integer()` (default: `5`)

**Note:** I picked `5` as the default basically randomly. If the implementor can do a proper analysis of some public Erlang codebases to determine what a _reasonable_ default should be, much better.

### Reasoning

Quoting from Credo docs:
> A function can take as many parameters as needed, but even in a functional language, there can be too many parameters.

### Origin (#281)

Inspired by the `Refactor.FunctionArity` rule from [Credo](https://hexdocs.pm/credo/Credo.Check.Refactor.FunctionArity.html)

elbrujohalcon:
I ended up creating two rules: one for regular functions (with 8 as the default) and one for anonymous ones (with 3 as the default max arity).


elbrujohalcon:
Yeah… I thought about that and it's more flexible this way and also the code to implement each rule is not hugely different but also not exactly equal. I'll keep it as 2 rules.

PR #298, opened 2023-03-01
Fix #290: New Rule: no_single_clause_case

### LINKED ISSUE #290: New rule: No Single Clause Case Statements
### Brief Description

The following code…

```erlang
case do:something() of
    {ok, Result} -> do:something(else)
end
```

…should actually be written using pattern-matching:

```erlang
{ok, Result} = do:something(),
do:something(else)
```

The `case` statement is unnecessary and the only difference between the two is that the exception raised in the case of error changes from `case_clause` to `badmatch`.

### Should be on by default?

**YES**

### Options

- None.

### Reasoning

Paraphrasing Elm-Review documentation:

> […] single-pattern case expressions […] may be written more concisely or removed entirely.

### Origin (#281)

Inspired by the `NoSinglePatternCase` rule from [elm-review](https://package.elm-lang.org/packages/SiriusStarr/elm-review-no-single-pattern-case/1.0.1/NoSinglePatternCase)

elbrujohalcon:


doc_rules/elvis_style/no_single_clause_case.md
```diff
@@ -0,0 +1,30 @@
+# No Single-Clause Case Statements
```
paulo-ferraz-oliveira, line 1: I'm Ok with the title, but usually it follows the rule name closely.

elbrujohalcon: Yeah, but `no_single_clause_case_statements` was too long for a rule name, IMHO.

paulo-ferraz-oliveira: Yeah, what I meant was the title could follow the rule name, not the other way around :)

PR #300, opened 2023-03-01
Fix #264: New Rule: Private Data Types

### LINKED ISSUE #264: New Rule: If a type is a record, it should be either opaque or not exported
### Brief Description

If your module defines a record and it also defines a type for it, that type should not be exported/accessible from the outside.

### Should be on by default?

**YES**

### Reasoning

Since the record is defined within the module, nobody from outside the module should _know_ how to build an instance of that record. That means that the type associated with the record should not be _public_.
It can be an opaque type, so that other modules can pass around record instances but not [de]construct them. Or it can just be not exported so no modules other than the current one have knowledge about the record.

#### Examples

```erlang
-module bad.
-record my_rec, {a, b, c}.

-type my_rec() :: #my_rec{}.
-export_type [my_rec/0].
```

```erlang
-module good.
-record my_rec, {a, b, c}.

-opaque my_rec() :: #my_rec{}.
-export_type [my_rec/0].
```

```erlang
-module also_good.
-record my_rec, {a, b, c}.

-type my_rec() :: #my_rec{}.
```

maco:
Fixes #264


doc_rules/elvis_style/private_data_types.md
```diff
@@ -0,0 +1,23 @@
...
+externally results in tightly-coupled code. A module's should be responsible
+for creating its own internal types. So, if the data type is needed outside the
+module, it should be
+[an opaque type](https://www.erlang.org/doc/reference_manual/opaques.html).
+
+> "Works on `.beam` file? Yes!"
+
+## Options
+
+- `applies_to :: [atom()]`
```
elbrujohalcon, line 15:
```suggestion
- `applies_to :: [record | map | tuple]`
```

src/elvis_style.erl
```diff
@@ -1217,10 +1210,103 @@ export_used_types(Config, Target, RuleConfig) ->
...
+is_tuple_type(#{type := type_attr,
+                node_attrs := #{type := #{attrs := #{name := tuple}}}}) ->
+    true;
+is_tuple_type(_Node) ->
+    false.
+
+is_map_type(#{type := type_attr, node_attrs := #{type := #{attrs := #{name := map}}}}) ->
+    true;
+is_map_type(_Node) ->
+    false.
```
elbrujohalcon, line 1229:
```suggestion
get_type_of_type(#{type := type_attr,
                 node_attrs := #{type := #{attrs := #{name := TypeOfType}}}}) ->
    TypeOfType;
get_type_of_type(_) ->
    undefined.
```

…or…

```suggestion
is_type_of_type(TypeOfType, #{type := type_attr,
                 node_attrs := #{type := #{attrs := #{name := TypeOfType}}}}) ->
    true;
is_type_of_type(_) ->
    false.
```

src/elvis_style.erl
```diff
@@ -1217,10 +1210,103 @@ export_used_types(Config, Target, RuleConfig) ->
...
+
+-type data_type() :: record | map | tuple.
+-type private_data_type_config() :: #{applies_to => [data_type()]}.
```
elbrujohalcon, line 1232:
```suggestion
-type private_data_type() :: record | map | tuple.
-type private_data_type_config() :: #{applies_to => [private_data_type()]}.
```
(Just to avoid possible collisions with other interpretations of `data_type/0` that may pop up in the future)

src/elvis_style.erl
```diff
@@ -1217,10 +1210,103 @@ export_used_types(Config, Target, RuleConfig) ->
...
+%     TupleTypes =
+%         [get_type_declared_name_arity(Node)
+%             || Node
+%                 <- elvis_code:find(fun is_tuple_type/1,
+%                                     TreeRootNode,
+%                                     #{traverse => all, mode => node})],
+%     lists:filter(fun({Name, Arity}) ->
+%         lists:member({Name, Arity}, ExportedTypes)
+%         end,
+%         TupleTypes).
```
elbrujohalcon: Please remove these commented-out code blocks. 🙏🏻

src/elvis_style.erl
```diff
@@ -1217,10 +1210,103 @@ export_used_types(Config, Target, RuleConfig) ->
...
+get_type_line_numbers(TreeRootNode) ->
+    AllTypes =
+        elvis_code:find(fun is_type_attribute/1, TreeRootNode, #{traverse => all, mode => node}),
+    lists:foldl(fun (#{attrs := #{location := {Line, _}, name := Name},
+                       node_attrs := #{args := Args}},
+                     Acc) ->
+                        maps:put({Name, length(Args)}, Line, Acc);
+                    (_, Acc) ->
+                        Acc
+                end,
```
elbrujohalcon, line 1284: Maybe we can encapsulate all or part of this logic in a `get_type_arity/1` function, either here or in `elvis_code`.
And then you could use `{Line, _} = ktn_code:attr(location, Node)` as we do in other functions within this module instead of nested pattern-matching.

src/elvis_style.erl
```diff
@@ -1217,10 +1210,103 @@ export_used_types(Config, Target, RuleConfig) ->
...
+public_data_types(Type, TreeRootNode, ExportedTypes) ->
+    FunMap =
+        #{record => fun is_record_type/1,
+          map => fun is_map_type/1,
+          tuple => fun is_tuple_type/1},
+    Fun = maps:get(Type, FunMap),
+    Types =
+        [get_type_declared_name_arity(Node)
+         || Node <- elvis_code:find(Fun, TreeRootNode, #{traverse => all, mode => node})],
+    lists:filter(fun({Name, Arity}) -> lists:member({Name, Arity}, ExportedTypes) end, Types).
```
elbrujohalcon, line 1264: If you replace all the `is_*_type/1` functions by `get_type_of_type/1`, following my comment above, you could traverse the whole tree just once (instead of using `lists:flatmap/2` in line 1243:

```suggestion
public_data_types(TypesToCheck, TreeRootNode, ExportedTypes) ->
    Fun = fun(Node) -> lists:member(get_type_of_type(Node), TypesToCheck) end,
    Types =
        [get_type_declared_name_arity(Node)
         || Node <- elvis_code:find(Fun, TreeRootNode, #{traverse => all, mode => node})],
    lists:filter(fun({Name, Arity}) -> lists:member({Name, Arity}, ExportedTypes) end, Types).
```

elbrujohalcon: …or if you prefer `is_type_of_type/2`…

```suggestion
public_data_types(TypesToCheck, TreeRootNode, ExportedTypes) ->
    Fun = fun(Node) -> lists:any(fun(TypeToCheck) -> is_type_of_type(TypeToCheck, Node) end, TypesToCheck) end,
    Types =
        [get_type_declared_name_arity(Node)
         || Node <- elvis_code:find(Fun, TreeRootNode, #{traverse => all, mode => node})],
    lists:filter(fun({Name, Arity}) -> lists:member({Name, Arity}, ExportedTypes) end, Types).
```

…or you can even turn `is_type_of_type/2` into `member_of_types/2` and pass the list directly…

```suggestion
public_data_types(TypesToCheck, TreeRootNode, ExportedTypes) ->
    Fun = fun(Node) -> member_of_types(Node, TypesToCheck) end,
    Types =
        [get_type_declared_name_arity(Node)
         || Node <- elvis_code:find(Fun, TreeRootNode, #{traverse => all, mode => node})],
    lists:filter(fun({Name, Arity}) -> lists:member({Name, Arity}, ExportedTypes) end, Types).
```

elbrujohalcon: My preferred way would be to use `get_type_of_type/1`, but any of the solutions is good.
OTOH… Following this module's usual structure… I would move that auxiliary function (`get_type_of_type/1` or `is_type_of_type/2` or whatever you choose) **below** the rule where it is used.

test/examples/fail_private_data_types.erl
```diff
@@ -0,0 +1,11 @@
+-module(fail_private_data_types).
+-hank ignore.
```
elbrujohalcon, line 2: It's not terribly bad to use `-hank ignore.` but maybe you could ignore only the specific warnings instead of all of them.
In any case, personally… I prefer to write some extra dumb code to avoid the warnings instead of ignoring them.
Your module will look more _realistic_ if it had functions using your types… even if those functions are super-dumb and don't do anything really.

test/style_SUITE.erl
```diff
@@ -26,7 +26,8 @@
          verify_no_author/1, verify_no_import/1, verify_no_catch_expressions/1,
          verify_no_single_clause_case/1, verify_numeric_format/1, verify_behaviour_spelling/1,
          verify_always_shortcircuit/1, verify_consistent_generic_type/1, verify_no_types/1,
-         verify_no_specs/1, verify_export_used_types/1, verify_consistent_variable_casing/1]).
+         verify_no_specs/1, verify_export_used_types/1, verify_consistent_variable_casing/1,
+         verify_private_data_types/1]).
```
elbrujohalcon, line 30: You should also add a `verify_elvis_attr_private_data_types/1` (like the ones in the export list below) testing the `-elvis(…).` attribute usage with your rule parameters.

test/style_SUITE.erl
```diff
@@ -1520,6 +1521,46 @@ verify_export_used_types(Config) ->
...
+    PathPass2 = "pass_private_data_types2." ++ Ext,
+    [] =
+        elvis_core_apply_rule(Config,
+                              elvis_style,
+                              private_data_types,
+                              #{applies_to => [record, map, tuple]},
+                              PathPass2),
+    % Default applies only to records
+    PathFail = "fail_private_data_types." ++ Ext,
+    [#{line_num := 5}] =
```
elbrujohalcon: It's probably better (particularly for rules that run on beam files like this one) to pattern-match on something other than line numbers (e.g. `[#{info := ["my_rec", 0, _]}]`) or just the number of warnings (`[_]`).

elbrujohalcon (changes requested):

RULES.md
```diff
@@ -58,6 +58,7 @@ identified with `(since ...)` for convenience purposes.
 - [Used Ignored Variable](doc_rules/elvis_style/used_ignored_variable.md)
 - [Variable Naming Convention](doc_rules/elvis_style/variable_naming_convention.md)
 - [Export Used Types](doc_rules/elvis_style/export_used_types.md)
+- [Private Data Types](doc_rules/elvis_style/private_data_types.md)
```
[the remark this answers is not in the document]
maco: OOOH

doc_rules/elvis_style/private_data_types.md
```diff
@@ -0,0 +1,23 @@
+# Private Data Types
+
+(since [3.0.0](https://github.com/inaka/elvis_core/releases/tag/3.0.0))
+
+Exporting a function's internal data type in order to construct those types
+externally results in tightly-coupled code. A module's should be responsible
```
paulo-ferraz-oliveira, line 6:
```suggestion
externally results in tightly-coupled code. A module should be responsible
```

doc_rules/elvis_style/private_data_types.md
```diff
@@ -0,0 +1,23 @@
+# Private Data Types
+
+(since [3.0.0](https://github.com/inaka/elvis_core/releases/tag/3.0.0))
+
+Exporting a function's internal data type in order to construct those types
```
paulo-ferraz-oliveira, line 5:
```suggestion
Exporting a function's internal data types in order to construct those types
```

doc_rules/elvis_style/private_data_types.md
```diff
@@ -0,0 +1,23 @@
...
+externally results in tightly-coupled code. A module's should be responsible
+for creating its own internal types. So, if the data type is needed outside the
+module, it should be
+[an opaque type](https://www.erlang.org/doc/reference_manual/opaques.html).
```
paulo-ferraz-oliveira, line 9:
```suggestion
Exporting functions' internally-defined data types, in order to consume those
types externally, results in tightly-coupled code. Modules should be responsible
for defining their own internal data types. If these are needed outside the
modules, they should be made
[opaque](https://www.erlang.org/doc/reference_manual/opaques.html).
```

doc_rules/elvis_style/private_data_types.md
```diff
@@ -0,0 +1,23 @@
...
+
+> "Works on `.beam` file? Yes!"
+
+## Options
+
+- `applies_to :: [atom()]`
+  - default: `record`.
+  - other options: `map` and `tuple`.
```
paulo-ferraz-oliveira, line 17:
```suggestion
```

paulo-ferraz-oliveira: Not required, since we're already mentioning all options above.

src/elvis_style.erl
```diff
@@ -25,11 +26,12 @@
...
               no_import_config/0, no_catch_expressions_config/0, numeric_format_config/0,
-              no_single_clause_case_config/0, consistent_variable_casing_config/0]).
+              no_single_clause_case_config/0, consistent_variable_casing_config/0,
+              private_data_type_config/0]).
 
 -define(INVALID_MACRO_NAME_REGEX_MSG,
-        "The macro named ~p on line ~p does not respect the format "
-        "defined by the regular expression '~p'.").
+        "The macro named ~p on line ~p does not respect the format defined "
+        "by the regular expression '~p'.").
```
paulo-ferraz-oliveira, line 34: What happened here? Was this the formatter? There seem to be a lot of these string-related changes that don't feel related to the pull request.

doc_rules/elvis_style/private_data_types.md
paulo-ferraz-oliveira, line 15: I'd rename the option `apply_to`. Thoughts?

elbrujohalcon: I'm fine with `applies_to`. I read it as "The rule applies to…".

I don't mind `apply_to` (as in "apply this rule to…"), but if it's a matter of taste, I would keep "applies".

paulo-ferraz-oliveira: I'd say it's a question of consistency:

https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/max_function_length.md 
- count, count

https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/no_macros.md
- allow

https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_text_style/no_trailing_whitespace.md
- ignore

elbrujohalcon: Good point.

src/elvis_style.erl
maco: Yep, formatter.

elbrujohalcon: Why didn't that happen on previous PRs, tho? 🤔

paulo-ferraz-oliveira: > Why didn't that happen on previous PRs, tho? 🤔

My thoughts exactly. Unless @maco's running a different version locally (but that would show up in CI - if there was inconsistency).

doc_rules/elvis_style/private_data_types.md
paulo-ferraz-oliveira: cf768b4

src/elvis_style.erl
maco: I _am_ running an old version of rebar apparently, but yeah, the CI part of that.

src/elvis_style.erl
maco: chatted on Slack, so summary:
- finding a type's arity is done differently depending on if it's in a `-type` line or `-spec` line
- this function is to find where the type was declared, not where it was used

I'm adjusting function names/comments

elbrujohalcon (approved):

PR #302, opened 2023-03-02
Fix #301: Handle maybe matches

### LINKED ISSUE #301: `rebar3 lint` unexpectedly warns `_Var ?= val` as `used_ignored_variable`
## Bug Description

The `_Var ?= val` expression in the following code seems valid but if `$ rebar3 lint` command is executed, it warns as the expression violates `used_ignored_variable` rule.

```erlang
-module(foo).

-export([foo/0]).

foo() ->
    maybe
        _ ?= val,  % OK
        _Var = val, % OK
        _Var ?= val  % NG
    end.
```

## To Reproduce

[//]:# (Steps to reproduce the behavior)

```console
// Create a new Erlang library and setup rebar.config
$ rebar3 new lib foo
$ cd foo/
$ echo '{erl_opts, [{feature, maybe_expr, enable}]}. {plugins, [rebar3_lint]}.' > rebar.config 

// Create an Erlang file (with the above code)
$ vim src/foo.erl

// Run linter
$ rebar3 lint
# src/foo.erl [FAIL]
  - used_ignored_variable (https://github.com/inaka/elvis_core/tree/main/doc_rules/elvis_style/used_ignored_variable.md)
    - Ignored variable is being used on line 9 and column 9.
```

## Expected Behavior

`$ rebar3 lint` should pass.

## `rebar3` Logs

[//]:# (If applicable, run `rebar3` with `DIAGNOSTIC=1` and attach all the logs to your report.)

## Additional Context

- OS: MacOS
- Erlang version: OTP 25.2.3
- rebar3 version: 3.18.0

elbrujohalcon:
This PR fixes both #301 and #303.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -22,8 +22,6 @@
   debug_info,
   {feature, maybe_expr, enable}]}.
 
-{minimum_otp_vsn, 23}.
-
 {profiles,
  [{test,
    [{extra_src_dirs, ["test/examples"]},
--- a/rebar.config
+++ b/rebar.config
@@ -29,25 +29,6 @@
     {erl_opts, [nowarn_missing_spec, nowarn_export_all]},
     {dialyzer, [{warnings, [no_return, unmatched_returns, error_handling]}]}]}]}.
 
-%% == Common Test ==
-
-{ct_compile_opts,
- [warn_unused_vars,
-  warn_export_all,
-  warn_shadow_vars,
-  warn_unused_import,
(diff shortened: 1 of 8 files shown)


pablocostass (approved):
Just dropping by to review this PR as per request of Brujo :)

tak30 (approved):

paulo-ferraz-oliveira (approved):
❤️

PR #304, opened 2023-03-02
Fix #289: New Rule: No match in conditions

### LINKED ISSUE #289: New rule: No Match in Condition
### Brief Description

Pattern-matching should not be used in `case` statement conditions, like this:

```erlang
case #{a := A} = do:something() of
    #{b := good} -> {a, really, nice, A};
    #{b := bad} -> {"not", a, good, A}
end
```

### Should be on by default?

**YES**

### Options

-  None

### Reasoning

While the code as written above is valid, it's much harder to understand (particularly for large statements) than the one below.

### Refactoring Proposal

```erlang
case do:something() of
    #{a := A, b := good} -> {a, really, nice, A};
    #{a := A, b := bad} -> {"not", a, good, A}
end
```

### Origin (#281)

Inspired by the `Refactor.MatchInCondition` rule from [Credo](https://hexdocs.pm/credo/Credo.Check.Refactor.MatchInCondition.html)

elbrujohalcon:
Note that I did my analysis and only `case` statement suffer from this problem, since:

* `if`s have no expression outside of the clauses
* same goes for `receive`s, `maybe`s and functions
* `try`s can totally have matches in their expressions. That's perfectly fine.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -41,6 +41,7 @@ identified with `(since ...)` for convenience purposes.
 - [No If Expression](doc_rules/elvis_style/no_if_expression.md)
 - [No Import](doc_rules/elvis_style/no_import.md)
 - [No Macros](doc_rules/elvis_style/no_macros.md)
+- [No Match in Condition](doc_rules/elvis_style/no_match_in_condition.md)
 - [No Nested try...catch Blocks](doc_rules/elvis_style/no_nested_try_catch.md)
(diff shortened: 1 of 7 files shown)


pablocostass (approved):
Just dropping by to review this PR as per request of Brujo :)

tak30 (approved):

paulo-ferraz-oliveira (approved):
Seems reasonable.

PR #305, opened 2023-03-02
Fix #284: New Rule: Param Pattern Matching

### LINKED ISSUE #284: Consistent Param Pattern-Matching
### Brief Description

The idea is to emit a warning in situations like the following ones (notice the duplication of keys everywhere):

```erlang
Map = #{key1 => value1, key2 => value2, key1 => value3}.
OtherMap = Map#{key2 := new_value2, key3 => new_value3, key2 := new_value4, key3 => new_value5},
#{key1 := V1, key2 := V2, key1 := V3} = OtherMap.
```

### Should be on by default?

**YES**

### Options

-  `preferred_style :: left | right` (default: `right`)

(I picked `right` as the default because it's my favorite, but if there is a good reason for it, `left` can also be chosen as the default. Also, the interpretation is open. In my mind `left` means `Var = #{pattern := _}` and `right` means `#{pattern := _} = Var`, but it's up to the implementor to document this correctly.

### Reasoning

> When capturing a parameter using pattern matching you can either put the parameter name before or after the value i.e.
> 
> ```erlang
> parse(#{ok := Values} = Pair) ->
> …
> ```
> 
> …or…
> 
> ```erlang
> parse(Pair = #{ok := Values}) ->
> …
> ```
> 
> Neither of these is better than the other, but it seems a good idea not to mix the two patterns in the same codebase.
> While this is not necessarily a concern for the correctness of your code, you should use a consistent style throughout your codebase.

### Origin (#281)

Inspired by the `Consistency.ParameterPatternMatching` rule from [Credo](https://hexdocs.pm/credo/Credo.Check.Consistency.ParameterPatternMatching.html)

elbrujohalcon:


doc_rules/elvis_style/param_pattern_matching.md
```diff
@@ -0,0 +1,21 @@
+# Param Pattern-Matching
+
+(since [3.0.0](https://github.com/inaka/elvis_core/releases/tag/3.0.0))
+
+When capturing a parameter using pattern matching you can either put the parameter
+name on the left (`Param = #{pattern := ToMatch}`) or right (`#{pattern := ToMatch} = Param`) side
+of the pattern that you use in the function clause.
+This rule will make sure you are consistent through your code and use always the same style.
```
paulo-ferraz-oliveira, line 8:
```suggestion
This rule will make sure you are consistent throughout your code and always use the same style.
```

tak30 (approved):

src/elvis_code.erl
```diff
@@ -76,7 +76,7 @@ content_zipper(Root) ->
 -spec all_zipper(ktn_code:tree_node()) -> zipper:zipper(_).
 all_zipper(Root) ->
     IsBranch =
-        fun(Node = #{}) -> ktn_code:content(Node) =/= [] orelse maps:is_key(node_attrs, Node) end,
+        fun(#{} = Node) -> ktn_code:content(Node) =/= [] orelse maps:is_key(node_attrs, Node) end,
```
paulo-ferraz-oliveira, line 79: Eating our dog food. I like it!

src/elvis_style.erl
```diff
@@ -121,9 +122,12 @@
 -define(NUMERIC_FORMAT_MSG,
         "Number ~p on line ~p does not respect the format "
         "defined by the regular expression '~p'.").
--define(BEHAVIOUR_SPELLING,
+-define(BEHAVIOUR_SPELLING_MSG,
         "The behavior/behaviour in line ~p is misspelt, please use the "
         "~p spelling.").
+-define(PARAM_PATTERN_MATCHING_MSG,
+        "Variable ~ts, used to match a parameter in line ~p is placed on "
```
paulo-ferraz-oliveira, line 129:
```suggestion
        "Variable ~ts, used to match a parameter in line ~p, is placed on "
```

src/elvis_style.erl
```diff
@@ -1125,7 +1131,12 @@ numeric_format(Config, Target, RuleConfig) ->
                          IntNodes,
                          check_numeric_format(FloatRegex, FloatNodes, [])).
 
--spec behaviour_spelling(elvis_config:config(), elvis_file:file(), empty_rule_config()) ->
+-type behaviour_spelling_config() ::
+    #{ignore => [ignorable()], spelling => behaviour | behavior}.
+
+-spec behaviour_spelling(elvis_config:config(),
+                         elvis_file:file(),
+                         behaviour_spelling_config()) ->
```
paulo-ferraz-oliveira, line 1139: Did this change just creep into the PR? Or is it expected?

elbrujohalcon: Yeah… It was a wrong spec that I silently fixed… 🤭

src/elvis_style.erl
```diff
@@ -1143,11 +1154,64 @@ behaviour_spelling(Config, Target, RuleConfig) ->
...
+
+is_function_clause(Zipper) ->
+    clause
+    == ktn_code:type(
+           zipper:node(Zipper))
+    andalso lists:member(
+                ktn_code:type(
+                    zipper:node(
+                        zipper:up(Zipper))),
+                [function, 'fun']).
```
paulo-ferraz-oliveira, line 1213: Is this the formatter? Maybe we can re-write it so that even upon formatting it looks more readable (?)

elbrujohalcon: It's the formatter … and I ❤️ it in this exquisite staircase style 😍

elbrujohalcon: Avoiding the issue requires 2 or 3 nested case statements… or several functions… It's always somewhat nasty, I think.

paulo-ferraz-oliveira: **Aside**: not sure anybody's tried to fight for pipe in Erlang, though, but this would be a case where it'd fit :) 

```
Zipper
|> zipper:up()
|> zipper:node()
|> ktn_code:type()
|> lists:member([function, 'fun'])
```

elbrujohalcon: [cb10984](https://github.com/inaka/elvis_core/pull/305/commits/cb10984fe08202868b57a6016d1447c72680cf7a)

paulo-ferraz-oliveira: Would something like the following be better?

```erlang
is_function_clause(Zipper) ->
    Funs = [function, 'fun'],
    is_clause(Zipper) andalso type_in(ktn_type_parent(Zipper), Funs).

ktn_type(Item) ->
  ktn_code:type(Item).

ktn_type_parent(Zipper) ->
  ktn_type(zipper:node(zipper:up(Zipper))).

type_in(Elem, In) ->
  lists:member(Elem, In)

is_clause(Zipper) ->
  ktn_type(Zipper) == clause.
```

(untested code)

elbrujohalcon: Too many functions for me… and why would you wrap `ktn_code:type/1` in `ktn_type/1`?

I don't like functions that just call other functions with the same arguments.

I updated the code in the commit I linked above. I think that's good enough.

paulo-ferraz-oliveira (approved):

PR #306, opened 2023-03-02
Make support visible in README

paulo-ferraz-oliveira:
There's a minor fix to another issue that I had done locally.


paulo-ferraz-oliveira:
@elbrujohalcon, shall we merge this now or only when 3.0 comes out?

PR #314, opened 2023-07-01
Feature a new katana

paulo-ferraz-oliveira:
This:

* removes dependencies on `maybe_expr` unless where necessary (i.e. running for OTP 25+)
* tweaks `rebar.config` a bit
* fixes a minor `spec`
* ~⚠️ is yet dependant on non-canonical, not tagged/released stuff, so requires an update before an approval+merge~

~CI's failing on [Run elvis on elvis](https://github.com/inaka/elvis_core/actions/runs/5428593917/jobs/9872926487#step:9:1), but that's expected, as per the above, and we have proper references/tags for dependencies and plugins.~

## On the use of `strong validation`

* (from [Erlang -- compile](https://www.erlang.org/doc/man/compile.html)): `strong_validation` Similar to option `basic_validation`. No code is generated...


elbrujohalcon (approved):

elbrujohalcon (approved):

paulo-ferraz-oliveira:
If the checks pass here, with your approval, we should tag/release, so you can import next to `rebar3_lint`.

elbrujohalcon (approved):

elbrujohalcon:
Done: https://hex.pm/packages/elvis_core/3.0.1

PR #326, opened 2023-11-03
Add wildcard support in no_call_functions

### LINKED ISSUE #325: Wildcards in no_call_functions MFA

Wildcards in no_call_functions MFA, i.e. rules that applies regardless of module name and/or function name and/or arity.

Being able to do:

```
#{dirs => ["."],
       filter => "*tests.erl",
       rules =>
           [{elvis_style, no_call,
             #{no_call_functions => [{'*', handle_event, 4}]]
      }
```

Dynamically generating rule definitions... too much work.

## Additional Context

It is ridiculous to test a state machine without a running process, and with state data that it has not produced itself.
Running the statem/server callbacks explicitly from tests is the biggest antipattern i know in testing.

attah:
Fixes  #325

NB: I made an uneducated guess what the next version will be called (and that this will make it in until then).

### DIFF
--- a/doc_rules/elvis_style/no_call.md
+++ b/doc_rules/elvis_style/no_call.md
@@ -15,8 +15,11 @@ functions).
 
 ## Options
 
-- `no_call_functions :: [{module(), function(), arity()} | {module(), function()}]`.
+- `no_call_functions :: [{module() | '_', function() | '_', arity() | '_'} |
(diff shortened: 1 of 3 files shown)


attah:
Whoops; didn't realize newline in bullet lists would render properly, but seems to work fine.

PR #332, opened 2024-01-17
Allow errorless consumption by `rebar3_lint`

paulo-ferraz-oliveira:
This will be required once we want to add more context to errors in the plugin. I'll prepare a pull request for that and only propose it for acceptance once this is merged, tagged and Hex-published, at least.

### DIFF
--- a/src/elvis_core.erl
+++ b/src/elvis_core.erl
@@ -38,7 +38,8 @@ start() ->
 
 %%% Rock Command
 
--spec rock(elvis_config:configs()) -> ok | {fail, [elvis_result:file()]}.
+-spec rock(elvis_config:configs()) ->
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
For context, the change comes from 

```erlang
-spec do_parallel_rock(elvis_config:config()) ->
                          ok |
                          {fail, [{throw, term()} | elvis_result:file() | elvis_result:rule()]}.
```

being used as

```erlang
    Results = lists:map(fun do_parallel_rock/1, NewConfig),
    lists:foldl(fun combine_results/2, ok, Results).
```

in `rock/1`, the updated function.

elbrujohalcon (approved):

paulo-ferraz-oliveira:
Going to release 3.2.1 on GitHub and Hex.

PR #341, opened 2024-05-21
Test over OTP 27 and rebar3 3.23

paulo-ferraz-oliveira:
## Notes on pull request

Found issues are around the development and maintenance of this library and `rebar3_format`. They do not seem to indicate `elvis_core` as non- OTP-27-ready.

### DIFF
--- a/test/style_SUITE.erl
+++ b/test/style_SUITE.erl
@@ -1238,6 +1238,8 @@ verify_no_nested_try_catch(Config) ->
                               Path).
 
 -spec verify_no_successive_maps(config()) -> any().
+-if(?OTP_RELEASE < 27).
+
 verify_no_successive_maps(Config) ->
     Group = proplists:get_value(group, Config, erl_files),
     Ext = proplists:get_value(test_file_ext, Config, "erl"),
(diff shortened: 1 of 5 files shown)


paulo-ferraz-oliveira:
Since this is now forbidden in OTP 27

```
===> Compiling elvis_core
compile: warnings being treated as errors
compile: warnings being treated as errors
/home/runner/work/elvis_core/elvis_core/test/examples/fail_no_successive_maps.erl:6:22: expression updates a literal
%    6|     M = #{this => is}#{wrong => "and"},
%     |                      ^

/home/runner/work/elvis_core/elvis_core/test/examples/pass_no_successive_maps_elvis_attr.erl:8:22: expression updates a literal
%    8|     M = #{this => is}#{wrong => "and"},
%     |                      ^
```

we might have to tweak the tests (macros 🎉) and think about dropping this check in the future 😄

paulo-ferraz-oliveira:
`rebar3 format --verify` fails for OTP 24, but if I format the files with OTP 24 it then fails for OTP 27 😕

test/style_SUITE.erl
```diff
@@ -1260,6 +1262,13 @@ verify_no_successive_maps(Config) ->
                               #{ignore => [Module]},
                               Path).
 
+-else.
+
+verify_no_successive_maps(_Config) ->
+    [].
```
paulo-ferraz-oliveira, line 1268: We basically avoid the test after having made sure compilation is Ok for all tested OTP versions.

elbrujohalcon (approved):

PR #354, opened 2024-08-16
Add opaque state as a type declaration

### LINKED ISSUE #349: `state_record_and_type` not considering `opaque` as a type declaration
## Bug Description

I have a state, in a `gen_server` declared as a record.

I have to export it for `export_used_types` (it's in the public interface).

If I try to make it opaque, I get `This module implements an OTP behavior and has a 'state' record but is missing a 'state()' type.`

Should it be "wrong" (for Elvis) to have

```erlang
-record(state, {
}).
-opaque state() :: #state{}.
-export_type([state/0]).
```

?

## To Reproduce

Check Description.

## Expected Behavior

`opaque` should be handled the same as `type`.

## `rebar3` Logs

Not relevant.

## Additional Context

Not relevant.

belltoy:
Fix #349

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1834,7 +1834,20 @@ has_state_record(Root) ->
 -spec has_state_type(ktn_code:tree_node()) -> boolean().
 has_state_type(Root) ->
     IsStateType =
-        fun(Node) -> (type_attr == ktn_code:type(Node)) and (state == ktn_code:attr(name, Node))
+        fun(Node) ->
+           case ktn_code:type(Node) of
+               type_attr ->
+                   state == ktn_code:attr(name, Node);
+               opaque ->
+                   case ktn_code:attr(value, Node) of
+                       {state, _, _} ->
+                           true;
+                       _ ->
+                           false
+                   end;
+               _ ->
+                   false
+           end
         end,
     elvis_code:find(IsStateType, Root) /= [].
 
--- a/test/examples/fail_state_record_and_type_plus_export_used_types.erl
+++ b/test/examples/fail_state_record_and_type_plus_export_used_types.erl
@@ -1,0 +1,42 @@
+-module(fail_state_record_and_type_plus_export_used_types).
+
+-dialyzer(no_behaviours).
+
+-behaviour(gen_server).
+
+-export([
+         init/1,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         code_change/3,
+         terminate/2
+        ]).
+
+-record(state, {}).
+
+-type state() :: #state{}.
+
+-spec init(term()) -> state().
+init(_Args) ->
+    #state{}.
+
+-spec handle_call(term(), term(), term()) -> ok.
+handle_call(_Request, _From, _State) ->
+    ok.
+
+-spec handle_cast(term(), term()) -> ok.
+handle_cast(_Request, _State) ->
+    ok.
+
+-spec handle_info(term(), term()) -> ok.
+handle_info(_Info, _State) ->
+    ok.
+
+-spec code_change(term(), term(), term()) -> term().
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+-spec terminate(term(), term()) -> ok.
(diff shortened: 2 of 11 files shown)


src/elvis_style.erl
```diff
@@ -1834,7 +1834,9 @@ has_state_record(Root) ->
 -spec has_state_type(ktn_code:tree_node()) -> boolean().
 has_state_type(Root) ->
     IsStateType =
-        fun(Node) -> (type_attr == ktn_code:type(Node)) and (state == ktn_code:attr(name, Node))
+        fun(Node) ->
+           TypeOrOpaque = (type_attr == ktn_code:type(Node)) or (opaque == ktn_code:type(Node)),
+           TypeOrOpaque and (state == ktn_code:attr(name, Node))
```
elbrujohalcon, line 1839:
```suggestion
        fun(Node) ->
          ktn_code:attr(name, Node) == state andalso lists:member(ktn_code:type(Node), [type_attr, opaque])
```

elbrujohalcon (changes requested):
Can you add a test case for this and, if needed, update the corresponding docs?
Thanks.

belltoy:
The first implementation was wrong. I've rebased it with some test cases. And also provided cases to verify rule `state_record_and_type` + rule `export_used_types`.

But something looks weird to me. In the [`pass_state_record_and_type.erl`](https://github.com/inaka/elvis_core/blob/main/test/examples/pass_state_record_and_type.erl) case, the type `state()` is defined without exported, but the `state()` is used in the exported function's spec.

test/examples/fail_state_record_and_type_plus_export_used_types.erl
```diff
@@ -0,0 +1,42 @@
...
+
+-type state() :: #state{}.
+
+-spec init(term()) -> state().
+init(_Args) ->
+    #state{}.
+
+-spec handle_call(term(), term(), term()) -> ok.
+handle_call(_Request, _From, _State) ->
+    ok.
```
elbrujohalcon, line 26:
```suggestion
-spec handle_call(term(), term(), state()) -> {noreply, State}.
handle_call(_Request, _From, State) ->
    {noreply, State}.
```

elbrujohalcon: And the same goes for all other `gen_server` implementations in this PR. Let's make them as _realistic_ as possible.

test/examples/fail_state_record_and_type_plus_export_used_types.erl
```diff
@@ -0,0 +1,42 @@
...
+handle_info(_Info, _State) ->
+    ok.
+
+-spec code_change(term(), term(), term()) -> term().
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+-spec terminate(term(), term()) -> ok.
+terminate(_Reason, _State) ->
+    ok.
```
elbrujohalcon, line 42: You can remote these 3 callbacks, since they are optional and they don't contribute to the validation of the rules we want to enforce here.
The same applies for all the other `gen_server` implementations in this PR.

elbrujohalcon: I noticed you copied them from `pass_state_record_and_type.erl`, let's take this chance to also simplify that one and make it more _realistic_ 🙏🏻

elbrujohalcon:
> But something looks weird to me. In the [`pass_state_record_and_type.erl`](https://github.com/inaka/elvis_core/blob/main/test/examples/pass_state_record_and_type.erl) case, the type `state()` is defined without exported, but the `state()` is used in the exported function's spec.

I'm not 100% sure, but it's likely because we're not running the `used_exported_types` rule on it.

elbrujohalcon (approved):
I approve the PR, in general.
So, if you don't have time or just don't feel like improving the `gen_server`s, let me know and I'll merge it.
OTOH, if you can improve that code (even when it's _just for tests_), that would be lovely.

belltoy:
Of course, I will improve those codes in test cases.

Issue #297, opened 2023-03-01
New rule: Prefer Unquoted Atoms


paulo-ferraz-oliveira:
### Name

`prefer_unquoted_atoms`

### Brief Description

Prefer unquoted atoms unless quotes are necessary.

### Reasoning

This one is not a technical concern, but a readability concern. The code will behave identical in both ways.

### Refactoring Proposal

`elvis` shall warn that your atoms are quoted when they don't have to be.

**Note**: if you're using a formatter this could/should already be taken into account by it.

### Origin 

Inspired by Credo's https://hexdocs.pm/credo/Credo.Check.Readability.PreferUnquotedAtoms.html.

paulo-ferraz-oliveira:
One thing I remembered is that language constructs need to be quoted as atoms. e.g. `'maybe'` (quoted) is not the same as `maybe` (unquoted). That'll have to be taken into account when implementing.

elbrujohalcon:
Maybe other formatters would leave atoms as-they-found-them, but at least rebar3_format (with the default formatter) will apply this concept appropriately (i.e., only atoms that need to be quoted will be quoted).

elbrujohalcon:
Not saying "no" to the rule, just that it may be low priority since other tools cover this aspect.

bormilan:
I had a long session discovering how to implement this, and I found a problem. I made a test file that contains quoted atoms that are not required to be quoted. I dug deep into the code of `elvis` and `ktn_code`, and I found out that quotes disappear if they are not needed.

Example file:
![image](https://github.com/user-attachments/assets/542faf17-0160-44fd-8cd2-9f582140aae9)

And this is the output of the file read by the function: `elvis_file:src`
![image](https://github.com/user-attachments/assets/13647f6a-55b9-4b91-b746-5d59f2ee7236)

paulo-ferraz-oliveira:
If you're analysing from BEAM this might not apply; if you're analysing from source it's still interesting.

bormilan:
I spent much time exploring how these work in the repo, but I'm still confused. I see that we have several rules that require .erl files not .beam for example macro naming conventions. But I also see that in the `style_SUITE.erl,` we only have a group called `beam_files` which contains only test that uses .beam files. I'm not sure why other test cases exist but have never been used. Or do I miss something important? I wanted to see how other rules read .erl files, bc this issue only can be solved in that way, but I'm struggling to find one.

bormilan:
[...]

PR #355, opened 2024-09-03
feat(#297): New rule: Prefer Unquoted Atoms

bormilan:
This is in an early stage, I have som things to fix.


src/elvis_style.erl
```diff
@@ -196,6 +200,8 @@ default(no_common_caveats_call) ->
            {erlang, size, 1}]};
 default(atom_naming_convention) ->
     #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$", enclosed_atoms => ".*"};
+default(prefer_unquoted_atoms) ->
+    #{regex => "^'([a-z_0-9)]+)'$", enclosed_atoms => ".*"};
```
elbrujohalcon, line 204: I would not add a `regex` parameter, nor an `enclosed_atoms` one. Those are fine for `atom_naming_convention` since they allow users to decide what atoms to forbid. In this case, the intention of the rule doesn't really allow configuration at all.

src/elvis_style.erl
```diff
@@ -1015,6 +1021,19 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+                            elvis_file:file(),
+                            empty_rule_config()) ->
+                               [elvis_result:item()].
+prefer_unquoted_atoms(_Config, Target, RuleConfig) ->
+    Regex = option(regex, RuleConfig, prefer_unquoted_atoms),
+    RegexEnclosed =
+        specific_or_default(option(enclosed_atoms, RuleConfig, atom_naming_convention), Regex),
+    {Content, #{encoding := _Encoding}} = elvis_file:src(Target),
+    Tree = ktn_code:parse_tree(Content),
+    AtomNodes = elvis_code:find(fun is_atom_node/1, Tree, #{traverse => all, mode => node}),
```
elbrujohalcon, line 1034: Same as above… I would avoid configuration options entirely.

src/elvis_style.erl
```diff
@@ -1456,6 +1475,34 @@ check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) -
...
+                AccIn;
+            {match, _Captured} ->
+                Msg = ?ATOM_PREFERRED_QUOTES_MSG,
+                {Line, _} = ktn_code:attr(location, AtomNode),
+                Info = [AtomName0, Line, RegexEnclosed],
+                Result = elvis_result:new(item, Msg, Info, Line),
+                AccIn ++ [Result]
+        end,
+    check_atom_quotes(Regex, RegexEnclosed, RemainingAtomNodes, AccOut).
+
```
elbrujohalcon, line 1505: Basically… the same as above, again… In my mind the check here should be something like…
```erlang
    AtomName0 = ktn_code:attr(text, AtomNode),
    {IsEnclosed, AtomName} = string_strip_enclosed(AtomName0),
    IsExceptionClass = is_exception_or_non_reversible(ValueAtomName),
    case AtomName of
        [$' | _] when not IsExceptionClass andalso not IsEnclosed ->
%% or… <<$', _/binary>> when not IsExceptionClass andalso not IsEnclosed ->
%% if it's a binary (I don't remember the data type here… sorry)
            Msg = ?ATOM_PREFERRED_QUOTES_MSG,
            {Line, _} = ktn_code:attr(location, AtomNode),
            Info = [AtomName0, Line, RegexEnclosed],
                Result = elvis_result:new(item, Msg, Info, Line),
                AccIn ++ [Result];
        _ ->
            AccIn
    end,
    check_atom_quotes(…).
```

elbrujohalcon (changes requested):
The code is fine, I would just simplify it by removing the options.

src/elvis_style.erl
bormilan: I see, I was not sure about them. Thanks for the explanation.

src/elvis_style.erl
bormilan: Yes, it's simpler than that, I was thinking about refactoring this, it was a quick solution copying from the `atom_naming_convention`. Thanks!

src/elvis_style.erl
```diff
@@ -12,7 +12,8 @@
          atom_naming_convention/3, no_throw/3, no_dollar_space/3, no_author/3, no_import/3,
          no_catch_expressions/3, no_single_clause_case/3, numeric_format/3, behaviour_spelling/3,
          always_shortcircuit/3, consistent_generic_type/3, export_used_types/3,
-         no_match_in_condition/3, param_pattern_matching/3, private_data_types/3, option/3]).
+         no_match_in_condition/3, param_pattern_matching/3, private_data_types/3, option/3
+         , prefer_unquoted_atoms/3]).
```
paulo-ferraz-oliveira, line 16: Before submitting, prefer executing `rebar3 test` locally, to have the formatter, linter, etc. do their work 😄

paulo-ferraz-oliveira: (we probably need to update the contributing guide to include this)

src/elvis_style.erl
```diff
@@ -107,6 +108,9 @@
 -define(ATOM_NAMING_CONVENTION_MSG,
         "Atom ~p on line ~p does not respect the format "
         "defined by the regular expression '~p'.").
+-define(ATOM_PREFERRED_QUOTES_MSG,
+        "Atom ~p on line ~p has quotes on them "
+        "but its not needed on it.").
```
paulo-ferraz-oliveira, line 113:
```suggestion
        "Atom ~p on line ~p is quoted "
        "but quotes are not needed.").
```

test/examples/fail_quoted_atoms.erl
paulo-ferraz-oliveira, line 1: One of the tests should at least contain a few reserved words in quote.

Something like `'maybe'` and `'and'`, for example, for better implementation coverage.

elbrujohalcon:
Once you're done with the code, @bormilan … you also need to add docs to this rule (in RULES.md) and you need to include it in the appropriate ruleset(s) (I guess `erl_files` and `hrl_files`)

bormilan:
Thanks for all the help and suggestions, currently I'm on a little vacation, but after that I will fix all of them and update the PR.

test/examples/fail_quoted_atoms.erl
bormilan: Do you mean to make exceptions for `'maybe'` and `'and'`? I guess we don't want to use them as `and` because they are keywords.

src/elvis_style.erl
bormilan: I'd be happy to do it if you want.

src/elvis_style.erl
bormilan: I fixed this, but `rebar3 test` fails on this because it says: 'The following pieces of code are dead and should be removed"
I can't delete them entirely bc all rules function needs to be with arity 3.
Could you help me with this, what do I miss?

test/examples/fail_quoted_atoms.erl
bormilan: Oh, I'm so dumb, You can't use a keyword like that, sorry. I never tried so I had not known this.

bormilan:
Today I made some changes, but I had some questions that I commented on your suggestions above.
Besides these, I have made two hardcoded variables that I want to put somewhere else(bc they are constant values), but I don't know where to. If you could help me with that I would appreciate it. These are `Regex` and `KeyWords`.

doc_rules/elvis_style/prefer_unquoted_atoms.md
```diff
@@ -0,0 +1,15 @@
+# Prefer unquoted atoms
+
+Do not use quotes on atoms that are not need it.
```
elbrujohalcon, line 3:
```suggestion
Do not use quotes on atoms that don't need to be quoted.
```

src/elvis_style.erl
```diff
@@ -1456,6 +1470,41 @@ check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) -
...
+    ValueAtomName = ktn_code:attr(value, AtomNode),
+    {IsEnclosed, AtomName} = string_strip_enclosed(AtomName0),
+
+    IsException = is_exception_prefer_quoted(ValueAtomName),
+    RE = re_compile_for_atom_type(IsEnclosed, Regex, RegexEnclosed),
+    AccOut =
+        case re:run(
+                 unicode:characters_to_list(AtomName, unicode), RE)
+        of
+            {match, _Captured} when not IsException ->
```
elbrujohalcon, line 1489:
```suggestion
        case unicode:characters_to_list(AtomName, unicode) of
            [$' | _] when not IsException ->
```

We don't need a regex for this, we can just use pattern-matching.

src/elvis_style.erl
```diff
@@ -1456,6 +1470,41 @@ check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) -
...
+                Msg = ?ATOM_PREFERRED_QUOTES_MSG,
+                {Line, _} = ktn_code:attr(location, AtomNode),
+                Info = [AtomName0, Line, RegexEnclosed],
```
elbrujohalcon, line 1492:
```suggestion
                Info = [AtomName0, Line],
```
The macro has only 2 parameters.

test/examples/pass_unquoted_atoms.erl
```diff
@@ -0,0 +1,9 @@
+-module(pass_unquoted_atoms).
+
+-export([test/1, test/2]).
+
+test(_Test) -> ok.
+
+test(_A, nice_atom_name) -> perfect_atomname;
+test(_A, _B) -> unquoted_atom.
```
elbrujohalcon, line 8:
```suggestion
test(_Reserved, _Words) -> ['after', 'and', 'andalso', 'band', 'begin', 'bnot', 'bor', 'bsl', 'bsr', 'bxor', 'case',
         'catch', 'cond', 'div', 'end', 'fun', 'if', 'let', 'not', 'of', 'or', 'orelse', 'receive',
         'rem', 'try', 'when', 'xor'];
test(_A, _B) -> unquoted_atom.
```

Can we add some reserved words here? Maybe even all of them?

elbrujohalcon (changes requested):
A few improvements and… I wonder what happens with `'maybe'` as an atom.
I checked erlang/otp's code and wherever there is a list of keywords/reserved words… they do **not** include `maybe`, but I guess… if the feature is enabled (and I think it's enabled by default on OTP27+), that one should also be considered _a keyword_, right?

src/elvis_style.erl
elbrujohalcon: I'm not sure if that would be _the definitive solution™️ _, but, considering that you're dealing with the text of the file (and not it's AST), it's possible that this rule actually belongs to `elvis_text_style`, instead of `elvis_style`. Hopefully, that will allow you to write more concise functions without the extra arguments.

doc_rules/elvis_style/prefer_unquoted_atoms.md
bormilan: Sorry for these mistakes.

src/elvis_style.erl
bormilan: Ooh, I see. I did not understand this line first, thank you for the clarification. I like this clean solution, then I can completely get rid of the regex.

elbrujohalcon (changes requested):
The code is perfectly fine, @bormilan … but I still believe this rule belongs to `elvis_text_style` and not `elvis_style`.

bormilan:
So, should we add "maybe" to the exceptions? I think we should. But because now it's not a keyword by default, we may wait and add it when erl/OTP27 comes.

elbrujohalcon:
Personally… I would add it right now… It will be a reserved word soon enough.
I wish Erlang/OTP would have a public function that returned the list of reserved words somewhere (`erl_syntax` may be a great place)… 🤷🏻

RULES.md
```diff
@@ -61,6 +61,7 @@ identified with `(since ...)` for convenience purposes.
 - [State Record and Type](doc_rules/elvis_style/state_record_and_type.md)
 - [Used Ignored Variable](doc_rules/elvis_style/used_ignored_variable.md)
 - [Variable Naming Convention](doc_rules/elvis_style/variable_naming_convention.md)
+- [Prefer Unquoted Atoms](doc_rules/elvis_style/prefer_unquoted_atoms.md)
```
elbrujohalcon, line 64:
```suggestion
- [Prefer Unquoted Atoms](doc_rules/elvis_text_style/prefer_unquoted_atoms.md)
```

doc_rules/elvis_style/prefer_unquoted_atoms.md
```diff
@@ -0,0 +1,15 @@
+# Prefer unquoted atoms
```
elbrujohalcon, line 1: This file should be moved to `elvis_text_style` folder.

elbrujohalcon (changes requested):

RULES.md
bormilan: I'm sorry that I forgot this. I will fix it.

elbrujohalcon (approved):

elbrujohalcon:
Excellent work, @bormilan !!! Thank you very much!

bormilan:
Thank you so much for all the help and patience. It was a great start to my journey.

bormilan:
I found a little problem, the formatter changes the `'maybe'` to `maybe` in the list of exceptions. Should I turn it off to that file, or we should consider changing this in the formatter itself to make an exception for it?

elbrujohalcon:
@bormilan The list of exceptions should have strings and not atoms… that would fix it :)

elbrujohalcon:
But… the formatter is broken, anyway… please report an issue there.

elbrujohalcon (approved):

bormilan:
> Personally… I would add it right now… It will be a reserved word soon enough. I wish Erlang/OTP would have a public function that returned the list of reserved words somewhere (`erl_syntax` may be a great place)… 🤷🏻

I think `f_reserved_word` does exactly what you want in the module [erl_scan](https://www.erlang.org/docs/17/man/erl_scan)

elbrujohalcon:
> > Personally… I would add it right now… It will be a reserved word soon enough. I wish Erlang/OTP would have a public function that returned the list of reserved words somewhere (`erl_syntax` may be a great place)… 🤷🏻
> 
> I think `f_reserved_word` does exactly what you want in the module [erl_scan](https://www.erlang.org/docs/17/man/erl_scan)

Yeah, but it's not exported, is it?

bormilan:
It is:
image

bormilan:
> > > Personally… I would add it right now… It will be a reserved word soon enough. I wish Erlang/OTP would have a public function that returned the list of reserved words somewhere (`erl_syntax` may be a great place)… 🤷🏻
> > 
> > 
> > I think `f_reserved_word` does exactly what you want in the module [erl_scan](https://www.erlang.org/docs/17/man/erl_scan)
> 
> Yeah, but it's not exported, is it?

It is:
image

elbrujohalcon:
> > > > Personally… I would add it right now… It will be a reserved word soon enough. I wish Erlang/OTP would have a public function that returned the list of reserved words somewhere (`erl_syntax` may be a great place)… 🤷🏻
> > > 
> > > 
> > > I think `f_reserved_word` does exactly what you want in the module [erl_scan](https://www.erlang.org/docs/17/man/erl_scan)
> > 
> > 
> > Yeah, but it's not exported, is it?
> 
> It is: image

Ah, right! I was thinking of something like `f_reserved_words/0` (that would return the list) :)

bormilan:
> > > > > Personally… I would add it right now… It will be a reserved word soon enough. I wish Erlang/OTP would have a public function that returned the list of reserved words somewhere (`erl_syntax` may be a great place)… 🤷🏻
> > > > 
> > > > 
> > > > I think `f_reserved_word` does exactly what you want in the module [erl_scan](https://www.erlang.org/docs/17/man/erl_scan)
> > > 
> > > 
> > > Yeah, but it's not exported, is it?
> > 
> > 
> > It is: image
> 
> Ah, right! I was thinking of something like `f_reserved_words/0` (that would return the list) :)
> 
> But… maybe we can actually use that here or in `rebar3_format`… 🤔 If you want to send a PR using `erl_scan:f_reserved_word/1` to verify that the atom is actually a reserved word instead of the hardcoded list of strings we now have… I'd merge it, for sure!

Ok, I will do it shortly!

PR #356, opened 2024-09-11
Fix/atom naming convention regex

bormilan:


test/examples/pass_atom_naming_convention.erl
```diff
@@ -5,4 +5,5 @@
 for_test() ->
     this_is_an_ok_atom,
     'and_so_is_this',
-    'and_this_1_too'.
+    'and_this_1_too',
+    non_200.
```
elbrujohalcon, line 9: Can we add these atoms to the test, please?

```erlang
'_', % used by ets/mnesia/etc.
non200, % valid, even without underscores
valid_200even_if_numb3rs_appear_between_letters,
```

test/examples/fail_atom_naming_convention.erl
```diff
@@ -12,4 +12,5 @@ for_test() ->
     '\'\'',
     '\'startswithbacktick',
     'backtick\'inside',
-    'backtick at the end\''.
+    'backtick at the end\'',
+    non200_.
```
elbrujohalcon, line 16: Can we add these atoms to the test, please?

```erlang
'__', % invalid, even when '_' is actually valid
'_something', % invalid because it starts with underscore
'42_invalid_because_it_starts_with_a_number',
'42invalid', %% even without underscores
weDontSupportCamelCaseHere.

**Review by elbrujohalcon — CHANGES_REQUESTED** [review]
Let's add some more atoms to the test to verify our regex even further.

File: test/examples/pass_atom_naming_convention.erl [hunk 3]
```diff
@@ -5,4 +5,8 @@
 for_test() ->
     this_is_an_ok_atom,
     'and_so_is_this',
-    'and_this_1_too'.
+    'and_this_1_too',
```
elbrujohalcon, line 8: Sorry, @bormilan, but I thought of another atom that should be valid but may be marked as invalid by the new regex: `x` (i.e., a single ASCII character). Can you add it here, please?

And also, can you add `blahblah_SUITE`.

src/elvis_style.erl
```diff
@@ -195,7 +195,7 @@ default(no_common_caveats_call) ->
            {timer, send_interval, 3},
            {erlang, size, 1}]};
 default(atom_naming_convention) ->
-    #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$", enclosed_atoms => ".*"};
+    #{regex => "^[a-z]+((_?[a-z0-9])*[a-z0-9]+)*(_SUITE)?$", enclosed_atoms => ".*"};
```
elbrujohalcon, line 198: I had not tested this one, but I think this better encapsulates the sprit of what we want (if it works)…
```suggestion
    #{regex => "^[a-z]+(_[a-z0-9]+|[a-z0-9]+)*(_SUITE)?$", enclosed_atoms => ".*"};
```
In my mind, this reads as…
* you start with a letter (`[a-z]`)
* then you can follow it with as many instances as you want of either…
    - an underscore and a word (`_[a-z0-9]+`); or…
    - just a word (`[a-z0-9]+`)
* then it can have a `_SUITE` at the end.

test/examples/fail_atom_naming_convention.erl
```diff
@@ -12,4 +12,11 @@ for_test() ->
     '\'\'',
     '\'startswithbacktick',
     'backtick\'inside',
-    'backtick at the end\''.
+    'backtick at the end\'',
+    non200_,
+    '__', % invalid, even when '_' is actually valid
+    two_underscores__together_are_not_valid,
```
elbrujohalcon, line 18: Can you add `not_even_in_a__SUITE`, please?

elbrujohalcon (changes requested):
Sorry, @bormilan … I think we still need another iteration on this.

paulo-ferraz-oliveira:
It'd be interesting to understand if this is interface-breaking, and signal that as a major in the release, in case it is, because we're changing an existing rule.

src/elvis_style.erl
paulo-ferraz-oliveira: I guess the first `+` is unnecessary, but it's a detail. However, would this work, too?

```
^[a-z]+(_[a-z0-9]+)*(_SUITE)?$
```

With it:
* you start with one or more letters (`[a-z]+`)
* then you can follow it with as many instances as you want of `_[a-z0-9]+`
* then it can have a `_SUITE` at the end.

bormilan: I think Brujo's one is better because atomname200 is also good but this does not allow it because it always expects an underscore.

paulo-ferraz-oliveira: ... which is why initially we had a `?` 😄 so... `^[a-z]+(_?[a-z0-9]+)*(_SUITE)?$`. Equivalent but less verbose?

elbrujohalcon: Then it can be even simpler: `^[a-z](_?[a-z0-9]+)*(_SUITE)?$`.
i.e.:
* The first character must be a smallcase letter.
* Then, the atom can have any number of "words" (built by letters and numbers) that may start with an underscore.
* It may end with `_SUITE`.

elbrujohalcon (approved):

elbrujohalcon:
I'll merge this as soon as CI is ✅ … @paulo-ferraz-oliveira I do believe that this deserves a major jump. So… 4.0.0 **is coming!!!**

elbrujohalcon:
To be clear: I would not release 4.0.0 yet… I would pick a few more tickets to tackle before it goes out. I'll make a release and associate a few issues with it.

elbrujohalcon:
There you go, folks: https://github.com/inaka/elvis_core/milestone/7

src/elvis_style.erl
paulo-ferraz-oliveira: Which is why I'd written "I guess the first `+` is unnecessary"...

PR #357, opened 2024-09-15
Add new ruleset, `gitignore`, with rule `gitignore_patterns`

### LINKED ISSUE #316: New ruleset `gitignore`
### Name

`gitignore`: similar in concept to `elvis_config` or `rebar_config`.

### Brief Description

Elvis should detect non-conformant content for an opinionated list of elements that should be present in the `.gitignore` file (and that it exists, too).

For the moment (hoping a discussion ensues and we come to a "fixed" initial list)

```
.rebar3
_build
_checkouts
doc/
erl_crash.dump
rebar3.crashdump
test/logs/
```

### Reasoning

We tend to forget one or the other, and this makes for easier analysis/consistency.

paulo-ferraz-oliveira:
It'll use files identified by `dirs`+`filter` (which should be `.gitignore`) and apply
(at the moment, the only existing) rule `gitignore_patterns` with the default value found in the documentation.

Closes #316.


RULES.md
```diff
@@ -83,6 +84,7 @@ The six pre-defined rulesets are:
 - `elvis_config`, for elvis configuration files.
 - `erl_files`, for Erlang source files (pre-defined rule set).
 - `erl_files_strict`, for Erlang source files (all available rules).
+- `gitignore`, for `.gitignore` files.
```
elbrujohalcon, line 87: Should it be `gitignores` (plural)… just to match the rest of the ruleset names? 🤔

elbrujohalcon (commented):
Can we make two rules, as follows?
* `required_patterns`: for stuff that must be present in the file (i.e., the current rule)
* `forbidden_patterns`: for stuff that must **not** be present in the file (including, for instance, `rebar.lock`)?

And maybe we can put them in their own module (`elvis_gitignore` or something like that), instead of `elvis_project`?

paulo-ferraz-oliveira:
> Can we make two rules, as follows?

Why not options?

RULES.md
paulo-ferraz-oliveira: `elvis_config` is not plural 😄, and there's others.

paulo-ferraz-oliveira:
> And maybe we can put them in their own module (elvis_gitignore or something like that)

`rebar_config` and `elvis_config` are part of "project", which is what this rule wants to be. Why should we separate into different files?

RULES.md
elbrujohalcon: Fair enogh. Let it be `gitignore`.

elbrujohalcon:
> > Can we make two rules, as follows?
> 
> Why not options?

Mostly to have a ruleset with 2 rules, instead of one. 😜

elbrujohalcon:
> > And maybe we can put them in their own module (elvis_gitignore or something like that)
> 
> `rebar_config` and `elvis_config` are part of "project", which is what this rule wants to be. Why should we separate into different files?

Fair enough. Let it be in `elvis_project`. :)

paulo-ferraz-oliveira:
Brujo I think you need to update the settings for this

image

paulo-ferraz-oliveira:
If it's not much, would you check the stuff above (initial description) that needs to be updated (?) - or maybe the checks are blocked to the author?

What's missing?

doc_rules/elvis_project/gitignore_patterns.md
```diff
@@ -0,0 +1,22 @@
+# `.gitignore` patterns
+
+(since [3.3.0](https://github.com/inaka/elvis_core/releases/tag/3.3.0))
+
+Include, in the project's `.gitignore` file, the patterns identified by the rule.
+
+## Options
+
+- `all_of :: [string()]`.
```
elbrujohalcon, line 9: If we're going to go with a single rule instead of 2… For the options I prefer the names `include` and `exclude`. Even more, `include_patterns` and `exclude_patterns`.

elbrujohalcon (changes requested):
I think I still prefer to have 2 rules: one for required patters and one for excluded patterns. In that case, I would call the option just `patterns` or (to match other rules) `regexes`.

doc_rules/elvis_project/gitignore_patterns.md
paulo-ferraz-oliveira: Yeah, `excludes` makes sense: I was initially just thinking about my use case. Would the default list be empty?

paulo-ferraz-oliveira:
We can go with two rules, sure. One's the mirror of the other one so duplicating code and tests + adjusting shouldn't take long.

paulo-ferraz-oliveira:
I've updated the description: I'm gonna make it two rules, each with the same option name. And I'm tentatively saying "since" is 4.0.0.

doc_rules/elvis_project/gitignore_patterns.md
elbrujohalcon: No, it should have `rebar.lock`. 😉

paulo-ferraz-oliveira: I'm assuming `rebar.lock` is really something you want, and the other one was a joke 😄.

elbrujohalcon: You're 100% right.

src/elvis_gitignore.erl
```diff
@@ -0,0 +1,107 @@
...
+                lists:any(fun(Line) -> re:run(Line, Pattern) =/= nomatch end, Lines);
+            forbidden ->
+                lists:all(fun(Line) -> re:run(Line, Pattern) =:= nomatch end, Lines)
+        end,
+    Results =
+        case ModeRespected of
+            true ->
+                Results0;
+            false when Mode =:= required ->
+                [elvis_result:new(item, ?REQUIRED_PATTERN, [Pattern]) | Results0];
```
paulo-ferraz-oliveira, line 81: Lineno is 0, here.

paulo-ferraz-oliveira: This should be Ok.

src/elvis_gitignore.erl
```diff
@@ -0,0 +1,107 @@
...
+            false when Mode =:= forbidden ->
+                [elvis_result:new(item, ?FORBIDDEN_PATTERN, [Pattern]) | Results0]
```
paulo-ferraz-oliveira, line 83: Lineno is 0, here.

To get it we might need to make the implementation a bit more complex.

test/dirs/apps/app1/.gitignore
```diff
@@ -0,0 +1,10 @@
+.rebar3/
+_build/
+
+_checkouts/
+extra
+doc/
+/erl_crash.dump
+/rebar3.crashdump
+test/logs/
+rebar0.lock
```
paulo-ferraz-oliveira, line 10: Not a typo, but a "test"

src/elvis_gitignore.erl
paulo-ferraz-oliveira, line 83: Lineno is 0, here.

paulo-ferraz-oliveira: To get a proper lineno we might need to make the code more complex.

test/dirs/apps/app1/.gitignore
elbrujohalcon: 👌🏻

src/elvis_gitignore.erl
elbrujohalcon: It's fine.

elbrujohalcon (approved):

elbrujohalcon:
It's approved, @paulo-ferraz-oliveira … feel free to merge it whenever the checks are ✅

paulo-ferraz-oliveira:
Good stuff. Thanks, @elbrujohalcon. Merging... (just gonna resolve the open conversations because of our rules to merge)

PR #364, opened 2024-09-23
Feature more `no_common_caveats_call` default options

### LINKED ISSUE #320: Propose `gen_server` et al `:call` "no-timeout" functions as `no_common_caveats_call`?
### Name

None yet.

### Brief Description

Should we propose some of the base functions (that have a timeout counterpart) to be added to the list of `no_common_caveats` functions? We could even call this "use-explicit-timeout".

There's a few considerations here:

* the list of functions would eventually have to be updated - and I prefer not to have to revisit rules, most of the time
* it's possible that the interface breaks in the future (it is already possible today, of course) and this is also extra-maintenance, because now the suggestion is OTP-version specific

This raised, though, another interesting thought, to me.

Should `elvis_core` consider rules to be grouped by `severity`, such as `info`, `warning`, `all`, etc.? There's already a proposal to categorize them by "type".

### Reasoning

Check this out: https://erlangforums.com/t/thoughts-on-the-default-timeout-for-gen-server-call/2812

### Refactoring Proposal

"Update your code."

paulo-ferraz-oliveira:
# Description

Closes #320.

Mind you this is only present in `erl_files_strict` so less far-reaching than it first it may seem.

### DIFF
--- a/doc_rules/elvis_style/no_common_caveats_call.md
+++ b/doc_rules/elvis_style/no_common_caveats_call.md
@@ -5,7 +5,8 @@
 The [Erlang Efficiency Guide](https://erlang.org/doc/efficiency_guide/commoncaveats.html) has a list
 of "Common Caveats" suggesting more efficient alternatives to several common functions.  This rule
 provides warnings if you call "inefficient" functions with entirely equivalent (efficient)
-alternatives.
+alternatives. It also warns you about functions with a more explicit interface (e.g. `gen_server:call/3`
+instead of `gen_server:call/2`) so you're sure to not have forgotten it.
 
 > Works on `.beam` file? Yes!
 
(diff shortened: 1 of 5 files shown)


doc_rules/elvis_style/no_common_caveats_call.md
```diff
@@ -5,7 +5,8 @@
 The [Erlang Efficiency Guide](https://erlang.org/doc/efficiency_guide/commoncaveats.html) has a list
 of "Common Caveats" suggesting more efficient alternatives to several common functions.  This rule
 provides warnings if you call "inefficient" functions with entirely equivalent (efficient)
-alternatives.
+alternatives. It also warns you about functions with a more explicit interface (e.g. `gen_server:call/3`
+instead of `gen_server:call/2`) so you're sure to not have forgotten it.
```
paulo-ferraz-oliveira, line 9: I was hesitant in either creating a new rule or adapting this one (as per the proposal on the linked issue). I went with the issue.

elbrujohalcon (approved):

paulo-ferraz-oliveira:
This implementation is a wonder. It touches tests and documentation, but no actual code 😄

elbrujohalcon:
> This implementation is a wonder. It touches tests and documentation, but no actual code 😄

And how did we make the tests pass, then? 🤔

paulo-ferraz-oliveira:
The tests are Ok because we manipulate the option 😄 as if it were the default.

PR #365, opened 2024-10-01
New `elvis_style` rule: `max_function_clause_length`

### LINKED ISSUE #335: max_function_length counts multiple function clauses as a single function
## Bug Description

**Note**: might not be a bug.

`max_function_length` complains if a function definition spans over a given number of lines (by default: 30)...

... **but** it's also taking into account the multiple function clauses. So 15 clauses with 1 line each, in the body, would count for 30 lines. Is this expected?

## To Reproduce

16 function clauses with 2 line (definition) each, 1 line for head, 1 line for body.

## Expected Behavior

? no warning, maybe?

## `rebar3` Logs

Not applicable.

## Additional Context

Not relevant.

bormilan:
# Description

I added a new rule called `max_function_clause_length`.
The old rule `max_function_length` stays the same, it checks the length of an entire function, but now we can check the size of the clauses separately.

Closes #335


bormilan:
I'm thinking about merging the two rule codes because they are so similar. I should make only one function, that gets parameters(e.g. function or clause), and then call that with the appropriate values.

But on the other hand, I think it's more readable and simple.

What do you think about it?

Oh, and I don't know that the error message is well enough.

src/elvis_style.erl
```diff
@@ -96,6 +96,9 @@
 -define(MAX_FUNCTION_LENGTH,
         "The code for function ~p/~w has ~p lines which exceeds the "
         "maximum of ~p.").
+-define(MAX_FUNCTION_CLAUSE_LENGTH,
+        "The code for clause ~p/~w has ~p lines which exceeds the "
+        "maximum of ~p.").
```
elbrujohalcon, line 101:
```suggestion
        "The code for the clause ~w of function ~p/~w has ~p lines which exceeds the "
        "maximum of ~p.").
```
Or something similar that would point to the actual clause with the issue instead of the whole function.

src/elvis_style.erl
```diff
@@ -891,6 +898,48 @@ max_function_arity(Config, Target, RuleConfig) ->
...
+           {Min, L}
+        end,
+    ClauseLenInfos = lists:map(PairClause, Clauses0),
+
+    MaxLengthPred = fun({_, L}) -> L > MaxLength end,
+    ClauseLenMaxPairs = lists:filter(MaxLengthPred, ClauseLenInfos),
+
+    ResultFun =
+        fun({StartPos, L}) ->
+           Info = [L, MaxLength],
```
elbrujohalcon, line 937:
```suggestion
           Info = [ClauseNumber, L, MaxLength],
```
…aaaaand you would need to somehow compute the clause number 😉

elbrujohalcon (commented):
I added some comments regarding the error message.
Personally, I would not merge the functions. In general, issues are usually raised for one particular rule and not for _all the rules that work in the same way_ so (once again: **in general**) it's usually better to keep each rule's logic in its own function, instead of sharing it… But this might be the exception, of course.

doc_rules/elvis_style/max_function_clause_length.md
```diff
@@ -0,0 +1,27 @@
...
+  - default: `false`
+
+## Example
+
+```erlang
+{elvis_style, max_function_clause_length}
+%% or
+{elvis_style, max_function_clause_length, #{ max_length => 30
+                                    , count_comments => false
+                                    , count_whitespace => false
```
paulo-ferraz-oliveira, line 25:
```suggestion
                                           , count_comments => false
                                           , count_whitespace => false
```

src/elvis_rulesets.erl
```diff
@@ -94,6 +94,7 @@ rules(erl_files_strict) ->
                          <- [always_shortcircuit,
                              consistent_generic_type,
                              max_function_length,
+                             max_function_clause_length,
```
paulo-ferraz-oliveira, line 97:
```suggestion
                             max_function_clause_length,
                             max_function_length,
```

paulo-ferraz-oliveira: (to keep it sorted)

src/elvis_style.erl
```diff
@@ -7,8 +7,8 @@
          invalid_dynamic_call/3, used_ignored_variable/3, no_behavior_info/3,
          module_naming_convention/3, state_record_and_type/3, no_spec_with_records/3,
          dont_repeat_yourself/3, max_module_length/3, max_anonymous_function_arity/3,
-         max_function_arity/3, max_function_length/3, no_call/3, no_debug_call/3,
-         no_common_caveats_call/3, no_nested_try_catch/3, no_successive_maps/3,
+         max_function_arity/3, max_function_length/3, max_function_clause_length/3, no_call/3,
+         no_debug_call/3, no_common_caveats_call/3, no_nested_try_catch/3, no_successive_maps/3,
```
paulo-ferraz-oliveira, line 11: I don't mind the order, here, but preferably changes to existing code bases should touch as little as required. This test could be written (exported) at the end of this "section" and it wouldn't change anything specific, but it would probably change less lines...

src/elvis_style.erl
paulo-ferraz-oliveira: I wouldn't mind having a small translation of

```
1 -> 1st
2 -> 2nd
3 -> 3rd
n -> nth
```

And use that in the message.

Then, as per Brujo's I'd slightly change it to

```diff
-        "The code for clause ~p/~w has ~p lines which exceeds the "
-        "maximum of ~p.").
+        "The code for the ~w clause of function ~p/~w has ~p lines which exceeds the "
+        "maximum of ~p.").

File: test/style_SUITE.erl [hunk 6]
```diff
@@ -17,13 +17,14 @@
...
          verify_dont_repeat_yourself/1, verify_max_module_length/1,
          verify_max_anonymous_function_arity/1, verify_max_function_arity/1,
-         verify_max_function_length/1, verify_no_debug_call/1, verify_no_common_caveats_call/1,
-         verify_no_call/1, verify_no_nested_try_catch/1, verify_no_successive_maps/1,
-         verify_atom_naming_convention/1, verify_no_throw/1, verify_no_dollar_space/1,
-         verify_no_author/1, verify_no_import/1, verify_no_catch_expressions/1,
-         verify_no_single_clause_case/1, verify_numeric_format/1, verify_behaviour_spelling/1,
-         verify_always_shortcircuit/1, verify_consistent_generic_type/1, verify_no_types/1,
-         verify_no_specs/1, verify_export_used_types/1, verify_consistent_variable_casing/1,
+         verify_max_function_length/1, verify_max_function_clause_length/1, verify_no_debug_call/1,
```
paulo-ferraz-oliveira, line 20: If I understand correctly you put `verify_max_function_clause_length/1` in the middle of existing elements, but the review is a bit harder because of this. No need to change for this pull request. Just food for thought...

paulo-ferraz-oliveira:
Thanks much. In general, this looks solid. I propose a few tweaks. From my end, I prefer it if you don't resolve the comments while you handle them, since it allows me to check your reply easier (and not having to press "Show resolved" all the time).

src/elvis_style.erl
bormilan: okay, I'm on it :D

src/elvis_style.erl
elbrujohalcon: Contrary to @paulo-ferraz-oliveira , I do like the fact that the new function was added right after `max_function_length/1`. I guess it's your choice, @bormilan … you get to decide if you want to keep it as-is or move it to the end of the list.

src/elvis_style.erl
bormilan: It looks nice to me, I'm on it.

elbrujohalcon: In that case, it should be…

```suggestion
        "The code for the ~ts  clause of function ~p/~w has ~p lines which exceeds the "
        "maximum of ~p.").
```

…since "1st"/"2nd"/etc… will definitely be strings.

I don't mind if you implement that translation, @bormilan … but careful that it's 11th/12th/13th and not 11st/12nd/13rd, but it's 21st/22nd/23rd (yeah… functions can have lots of clauses) and not 21th/22th/23th.  This is one of those typicall interview exercises… like translating to/from roman numerals 🤦🏻 .
…and you were wondering when will any of this be useful in real-life™️ , right? 😝

src/elvis_style.erl
elbrujohalcon: You'll need to apply the nth/snd/rd translation here, if you decide to go that way.

test/style_SUITE.erl
elbrujohalcon: As I stated before, I do appreciate the fact that the new function appears after the existing one for `max_function_length`… and we don't really need to review changes to this list… if a function needs to be exported and it is not, the test will fail anyway ;)

src/elvis_style.erl
bormilan: Oh okay, it's a little tricky, but I will try to make it as simple as possible.

src/elvis_style.erl
bormilan: Yes, I like it too, but the other side is understandable too. I think I will leave it here, because it's better to keep them close to each other, in my opinion.

bormilan:
I'm thinking about that, we should check the error messages during the tests, to make sure we caught the error message we want. 

We should make a function that takes the expected message/messages and returns a boolean based on the result.

If you think it's useful I would be happy to implement it in another issue maybe, or if not I will just write this test by checking the messages by hand.

bormilan:
I have a short time now, that's why it's half done.

I want to add better test cases, and unit tests for the `case_num` function.

And the "function name" is now bad, so maybe I need to parse the clause_nodes and add the function name and arity to it.

src/elvis_style.erl
```diff
@@ -891,6 +898,73 @@ max_function_arity(Config, Target, RuleConfig) ->
...
+
+parse_case_num(Num, 1) ->
+    io_lib:format("~pst", [Num]);
+parse_case_num(Num, 2) ->
+    io_lib:format("~pnd", [Num]);
+parse_case_num(Num, 3) ->
+    io_lib:format("~prd", [Num]);
+parse_case_num(Num, _) ->
+    io_lib:format("~pth", [Num]).
+
```
elbrujohalcon, line 967: This function seems too complex and with too many parameters.
I would think this version should work:

```suggestion
%% 11th, 12th, 13th, 111th, …
case_num(X) when X rem 100 >= 11, X rem 100 =< 13 -> integer_to_list(X) ++ "th";
%% 1st, 21st, 101st, …
case_num(X) when X rem 10 == 1 -> integer_to_list(X) ++ "st";
%% 2nd, 22nd, 102nd, …
case_num(X) when X rem 10 == 2 -> integer_to_list(X) ++ "nd";
%% 3rd, 23rd, 103rd, …
case_num(X) when X rem 10 == 3 -> integer_to_list(X) ++ "rd";
%% 4th, 5th, 125th, …
case_num(X) -> integer_to_list(X) ++ "th".
```

This is what I usually expect from interviewees when I present them with the problem. :)

elbrujohalcon: But you only use `case_num/3` to get the clause number. In fact, shouldn't it be called `clause_num/1`, instead?

bormilan:
Now, I fixed it a little bit, I hope you find it better too.

elbrujohalcon:
Sorry, @bormilan … but I still fail to see why you need all the recursiveness in that function. In the end it just needs to return a number in string format, doesn't it?

src/elvis_style.erl
```diff
@@ -891,6 +898,71 @@ max_function_arity(Config, Target, RuleConfig) ->
...
+    1.
+
+parse_clause_num(Num) when Num rem 10 == 1 ->
+    integer_to_list(Num) ++ "st";
+parse_clause_num(Num) when Num rem 10 == 2 ->
+    integer_to_list(Num) ++ "nd";
+parse_clause_num(Num) when Num rem 10 == 3 ->
+    integer_to_list(Num) ++ "rd";
+parse_clause_num(Num) ->
+    integer_to_list(Num) ++ "th".
```
elbrujohalcon, line 964: Wouldn't this return `"11st"` for clause number 11 or 111?
That would be incorrect.

That's why I recommended using:
```erlang
%% 11th, 12th, 13th, 111th, …
case_num(X) when X rem 100 >= 11, X rem 100 =< 13 -> integer_to_list(X) ++ "th";
%% 1st, 21st, 101st, …
case_num(X) when X rem 10 == 1 -> integer_to_list(X) ++ "st";
%% 2nd, 22nd, 102nd, …
case_num(X) when X rem 10 == 2 -> integer_to_list(X) ++ "nd";
%% 3rd, 23rd, 103rd, …
case_num(X) when X rem 10 == 3 -> integer_to_list(X) ++ "rd";
%% 4th, 5th, 125th, …
case_num(X) -> integer_to_list(X) ++ "th".
```

bormilan:
Do you mean why do I need the function `clause_num` alongside the fun `parse_clause_num`?
Because I need to figure out the clause's number that is not stored in the data of the node, so I need to calculate it by hand. We have a list of all the clauses in the module, so I need to go through them with a `lists:foldl` and calculate the numbers at every node.

src/elvis_style.erl
bormilan: Oh, you are right, I forgot about 11, 12, 13 ...

bormilan:
Oh maybe my fault was that I searched all the clauses in the `root`, but I might search for them only in the function node. So I need to search for the functions, and in an inside recursive "iteration" I run through the clauses. This way I will end up with a two-deep algorithm, but I don't need to calculate the clause numbers like I did.

bormilan:
This solution would be good in my opinion, but because the fact that I need the function name and arity too, I need to switch to the two-deep solution. When I started this looked better because of simplicity, but I was wrong now I see. (sad)

bormilan:
Now I made the new version, with this I don't need that plus function for the clause numbers. I made a new test, and before merging, I want to make an assertion about the messages(to prove that clause numbers and the message itself are correct) if you think it's good too.

(and of course, I will delete the comments)

elbrujohalcon (approved):

elbrujohalcon:
I'm approving. I think this is mergeable as it is, but if you feel like adding a test for the clause numbers, running elvis on a file like this should produce one error per clause.

```erlang
-module( … ).

-elvis([{max_length_function_clause, #{max_length => 0}}]).

-export([clause/1]).

clause(1) -> "1st";
clause(2) -> "2nd";
…
clause(111) -> "111th";
clause(112) -> "112th";
clause(113) -> "113th";
–
clause(125) -> "125th".
```

elbrujohalcon (approved):
I like it. I'll defer final approval to @paulo-ferraz-oliveira … but everything looks good from my perspective.

bormilan:
Thank you so much!

paulo-ferraz-oliveira:
I tweaked the PR title since this'll go directly into the generated release Changelog.

src/elvis_style.erl
paulo-ferraz-oliveira: If it becomes too convoluted just go with clause #...

src/elvis_style.erl
paulo-ferraz-oliveira: Which is why I prefer a single function per line, in any case... easier to sort, easier to walk through.

test/style_SUITE.erl
paulo-ferraz-oliveira: I wouldn't mind the order if they were sorted already. They're not 😄

paulo-ferraz-oliveira (approved):
I'm Ok, I guess, but I don't know what changed since my last review, since all the commits were squashed.

doc_rules/elvis_style/max_function_clause_length.md
```diff
@@ -0,0 +1,27 @@
...
+This specifies an upper bound on function clause **line** length. Lines that are comments and/or whitespace
+can be either included or excluded from the line count.
+
+> Works on `.beam` file? Not really! (it consumes results Ok, but these might be unexpected, since
+the files are pre-processed)
+
+## Options
+
+- `max_length :: non_neg_integer()`.
+  - default: `30`.
```
paulo-ferraz-oliveira, line 12:
```suggestion
- `max_length :: non_neg_integer()`
  - default: `30`
```

Untitled
paulo-ferraz-oliveira, line 1: Is this on purpose? I'd remove it otherwise...

src/elvis_style.erl
```diff
@@ -96,6 +96,9 @@
 -define(MAX_FUNCTION_LENGTH,
         "The code for function ~p/~w has ~p lines which exceeds the "
         "maximum of ~p.").
+-define(MAX_FUNCTION_CLAUSE_LENGTH,
+        "The code for the ~ts  clause of function ~p/~w has ~p lines which exceeds the "
```
paulo-ferraz-oliveira, line 100:
```suggestion
        "The code for the ~ts clause of function ~p/~w has ~p lines which exceeds the "
```

paulo-ferraz-oliveira:
I left a couple of comments above, that should be handled; after that we're good to go. Thanks.

bormilan:
> I'm Ok, I guess, but I don't know what changed since my last review, since all the commits were squashed.

Oh, sorry next time I will wait with it till the very end.

Untitled
paulo-ferraz-oliveira, line 1: Is this file (named `Untitled` with content `tyl`) added on purpose? Or can it be removed?

bormilan: Sorry, I removed it. I don't know how I missed that when I committed my changes.

test/style_SUITE.erl
```diff
@@ -1142,6 +1143,78 @@ verify_max_function_length(Config) ->
...
+             "7th",
+             "11th",
+             "13th",
+             "19th",
+             "20th",
+             "21st",
+             "29th",
+             "30th",
+             "31st",
+             "33rd"].
```
elbrujohalcon, line 1216: This won't test anything… for this to assert something, you should use…
```suggestion
    Numbers = ["1st",
             "5th",
             "6th",
             "7th",
             "11th",
             "13th",
             "19th",
             "20th",
             "21st",
             "29th",
             "30th",
             "31st",
             "33rd"].
```

elbrujohalcon: But, while that would work, in correct CT _verbiage_, the right way of write an assertion is…
```suggestion
    ["1st",
             "5th",
             "6th",
             "7th",
             "11th",
             "13th",
             "19th",
             "20th",
             "21st",
             "29th",
             "30th",
             "31st",
             "33rd"] = Numbers.
```

(i.e., with the expected value to the left)

elbrujohalcon (changes requested):

elbrujohalcon:
Almost there, @bormilan

test/style_SUITE.erl
bormilan: Oh, you're right. Always the expected on the left.

elbrujohalcon (approved):

elbrujohalcon:
Thanks, @bormilan !!! Amazing work.

PR #367, opened 2024-10-07
New `elvis_style` rule: `no_init_lists`

### LINKED ISSUE #329: Rule: Do not use lists on start_link/3,4 and init/1
### Name

# No Init Lists

### Brief Description

The idea is to avoid code like this one…

```erlang
-module my_server.
% …exports and everything else…
start_link(AParam) ->
  gen_server:start_link(?MODULE, [AParam], []).

init([AParam]) ->
  …
```

### Reasoning

See [this thread in the forums](https://erlangforums.com/t/args-in-gen-init-1/3169/5?u=elbrujohalcon)

### Refactoring Proposal

Use maps, tuples, or… if you need to ignore the parameter at all, use an atom like `undefined` or `no_argument`.

bormilan:
# Description

A brief description of your changes.

Closes #329;.

---

EDIT: I made a checklist based on @elbrujohalcon 's comment to to make the progress traceable.

- [x] has init([_|_]) -> …: Should fail.
- [x] has init(A = [1,2,3]) -> …: Should fail
- [x] has init/1 implemented with multiple clauses, all of them expecting a list: Should fail.
- [x] - [ ] A module that has two behaviours and has init([]) -> should fail

- [x] has init/1 implemented with multiple clauses, one of them not being a list: Should pass, even if other clauses expect a list.
- [x] has init/1 implemented using a map, but init/2 implemented using a list in one of the arguments: Should pass.
- [x] has init/0 implemented using a tuple, but also init/0: Should pass.
- [x] A module that doesn't implement a behaviour and has init([1,2,3]) -> …: Should pass.
- [x] A module that has a behaviour that is not in the list of the rule's config -> should pass(with default config)


RULES.md
```diff
@@ -61,6 +61,7 @@ identified with `(since ...)` for convenience purposes.
 - [State Record and Type](doc_rules/elvis_style/state_record_and_type.md)
 - [Used Ignored Variable](doc_rules/elvis_style/used_ignored_variable.md)
 - [Variable Naming Convention](doc_rules/elvis_style/variable_naming_convention.md)
+- [Not Use List In Init Functions](doc_rules/elvis_style/no_init_lists.md)
```
elbrujohalcon, line 64:
```suggestion
- [No Init Lists](doc_rules/elvis_style/no_init_lists.md)
```

It's better if the name of the rule matches with the md file… and it's also the title there, so… everything matches :)

bormilan:
Do we want to check that the module implements a `gen_*` behavior, or should we check every module like now?

doc_rules/elvis_style/no_init_lists.md
```diff
@@ -0,0 +1,17 @@
+# No Init Lists
+
+This warns you if you use list as a parameter in an init function in a gen_* module.
+[Reasoning](https://erlangforums.com/t/args-in-gen-init-1/3169/4?u=elbrujohalcon)
```
elbrujohalcon, line 4:
```suggestion
Do not use a list as the parameter for the `init/1` callback when implementing `gen_*` behaviours. It's semantically clearer to use a tuple or a map. [More info](https://erlangforums.com/t/args-in-gen-init-1/3169/4?u=elbrujohalcon).
```

doc_rules/elvis_style/no_init_lists.md
```diff
@@ -0,0 +1,17 @@
...
+
+> Works on `.beam` file? Not really! (it consumes results Ok, but these might be unexpected, since
+the files are pre-processed)
```
elbrujohalcon, line 7:
```suggestion
> Works on `.beam` files? Yes!
```

src/elvis_style.erl
```diff
@@ -30,6 +31,10 @@
               no_match_in_condition_config/0, behaviour_spelling_config/0,
               param_pattern_matching_config/0, private_data_type_config/0]).
 
+-hank([{unnecessary_function_arguments, [{no_init_lists, 3}]}]).
+
+-define(NO_INIT_LISTS_MSG,
+        "Don't use a list as a parameter in 'init' function at position ~p.").
```
elbrujohalcon, line 37:
```suggestion
        "Do not use a list as the parameter for the 'init' callback at position ~p.").
```

test/examples/fail_verify_no_init_lists.erl
```diff
@@ -0,0 +1,9 @@
+-module(fail_verify_no_init_lists).
+
+-export([start_link/1, init/1]).
+
+start_link(AParam) ->
+  gen_server:start_link(?MODULE, [AParam], []).
+
+init([_AParam]) ->
```
elbrujohalcon, line 8: This file should actually _work_… It should not fail since it's not implementing any behaviour.
If the file is not implementing a behaviour that requires an `init/1` callback, we don't care how `init/1` is implemented.

elbrujohalcon (changes requested):
This needs more tests, specifically:
* A module that implements `gen_server` or `gen_statem` and…
  - has `init([_|_]) -> …`: Should fail.
  - has `init(A = [1,2,3]) -> …`: Should fail
  - has `init/1` implemented with multiple clauses, all of them expecting a list: Should fail.
  - has `init/1` implemented with multiple clauses, one of them not being a list: Should **pass**, even if other clauses expect a list.
  - has `init/1` implemented using a map, but `init/2` implemented using a list in one of the arguments: Should **pass**.
  - has `init/0` implemented using a tuple, but also `init/0`: Should **pass**.
* A module that doesn't implement a behaviour and has `init([1,2,3]) -> …`: Should **pass**.

doc_rules/elvis_style/no_init_lists.md
paulo-ferraz-oliveira: I think it's safe to remove `?u=elbrujohalcon` from the link.

test/examples/fail_verify_no_init_lists.erl
paulo-ferraz-oliveira: Agreed, to start we should just go with a few hard-coded Erlang behaviours; then we can extend as time goes by (they don't come by that often, in any case).

paulo-ferraz-oliveira (commented):
You're on the right path here ❤️ Just a few more tweaks...

paulo-ferraz-oliveira:
I tweaked the PR title since this'll go directly into the generated release Changelog.

bormilan:
Thanks for the comments, I will solve all of them if I have some time. <3

doc_rules/elvis_style/no_init_lists.md
bormilan: You are right I removed and changed the last id from `4` to `5` to point the link to the right comment.

src/elvis_style.erl
```diff
@@ -1029,6 +1035,36 @@ atom_naming_convention(Config, Target, RuleConfig) ->
     AtomNodes = elvis_code:find(fun is_atom_node/1, Root, #{traverse => all, mode => node}),
     check_atom_names(Regex, RegexEnclosed, AtomNodes, []).
 
+-spec no_init_lists(elvis_config:config(), elvis_file:file(), empty_rule_config()) ->
+                       [elvis_result:item()].
+no_init_lists(_Config, Target, _RuleConfig) ->
+    Root = get_root(#{}, Target, #{}),
```
elbrujohalcon, line 1041:
```suggestion
no_init_lists(Config, Target, RuleConfig) ->
    Root = get_root(Config, Target, RuleConfig),
```

Did you need to use empty maps there for a reason, @bormilan ?

src/elvis_style.erl
```diff
@@ -1029,6 +1035,36 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+           Name = ktn_code:attr(name, FunctionNode),
+           Location = ktn_code:attr(location, FunctionNode),
+           [Content] = ktn_code:content(FunctionNode),
+           Attributes = ktn_code:node_attr(pattern, Content),
+           {Name, Location, [Attr || #{type := Type} = Attr <- Attributes, Type == cons]}
+        end,
+
+    FunListAttributeInfos = lists:map(PairFun, FunctionNodes),
+
+    FilterFun = fun({Name, _, C}) -> length(C) > 0 andalso Name =:= init end,
```
elbrujohalcon, line 1057: I'm not sure but… if `C` is the list of arguments… the it has to be **exactly** 1.
```suggestion
    FilterFun = fun({Name, _, Args}) -> length(Args) =:= 1 andalso Name =:= init end,
```

elbrujohalcon (changes requested):
Still missing some changes, but I just wanted to add feedback about the recent additions.

src/elvis_style.erl
bormilan: Yes, you are right it's a dumb fault.

src/elvis_style.erl
bormilan: Yes, you are right again, I just left it in...

bormilan:
I still have some work to do, but it keeps getting better.

elbrujohalcon:
@bormilan do you mind merging `main` into your branch, please?

src/elvis_style.erl
```diff
@@ -1,5 +1,7 @@
 -module(elvis_style).
 
+-feature(maybe_expr, enable).
```
elbrujohalcon, line 3: I don't think you need this anymore, do you?

elbrujohalcon: Actually… this is a great idea for a Hank rule: _Unused features_ 🤔

elbrujohalcon: https://github.com/AdRoll/rebar3_hank/issues/221

src/elvis_style.erl
```diff
@@ -30,6 +33,8 @@
               no_match_in_condition_config/0, behaviour_spelling_config/0,
               param_pattern_matching_config/0, private_data_type_config/0]).
 
+% -define(NO_INIT_LISTS_MSG,
+%         "Do not use a list as the parameter for the 'init' callback at position ~p.").
```
elbrujohalcon, line 37: I'm gonna guess we'll need this macro before merging the PR, right?

src/elvis_style.erl
```diff
@@ -1029,6 +1036,59 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+    case elvis_code:find(IsBehaviour, Root) of
+        [BehaviourNode] ->
+            case lists:member(
+                     ktn_code:attr(value, BehaviourNode), Behaviors)
+            of
+                true ->
+                    true;
+                false ->
+                    false
+            end;
```
elbrujohalcon, line 1087:
```suggestion
            lists:member(ktn_code:attr(value, BehaviourNode), Behaviors)
```

test/examples/no_init_lists_examples/pass_no_init_lists2.erl
```diff
@@ -0,0 +1,18 @@
+-module(pass_no_init_lists2).
+
+-behaviour(gen_server).
+
+-export([start_link/0, init/1, init/2, handle_cast/2, handle_call/3]).
+
+start_link() ->
+  gen_server:start_link(?MODULE, undefined, []).
```
elbrujohalcon, line 8:
```suggestion
  gen_server:start_link(?MODULE, #{a => map}, []).
```
Subtle and pointless, but just for consistency (i.e., to match what `init/1` expects).

test/examples/no_init_lists_examples/pass_no_init_lists3.erl
```diff
@@ -0,0 +1,18 @@
+-module(pass_no_init_lists3).
+
+-behaviour(gen_server).
+
+-export([start_link/0, init/0, init/1, handle_cast/2, handle_call/3]).
+
+start_link() ->
+  gen_server:start_link(?MODULE, undefined, []).
```
elbrujohalcon, line 8:
```suggestion
  gen_server:start_link(?MODULE, {1, 2}, []).
```
Subtle and pointless, but just for consistency (i.e., to match what `init/1` expects).

test/examples/no_init_lists_examples/pass_no_init_lists.erl
```diff
@@ -0,0 +1,21 @@
+-module(pass_no_init_lists).
+
+-behaviour(gen_server).
+
+-export([start_link/0, init/1, handle_cast/2, handle_call/3]).
+
+start_link() ->
+  gen_server:start_link(?MODULE, undefined, []).
```
elbrujohalcon, line 8:
```suggestion
  gen_server:start_link(?MODULE, 1, []).
```
Subtle and pointless, but just for consistency (i.e., to match what `init/1` expects).

elbrujohalcon (changes requested):
A few small suggestions, but overall great progress @bormilan 👍🏻

src/elvis_style.erl
bormilan: Oh...
🤦‍♂️

src/elvis_style.erl
bormilan: Yes, I commented it out because I just experienced a weird bug that said it's not used. (But I clearly used it) I will investigate it, this is a temporary solution.

src/elvis_style.erl
```diff
@@ -1029,6 +1034,52 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+                       Location = ktn_code:attr(location, FunctionNode),
+                       [Content] = ktn_code:content(FunctionNode),
+                       Attributes = ktn_code:node_attr(pattern, Content),
+                       {Name,
+                        Location,
+                        [Attr || #{type := Type} = Attr <- Attributes, Type == cons]}
+                    end,
+                FunListAttributeInfos = lists:map(PairFun, FunctionNodes),
+                FilterFun = fun({Name, _, Args}) -> length(Args) =:= 1 andalso Name =:= init end,
+                lists:filter(FilterFun, FunListAttributeInfos);
```
elbrujohalcon, line 1143:
```suggestion
                [Location || {init, Location, [_]} <- FunListAttributeInfos];
```
You can use list comprehensions to filter… and, since you only need the location later, you can just retrieve that.

src/elvis_style.erl
```diff
@@ -1029,6 +1034,52 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+                []
+        end,
+
+    ResultFun =
+        fun({_, Location, _}) ->
+           Info = [Location],
+           Msg = ?NO_INIT_LISTS_MSG,
+           elvis_result:new(item, Msg, Info, Location)
+        end,
+    lists:map(ResultFun, FunListAttributes).
```
elbrujohalcon, line 1154: If you follow my suggestion above, you can simplify this…
```suggestion
    ResultFun =
        fun(Location) ->
           Info = [Location],
           Msg = ?NO_INIT_LISTS_MSG,
           elvis_result:new(item, Msg, Info, Location)
        end,
    lists:map(ResultFun, FunListAttributes).
```

elbrujohalcon, line 1154: You can (probably… I had not tested it) simplify this multi-loop algorithm on a single list comprehension…
```suggestion
    case is_rule_behaviour(Root, RuleConfig) of
        true ->
            IsInit =
                fun(Node) ->
                    ktn_code:type == function
                    andalso ktn_code:attr(name, FunctionNode) == init
                end,
            InitFunctions = elvis_code:find(IsInit, Root),
            [ elvis:result:new(
                item,
                ?NO_INIT_LISTS_MSG,
                [ktn_code:attr(location, FunctionNode)],
                ktn_code:attr(location, FunctionNode))
            || FunctionNode <- InitFunctions, 
                Content <- ktn_code:content(FunctionNode),
                length(ktn_code:node_attr(pattern, Content)) == 1, %% We only care about init/1
                Attr <- ktn_code:node_attr(pattern, Content),
                cons == ktn_code:type(Attr) %% We only raise warnings if the pattern is a list
            ];
        false ->
            []
    end.
```

src/elvis_style.erl
```diff
@@ -1029,6 +1034,52 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+
+is_rule_behaviour(Root, RuleConfig) ->
```
elbrujohalcon, line 1156: Renaming suggestion: `is_relevant_behaviour` or `is_important_behaviour` or `is_known_behaviour`.
`is_rule_behaviour`, to me, sounds like if it's looking for `-behaviour(rule).`

elbrujohalcon (changes requested):
One more test (or maybe, two), @bormilan:

* A module that implements `gen_server` AND another behaviour…

```erlang
-module(fail…).

-behaviour(gen_server).
-behaviour(another_behaviour).

-export([init/1]).

init([]) -> {error, "should not be a list"}.
```

bormilan:
> One more test (or maybe two, if you want a fail and a pass), @bormilan:
> 
> * A module that implements `gen_server` AND another behaviour…
> 
> ```erlang
> -module(fail…).
> 
> -behaviour(gen_server).
> -behaviour(another_behaviour).
> 
> -export([init/1]).
> 
> init([]) -> {error, "should not be a list"}.
> ```

Yes! I like making more and more tests to feel that my code is 100% good.

elbrujohalcon:
> Yes! I like making more and more tests to feel that my code is 100% good.

HA! I just saw that you were matching against a single `[BehaviourNode]` in your code… and I wondered "what would happen if we have multiple ones?"… so… we have to test that, right?

bormilan:
> > Yes! I like making more and more tests to feel that my code is 100% good.
> 
> HA! I just saw that you were matching against a single `[BehaviourNode]` in your code… and I wondered "what would happen if we have multiple ones?"… so… we have to test that, right?

You're right. I'm working on the new logic right now, so it's the perfect time to add new tests and consider all the possible barriers.

src/elvis_style.erl
bormilan: Oh, I was so engrossed in work that I had not checked your comments and I reworked all this... (And I didn't get notifications, which is weird)

src/elvis_style.erl
bormilan: Good idea, thanks

src/elvis_style.erl
bormilan: But I think I can make something like this in my new code too.

bormilan:
The only test that is not working now is the one with the `init([])` because its attribute type is `nil`. I don't know how I will solve this, it may need some extra magic. Or is it a bug in `katana_code`?

elbrujohalcon:
> The only test that is not working now is the one with the `init([])` because its attribute type is `nil`. I don't know how I will solve this, it may need some extra magic. Or is it a bug in `katana_code`?

Isn't it as simple as `lists:member(ktn_code:type(Attr), [cons, nil])`?

I think `nil` is precisely `[]`, which is fine.

src/elvis_style.erl
```diff
@@ -1112,6 +1117,82 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+
+-spec no_init_lists(elvis_config:config(), elvis_file:file(), no_init_lists_config()) ->
+                       [elvis_result:item()].
+no_init_lists(Config, Target, RuleConfig) ->
+    Root = get_root(Config, Target, RuleConfig),
+
+    InitFuns =
+        case is_relevant_behaviour(Root, RuleConfig) of
+            true ->
+                IsFunction = fun(Node) -> ktn_code:type(Node) == function end,
```
elbrujohalcon, line 1130: I still think it would be better to filter out more functions here…
```suggestion
                IsFunction = fun(Node) -> ktn_code:type(Node) == function andalso ktn_code:attr(name, Node) == init end,
```
…even more… if we can only keep the ones that have a single argument through this filter, that would simplify the remaining part of this function code much more.

src/elvis_style.erl
```diff
@@ -1112,6 +1117,82 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+                                       end
+                                    end,
+                                    FunctionNodes));
+            false ->
+                []
+        end,
+
+    IsBreakRule =
+        lists:all(fun({_, Attributes}) ->
+                     length(lists:filter(fun(Elem) -> is_list_node(Elem) end, Attributes)) =:= 1
```
elbrujohalcon, line 1160: This would be true for a function like `init([_], second_argument) -> …` (it has 1 list argument) and that function does **not** break the rule.

elbrujohalcon: Please add a test for that.

elbrujohalcon (changes requested):

src/elvis_style.erl
bormilan: Okay, thanks for the feedback!

bormilan:
> precisely

Oh! Okay, that's nice, I started to dig into the code of katana, thanks for the clarification.

bormilan:
> > The only test that is not working now is the one with the `init([])` because its attribute type is `nil`. I don't know how I will solve this, it may need some extra magic. Or is it a bug in `katana_code`?
> 
> Isn't it as simple as `lists:member(ktn_code:type(Attr), [cons, nil])`?
> 
> I think `nil` is precisely `[]`, which is fine.

Oh! Okay, that's nice, I started to dig into the code of katana, thanks for the clarification.

src/elvis_style.erl
bormilan: Oh yes, thanks!

bormilan: If I have a full understanding, you say this because we don't want to really know what the args of `init/2` are, because `gen_server` needs `init/1` at every case, and we only want to check that if `init/1` expects a list as its parameter.

So we don't care if `init/2` expects a list at any of its parameters, because if `init/1` expects a map we will pass it.
And we don't care if `init/2` expects two non-list parameters, because if `init/1` expects a list we will fail.

Am I right on that, or do you think something else?

elbrujohalcon: Exactly! All OTP behaviours expect a single argument in `init(…)`. That's what kickstarted this whole thing. `gen_server` docs use a list as _InitArgs_… while in reality it's just a single _InitArg_ that doesn't need to be a list at all.
That's what confuses devs (particularly, newcommers) about this thing.

bormilan: Ohhh I see. Now it's finally clear, it's nice that I started working on this so I could learn this :D

src/elvis_style.erl
```diff
@@ -147,6 +150,8 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 -spec default(Rule :: atom()) -> DefaultRuleConfig :: term().
+default(no_init_lists) ->
+    #{behaviours => [gen_server, gen_statem]};
```
elbrujohalcon, line 154:
```suggestion
    #{behaviours => [gen_server, gen_statem, gen_fsm, supervisor, supervisor_bridge, gen_event]};
```

Let's add the other ones that I found while grepping for `-callback init` in OTP's codebase :)

elbrujohalcon: If you add a simple test for each one of them… like the module below… much better. If not, no big deal.

```erlang
-module(fail_{BEHAVIOUR}).

-behaviour({BEHAVIOUR}).

-export([init/1]).

init([]) -> {error, "Don't use lists for init/1"}.
```

(Replacing `{BEHAVIOUR}` with the name of each behaviour, obviously).

elbrujohalcon (changes requested):
We're almost there, @bormilan … really really close to the final version.

src/elvis_style.erl
```diff
@@ -1112,6 +1117,76 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+    ListInitClauses =
+        case is_relevant_behaviour(Root, RuleConfig) of
+            true ->
+                IsInit1Function =
+                    fun(Node) ->
+                       ktn_code:type(Node) == function
+                       andalso ktn_code:attr(name, Node) == init
+                       andalso ktn_code:attr(arity, Node) == 1
+                    end,
+                [Init1Fun] = elvis_code:find(IsInit1Function, Root),
```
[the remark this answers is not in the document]
bormilan: umm, I'm trying but the compilation fails if there are some undefined callback functions.

src/elvis_style.erl
bormilan: yes, I'm going to add all of them.

src/elvis_style.erl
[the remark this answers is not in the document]
elbrujohalcon: So the compiler compiles all the modules that we use for testing? I was not aware of that. In any case, you can add a `-compile(…).` attribute to the module to avoid warnings. I'll see if I can find the right value to put there.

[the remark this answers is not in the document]
elbrujohalcon: I added the module (granted: I didn't add no test for it) locally and it seems to compile just fine. What error are you seeing?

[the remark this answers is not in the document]
bormilan: I see, something like this: -compile([{nowarn_unused_function}]).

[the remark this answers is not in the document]
elbrujohalcon: Well… it seems like there is no way to bypass the behaviour check in the Erlang compiler, so… Do not add such a test, but do fix the code to account for `elvis_code:find(…)` returning `[]` here.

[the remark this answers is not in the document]
elbrujohalcon: Drop a note like this one when you do it…
```erlang
case elvis_code:find(IsInit1Function, Root) of
    [] ->
        %% A module that implements the behaviour without defining init/1 is valid
        %% but adding it to our tests will prevent us from compiling the project.
        %% That's why this part of the code will remain not covered by tests.
        [];
    [Init1Fun] ->
        …
```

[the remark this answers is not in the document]
bormilan: Okay, I will make it! Thank you for the help!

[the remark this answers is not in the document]
elbrujohalcon: https://github.com/erlang/otp/issues/8985

src/elvis_style.erl
```diff
@@ -1112,6 +1118,83 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+                       andalso ktn_code:attr(arity, Node) == 1
+                    end,
+
+                Init1Fun =
+                    case elvis_code:find(IsInit1Function, Root) of
+                        [] ->
+                            [];
+                        [Fun] ->
+                            Fun
+                    end,
```
elbrujohalcon, line 1144: This won't work, @bormilan … `Init1Fun` being `[]` will only cause errors down the line. I'll make a code suggestion that's more likely to work in this case…

src/elvis_style.erl
```diff
@@ -1112,6 +1118,83 @@ atom_naming_convention(Config, Target, RuleConfig) ->
...
+                ListAttrClauses = lists:filtermap(FilterClauses, Content),
+                case length(ListAttrClauses) =:= length(Content) of
+                    true ->
+                        ListAttrClauses;
+                    false ->
+                        []
+                end;
+            false ->
+                []
+        end,
```
elbrujohalcon, line 1167:
```suggestion
    ListInitClauses =
        case is_relevant_behaviour(Root, RuleConfig) of
            true ->
                IsInit1Function =
                    fun(Node) ->
                       ktn_code:type(Node) == function
                       andalso ktn_code:attr(name, Node) == init
                       andalso ktn_code:attr(arity, Node) == 1
                    end,

                case elvis_code:find(IsInit1Function, Root) of
                    [] ->
                        [];
                    [Init1Fun] ->
                        FilterClauses =
                            fun(Clause) ->
                                [Attribute] = ktn_code:node_attr(pattern, Clause),
                                case is_list_node(Attribute) of
                                    true ->
                                        {true, ktn_code:attr(location, Clause)};
                                    false ->
                                        false
                                end
                            end,

                        Content = ktn_code:content(Init1Fun),
                        ListAttrClauses = lists:filtermap(FilterClauses, Content),
                        case length(ListAttrClauses) =:= length(Content) of
                            true ->
                                ListAttrClauses;
                            false ->
                                []
                        end;
                end;
            false ->
                []
        end,
```

elbrujohalcon: Something like this ☝🏻 should work

elbrujohalcon (changes requested):

src/elvis_style.erl
bormilan: yes, I just wrote this without thinking. The problem was exactly this even at the beginning of this fix.. 😅

Untitled
```diff
@@ -0,0 +1 @@
+
```
elbrujohalcon, line 1: Is this file needed?

elbrujohalcon (changes requested):
Excellent job, @bormilan !! Only one more change.

Untitled
bormilan: I'm sorry about that. I have two computers, and they have different nvim setups, so I usually mess up with the shortcuts. 😅

elbrujohalcon (approved):

paulo-ferraz-oliveira:
Too bad we can't silence this (by default) if the function it detects the issue in has spec `-> supervisor:child_spec().` because as-is we're going "against" OTP documentation: not sure we have an issue to propose a change there... 🤔

elbrujohalcon:
@paulo-ferraz-oliveira do you have an example of that? Because `child_spec/0` should define the parameters you would pass to `your_module:start_link/#`, not the ones you pass to `gen_something:start_link/#`, AFAIR.

paulo-ferraz-oliveira:
Yeah, so I did it in a hurry, and clearly `{M, F, {}}` is wrong, since it should be `{M, F, [{}]}`, but this is again something that might be confusing to first-time users. It is what it is, though.

paulo-ferraz-oliveira:
Fwiw, here's the correct example

```erlang
child_spec() ->
  #{
    id => mylib,
    start => {mylib, start_link, [_StartLinkArgs = {}]}
  }.

start_link(StartLinkArgs) ->
  gen_server:start_link(?MODULE, _InitArgs = {}, []).

init(InitArgs) ->
  {ok, state}. % state, in this case, is "whatever".
```

PR #368, opened 2024-10-30
Closes #311 - fix no_match_in_condition reports a warning even when the condition includes a valid match

### LINKED ISSUE #311: no_match_in_condition reports a warning even when the condition includes a valid match
## Bug Description

In the scenario below, the rule should not emit a warning.

## To Reproduce

Run Elvis on…

```erlang
-module(x).

-export([valid/1]).

valid(List) ->
    case lists:all(fun ({K, V}) ->
                           Something = find:something(for, K),
                           check:something(V, Something)
                   end,
                   List)
        of
        true -> all_true;
        false -> found_a_bad_one
    end.
```

You'll get…

```erlang
# src/x.erl [FAIL]
  - no_match_in_condition (https://github.com/inaka/elvis_core/tree/main/doc_rules/elvis_style/no_match_in_condition.md)
    - Case statement with a match in its condition found on line 6.
```

## Expected Behavior

No warnings.

bormilan:
# Description

My idea was to check only the "first layer of children" of the `case` type Node.
And I added the type `block` too.

Closes #311;.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1302,7 +1302,8 @@ no_match_in_condition(Config, Target, RuleConfig) ->
               CaseNodes).
 
 is_match_in_condition(Node) ->
-    ktn_code:type(Node) == case_expr andalso [] =/= elvis_code:find(fun is_match/1, Node).
+    (ktn_code:type(Node) == case_expr orelse ktn_code:type(Node) == block)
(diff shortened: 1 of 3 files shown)


elbrujohalcon (approved):

elbrujohalcon:
I'll let @paulo-ferraz-oliveira review, too... but from my perspective, this is 100% mergeable.

PR #369, opened 2024-10-31
Closes #351 - fix false positive on no_nested_try_catch

### LINKED ISSUE #351: False positive on no_nested_try_catch
## Bug Description

`no_nested_try_catch` reports a warning on a `try…catch` block that's not really _nested_ within another.

## To Reproduce

Run Elvis on the following module:

```erlang
-module(example).

-export([example/0]).

%% @doc If this fails in do:something/0, we handle the error.
%%      But if it fails in do:something_else/2, we ignore it.
example() ->
    try do:something() of
        {ok, KeepGoing} ->
            try
                do:something_else("and", KeepGoing)
            catch
                _:_ -> {ignore, errors, here}
            end
    catch
        Kind:Error ->
            {Kind, Error}
    end.
```

You'll get the following warning:

```erlang
# src/example.erl [FAIL]
  - no_nested_try_catch (https://github.com/inaka/elvis_core/tree/main/doc_rules/elvis_style/no_nested_try_catch.md)
    - Nested try...catch block starting at line 10.
```

## Expected Behavior

No warnings, since the second `try…catch` is not really nested within the first one (i.e., it's not `try … try … catch … end … catch … end`, which is what the rule tries to prevent).

## Additional Context

This _equivalent_ version of the code works:

```erlang
-module(example).

-export([example/0]).

%% @doc If this fails in do:something/0, we handle the error.
%%      But if it fails in do:something_else/2, we ignore it.
example() ->
    Result =
        try
            do:something()
        catch
            Kind:Error ->
                {Kind, Error}
        end,
    case Result of
        {ok, KeepGoing} ->
            try
                do:something_else("and", KeepGoing)
            catch
                _:_ -> {ignore, errors, here}
            end;
        {Kind, Error} ->
            {Kind, Error}
    end.
```

bormilan:
# Description

The same as [here](https://github.com/inaka/elvis_core/pull/368)

Maybe I should add more tests, If you have suggestions, feel free to comment. Thanks!

Closes #351.

### DIFF
--- a/test/examples/pass_no_nested_try_catch.erl
+++ b/test/examples/pass_no_nested_try_catch.erl
@@ -14,5 +14,7 @@ example() ->
             end
     catch
         Kind:Error ->
-            {Kind, Error}
+            try this:block(is, also, not_nested) catch _:_ -> {Kind, Error} end
+    after
+        try this:one(is, "not", nested, either) catch _:_ -> ok end
     end.
--- a/test/examples/pass_no_nested_try_catch.erl
+++ b/test/examples/pass_no_nested_try_catch.erl
@@ -1,0 +1,18 @@
+-module(pass_no_nested_try_catch).
+
+-export([example/0]).
+
+%% @doc If this fails in do:something/0, we handle the error.
+%%      But if it fails in do:something_else/2, we ignore it.
+example() ->
+    try do:something() of
+        {ok, KeepGoing} ->
+            try
+                do:something_else("and", KeepGoing)
+            catch
+                _:_ -> {ignore, errors, here}
+            end
+    catch
+        Kind:Error ->
+            {Kind, Error}
+    end.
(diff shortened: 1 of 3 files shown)


test/examples/pass_no_nested_try_catch.erl
```diff
@@ -0,0 +1,18 @@
...
+    try do:something() of
+        {ok, KeepGoing} ->
+            try
+                do:something_else("and", KeepGoing)
+            catch
+                _:_ -> {ignore, errors, here}
+            end
+    catch
+        Kind:Error ->
+            {Kind, Error}
```
elbrujohalcon, line 17:
```suggestion
        Kind:Error ->
            try this:block(is, also, not_nested) catch _:_ -> {Kind, Error} end
    after
        try this:one(is, "not", nested, either) catch _:_ -> ok end
```

This would ensure that no part of the try…catch…after…end block is checked, other than the actual expresion in the `try`.

elbrujohalcon (changes requested):
I would extend the test as I suggested, but… other than that, I think this is great.

elbrujohalcon (approved):

PR #371, opened 2024-11-05
change(export_used_types): don't warn on behaviour callbacks

### LINKED ISSUE #308: There are instances where not exporting a type is valid, even if it's used in a spec.
## Bug Description

In the module I'm showing below, Elvis will ask me to export the `state/0` type. That's not a good idea.

## To Reproduce

Run Elvis on this module:

```erlang
-module(x).

-behaviour(gen_server).

-record(state, {a = field}).

-type state() :: #state{}.

-export([init/1, handle_cast/2]).

-spec init(_) -> {ok, state()}.
init(_) -> {ok, #state{}}.

-spec handle_cast(_, state()) -> {noreply, state()}.
handle_cast(_, State) -> {noreply, State}.
```

You'll get…

```erlang
# src/x.erl [FAIL]
  - export_used_types (https://github.com/inaka/elvis_core/tree/main/doc_rules/elvis_style/export_used_types.md)
    - Type state/0, defined on line 7, is used by an exported function but not exported itself
```

## Expected Behavior

No warnings.

## Proposed Solution
We already faced a similar issue in `rebar3_hank`. Until erlang/otp#4847 is dealt with and we get the ability to specify dynamic callbacks, Elvis should avoid emitting this type of warning for types used in exported functions of modules implementing a behavior.
It may be slightly smarter and (just like hank) use the fact that some behaviors (like `gen_server`) are _well-known_ and they don't have dynamic callbacks.

bormilan:
# Description

I implemented a way to get the callbacks for the behaviors in the module and subtract them from the function used in the rule's logic. This approach ensures that all callback types are ignored.

Additionally, there was a test that expected a warning from a module that no longer generates one. So, I removed that assertion, but I'm not sure if that was the correct decision.
Closes #308.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1500,8 +1500,13 @@ get_behaviour_callbacks(Root) ->
     BehaviourNames =
         lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
 
-    lists:append(
-        lists:map(fun(B) -> apply(B, behaviour_info, [callbacks]) end, BehaviourNames)).
+    try lists:map(fun(B) -> apply(B, behaviour_info, [callbacks]) end, BehaviourNames) of
+        DeepList ->
+            lists:append(DeepList)
+    catch
+        _ ->
+            []
+    end.
 
 get_type_of_type(#{type := type_attr,
                    node_attrs := #{type := #{attrs := #{name := TypeOfType}}}}) ->
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1460,13 +1460,15 @@ always_shortcircuit(Config, Target, RuleConfig) ->
                            [elvis_result:item()].
 export_used_types(Config, Target, RuleConfig) ->
     TreeRootNode = get_root(Config, Target, RuleConfig),
+    IgnoredFunctions = get_behaviour_callbacks(TreeRootNode),
     ExportedFunctions = elvis_code:exported_functions(TreeRootNode),
+    FilteredExportedFunctions = lists:subtract(ExportedFunctions, IgnoredFunctions),
     ExportedTypes = elvis_code:exported_types(TreeRootNode),
     SpecNodes =
         elvis_code:find(fun is_spec_attribute/1, TreeRootNode, #{traverse => all, mode => node}),
     ExportedSpecs =
         lists:filter(fun(#{attrs := #{arity := Arity, name := Name}}) ->
-                        lists:member({Name, Arity}, ExportedFunctions)
+                        lists:member({Name, Arity}, FilteredExportedFunctions)
                      end,
                      SpecNodes),
     UsedTypes =
@@ -1491,6 +1493,15 @@ export_used_types(Config, Target, RuleConfig) ->
                  elvis_result:new(item, ?EXPORT_USED_TYPES_MSG, [Name, Arity, Line], Line)
               end,
               UnexportedUsedTypes).
+
+get_behaviour_callbacks(Root) ->
+    IsBehaviour = fun(Node) -> ktn_code:type(Node) == behaviour end,
+    Behaviours = elvis_code:find(IsBehaviour, Root),
+    BehaviourNames = lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
+
+    lists:append(
+      lists:map(fun(B) -> B:behaviour_info(callbacks) end, BehaviourNames
+    )).
 
 get_type_of_type(#{type := type_attr,
                    node_attrs := #{type := #{attrs := #{name := TypeOfType}}}}) ->
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1504,7 +1504,7 @@ get_behaviour_callbacks(Root) ->
         DeepList ->
             lists:append(DeepList)
     catch
-        _ ->
+        _:_ ->
             []
     end.
 
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1497,11 +1497,11 @@ export_used_types(Config, Target, RuleConfig) ->
 get_behaviour_callbacks(Root) ->
     IsBehaviour = fun(Node) -> ktn_code:type(Node) == behaviour end,
     Behaviours = elvis_code:find(IsBehaviour, Root),
-    BehaviourNames = lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
+    BehaviourNames =
+        lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
 
     lists:append(
-      lists:map(fun(B) -> B:behaviour_info(callbacks) end, BehaviourNames
-    )).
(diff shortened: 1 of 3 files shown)


src/elvis_style.erl
```diff
@@ -1492,6 +1494,15 @@ export_used_types(Config, Target, RuleConfig) ->
               end,
               UnexportedUsedTypes).
 
+get_behaviour_callbacks(Root) ->
+    IsBehaviour = fun(Node) -> ktn_code:type(Node) == behaviour end,
+    Behaviours = elvis_code:find(IsBehaviour, Root),
+    BehaviourNames = lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
+
+    lists:append(
+      lists:map(fun(B) -> B:behaviour_info(callbacks) end, BehaviourNames
```
bormilan, line 1503: Elvis warns me on this line to "remove the dynamic function call", but I don't really know how to make this not a dynamic way.

src/elvis_style.erl
```diff
@@ -1492,6 +1494,15 @@ export_used_types(Config, Target, RuleConfig) ->
...
               UnexportedUsedTypes).
 
+get_behaviour_callbacks(Root) ->
+    IsBehaviour = fun(Node) -> ktn_code:type(Node) == behaviour end,
+    Behaviours = elvis_code:find(IsBehaviour, Root),
+    BehaviourNames =
+        lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
+
+    lists:append(
+        lists:map(fun(B) -> apply(B, behaviour_info, [callbacks]) end, BehaviourNames)).
```
elbrujohalcon, line 1504: A `try…catch` is needed here, just in case someone introduces a typo (`-behaviour(gen_srvr).`) or uses a behaviour that's defined outside of OTP (`-behaviour(my_behaviour).`).

elbrujohalcon (changes requested):
I'm not sure about this _fix_.
This will still fail on the following scenarios:
* On `gen_statem`, `ct_suite`, and other behaviours with dynamic callbacks, until https://github.com/erlang/otp/issues/4847 is taken care of.
* On custom behaviours, defined by the user in the same app that we're analyzing.

I'm not rejecting the PR entirely, but I'll leave it in the hands of @paulo-ferraz-oliveira to decide if this improving is worth merging or not.

bormilan:
You are absolutely right. It was my best shot to solve this issue, until the issue erlang/otp#4847 will be solved.

I'm curious what do you think about it, or do you think that we should wait for the mentioned otp request to be solved?

elbrujohalcon:
> You are absolutely right. It was my best shot to solve this issue, until the issue [erlang/otp#4847](https://github.com/erlang/otp/issues/4847) will be solved.
> 
> I'm curious what do you think about it, or do you think that we should wait for the mentioned otp change request to be solved?

I think we should wait… but let's hear what @paulo-ferraz-oliveira has to say.

paulo-ferraz-oliveira:
I wouldn't mind waiting for the OTP changes, but these can take a while to happen...

paulo-ferraz-oliveira:
@elbrujohalcon, otoh, what would prevent us from implementing an incomplete solution (I guess it's what this PR is doing), for the time being, that didn't take dynamic callbacks into account? We could add this caveat to the documentation, and create a task to deal with the rest when the OTP issue is solved.

elbrujohalcon:
@paulo-ferraz-oliveira it has to be super-clear that this is a partial solution (in the docs and everywhere else), so that we don't get complaints about incorrect or missing warnings.
I'm not saying that we shouldn't do it. In my mind it's not worthy, but if you think it's useful, as long as you don't create false hopes, go for it!

paulo-ferraz-oliveira:
We could make it as an option, then? I don't mind not having it, but I think the benefits outweigh the disadvantages. And how to you foresee complaints? This will warn less, not more, right?

elbrujohalcon:
@paulo-ferraz-oliveira either…
- We only avoid the warning if we know the behaviour (i.e., we have a list of known behaviours) and somebody creates a behaviour _on top of_ `gen_server` and gets the warning on missing exported types. Or…
- We just avoid the warning for _any_ behaviour (i.e., if the module defines a behaviour any exported function _can be_ a callback) and someone doesn't get the warning that they're expecting the rule to emit.

It's not a big deal, but yeah… keep it as an option, that's a good idea. Be clear in the docs about the option that it's _heuristically implemented_ and it doesn't cover all the cases (or if you go the other way… it covers too many cases) and it would be fine with me.

src/elvis_style.erl
```diff
@@ -1492,6 +1494,20 @@ export_used_types(Config, Target, RuleConfig) ->
...
+    IsBehaviour = fun(Node) -> ktn_code:type(Node) == behaviour end,
+    Behaviours = elvis_code:find(IsBehaviour, Root),
+    BehaviourNames =
+        lists:map(fun(#{attrs := #{value := Behaviour}}) -> Behaviour end, Behaviours),
+
+    try lists:map(fun(B) -> apply(B, behaviour_info, [callbacks]) end, BehaviourNames) of
+        DeepList ->
+            lists:append(DeepList)
+    catch
+        _ ->
```
elbrujohalcon, line 1507:
```suggestion
        _:_ ->
```
You need to catch errors, too… not just throws :)

elbrujohalcon (changes requested):
The way in which this is implemented now will only ignore _non-dynamic_ callbacks from modules implementing behaviours that are already known at compile time (i.e. OTP behaviours without dynamic callbacks).
@paulo-ferraz-oliveira / @bormilan: You should decide if this is what you want.
Alternatively, you can just ignore the module entirely if it implements an unknown behaviour, or a behaviour that we know that it has dynamic callbacks (e.g. `gen_statem`).

elbrujohalcon (changes requested):
It's good, @bormilan … but please add:
* A note to the rule docs explaining that the rule will ignore regular (i.e. not dynamic) callbacks of OTP behaviours (And maybe other behaviours if their `behaviour_info` is _accesible_ at compile time).
* A module implementing `gen_statem` that fails the rule because a non-exported type is used in the spec of _dynamic_ callback.
* A module implementing a custom behaviour that fails the rule because a non-exported type is used in a regular callback.

bormilan:
I'm not sure if this implementation is worth pursuing. The concerns you've raised are entirely valid, and if we need to discuss an issue for this long, it may be a sign that we should reconsider proceeding.

Regarding the suggestion to "ignore the module entirely if it implements unknown behavior or behavior with dynamic callbacks," I believe that adding this could be beneficial, but it's only part of the overall goal.

In conclusion, I am uncertain about my feelings on this matter. Perhaps it is a sign to wait before moving forward. I began working on this task because I am focused on the 4.0.0 milestone, not out of personal interest.

If we decide to hold off, we can move this issue to a different milestone.

elbrujohalcon:
@bormilan I changed the milestone in the issue to "eventually"… and took this chance to reorganize tickets a bit… so I added a few more to the 4.0.0 milestone. All of them qualify as _low-hanging fruit_ in principle.

bormilan:
Thank you! I will do this too, but I feel it's better to wait.

PR #372, opened 2024-11-10
fix(no_successive_maps): prevent failing on nested maps

### LINKED ISSUE #339: no_successive_maps doesn't work when nested
## Bug Description

The `no_successive_maps` rule (see #157 and #255) doesn't warn when using nested maps.

## To Reproduce

This triggers the warning:

```
_ = [
    #{a => b, c => d},
    #{a => b, c => d} % elvis: Found map update after map construction/update
    #{a => b, c => d},
    #{a => b, c => d}
],
```

This doesn't:

```
_ = #{
    x => [
        #{a => b, c => d},
        #{a => b, c => d} % expected a warning here; didn't get one
        #{a => b, c => d},
        #{a => b, c => d}
    ]
},
```

## Additional Context

- OS: macOS
- Erlang/OTP 26.2.1
- erlang.mk; see below

```
BUILD_DEPS = elvis_mk
DEP_PLUGINS = elvis_mk
ELVIS_VERSION = 3.0.1
dep_elvis_mk = git https://github.com/inaka/elvis.mk.git
```

bormilan:
# Description

I just added the option: `traverse => all` to the rule because when I investigated the problem with the provided example, I realized that the `elvis_code:find` does not go deep into this nested map, because the zippers can go deeper only if a `node` is a `branch`.  A node is identified as a `branch` if it has something in its `content` field(if the traverse is set to `content`). So we need to change `traverse` to `all` so the `elvis_code` can use the `node_attrs` as the children of the node.

In this example, the rule fails because a `map_field_assoc` type node contains no content, preventing the function `elvis_code:find` from going deeper; it only returns that one map.

Closes #339.

### DIFF
--- a/test/examples/fail_no_successive_maps2.erl
+++ b/test/examples/fail_no_successive_maps2.erl
@@ -1,0 +1,31 @@
+-module(fail_no_successive_maps2).
+
+-export([test1/0, test2/0]).
+
+test1() ->
+  _ = #{
+    x => [
+        #{a => b, c => d},
+        #{a => b, c => d} % expected a warning here
+        #{a => b, c => d},
+        #{a => b, c => d}
+    ]
+}.
+
+test2() ->
+  _ = #{
+    map => #{
+          x => [
+              #{a => b, c => d},
+              #{a => b, c => d} % expected a warning here
+              #{a => b, c => d},
+              #{a => b, c => d}
+          ],
+          y => [
+              #{a => b, c => d},
+              #{a => b, c => d} % expected a warning here
+              #{a => b, c => d},
+              #{a => b, c => d}
+          ]
+      }
+}.
--- a/test/examples/fail_no_successive_maps2.erl
+++ b/test/examples/fail_no_successive_maps2.erl
@@ -2,6 +2,7 @@
 
 -export([test1/0, test2/0]).
 
+-if(?OTP_RELEASE<27).
 test1() ->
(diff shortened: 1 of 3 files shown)


paulo-ferraz-oliveira (approved):
Looks Ok to me. Thanks.

test/examples/fail_no_successive_maps2.erl
```diff
@@ -0,0 +1,41 @@
...
+    x => [
+        #{a => b, c => d},
+        #{a => b, c => d} % expected a warning here
+        #{a => b, c => d},
+        #{a => b, c => d}
+    ]
+}.
+-else.
+test1() ->
+    #{}.
```
elbrujohalcon, line 17: How does this fail in OTP27? 🤔 
The other branch of the `-if(…). … -else. …` does emit a warning.

elbrujohalcon (changes requested):
I don't understand how you made tests pass on OTP27, but the rest of the PR is cool.

elbrujohalcon (approved):

test/examples/fail_no_successive_maps2.erl
elbrujohalcon: Oh, I see… elvis ignores the macros… HAHAHA!
It's a clever trick… but maybe in the future, we should add a comment so it doesn't surprise future us :P

PR #373, opened 2024-11-10
Fix `protocol_for_deps` allow for `git_subdir`'s `ref` and `tag`

### LINKED ISSUE #370: `rebar3` dependency "git_subdir" tag not supported for `tag` (rule: `protocol_for_deps`)
## Bug Description

#343 is not complete, in that it doesn't support `tag` where it supports `branch`.

## To Reproduce

Take the test file on that pull request and replace `branch` with `tag`, [here](https://github.com/inaka/elvis_core/pull/343/files#diff-aced31f01c95c48279adfa9de16d8c0f1a43f52f6d5e30ebe37caa3a9ad24158R9).

## Expected Behavior

No function clause.

## `rebar3` Logs

Not applicable.

## Additional Context

- OS: all
- Erlang version: all
- rebar3 version: all

paulo-ferraz-oliveira:
# Description

We:

* improve matching on `branch` for `git_subdir`
* identify `raw` resources with comments (so future-us knows where those references come from)
* allow for a new (though only for backwards compatibility) "with opts" match
  * it's unlikely it's commonly used, but since `rebar3` supports it as legacy, I believe we should too.

Closes #370.


test/examples/rebar3_2.config.success
paulo-ferraz-oliveira, line 1: Added this one (almost a copy from the sample in `rebar3`) so I didn't have to adapt the previous tests/results, that way kinda proving that the previous input is still Ok to use.

test/examples/rebar3_2.config.success
```diff
@@ -0,0 +1,19 @@
...
+    {rebar, "1.0.0"},
+    {rebar, {pkg, rebar_fork}},
+    {rebar, "1.0.0", {pkg, rebar_fork}},
+    {rebar, {git, "https://github.com/rebar/rebar.git", {branch, "main"}}},
+    {rebar, {git, "https://github.com/rebar/rebar.git", {tag, "1.0.0"}}},
+    {rebar, {git, "https://github.com/rebar/rebar.git", {ref, "7f73b8d6"}}},
+    {rebar, {hg, "https://github.com/rebar/rebar.git", {tag, "1.0.0"}}},
+    {rebar, {git_subdir, "https://github.com/rebar/rebar.git", {branch, "main"}, "subdir"}},
+    {rebar, {git_subdir, "https://github.com/rebar/rebar.git", {tag, "1.0.0"}, "sub/dir"}},
+    {rebar, {git_subdir, "https://github.com/rebar/rebar.git", {ref, "7f73b8d6"}, "dir"}},
```
paulo-ferraz-oliveira, line 13: If `git://` becomes part of the default regex we could tweak some of these, as per `rebar3`'s sample config.

src/elvis_project.erl
```diff
@@ -124,25 +127,37 @@ is_hex_dep(_AppName) when is_atom(_AppName) ->
 is_hex_dep({_AppName, _Vsn, {pkg, _PackageName}})
     when is_atom(_AppName), is_list(_Vsn), is_atom(_PackageName) ->
     true;
+is_hex_dep({_AppName, {pkg, _OtherName}}) when is_atom(_AppName), is_atom(_OtherName) ->
+    true;
```
paulo-ferraz-oliveira, line 131: Moved from below, not new...

src/elvis_project.erl
```diff
@@ -124,25 +127,37 @@ is_hex_dep(_AppName) when is_atom(_AppName) ->
...
 is_hex_dep({_AppName, _Vsn}) when is_atom(_AppName), is_list(_Vsn) ->
     true;
 is_hex_dep(_) ->
     false.
 
 %% @private
-is_not_git_dep({_AppName, {pkg, _OtherName}}, _Regex) ->
-    false;
```
paulo-ferraz-oliveira, line 134: Moved above, not actually removed...

src/elvis_project.erl
```diff
@@ -124,25 +127,37 @@ is_hex_dep(_AppName) when is_atom(_AppName) ->
...
     is_not_git_dep({AppName, DepResourceSpecification}, Regex);
+%% Alternative formats, backwards compatible declarations
 is_not_git_dep({_AppName, {_SCM, Url}}, Regex) ->
     nomatch == re:run(Url, Regex, []);
 is_not_git_dep({_AppName, _Vsn, {_SCM, Url}}, Regex) ->
     nomatch == re:run(Url, Regex, []);
 is_not_git_dep({_AppName, _Vsn, {_SCM, Url, _Branch}}, Regex) ->
     nomatch == re:run(Url, Regex, []);
-is_not_git_dep({_AppName, {git_subdir, Url, {branch, _Branch}, _SubDir}}, Regex) ->
+is_not_git_dep({_AppName, _Vsn, {_SCM, Url, {BranchTagOrRefType, _Branch}}, _Opts}, Regex)
```
paulo-ferraz-oliveira, line 157: Notice the `branch` being matched against `tag` and `ref` now, too...

src/elvis_project.erl
```diff
@@ -124,25 +127,37 @@ is_hex_dep(_AppName) when is_atom(_AppName) ->
...
-    false;
 is_not_git_dep({_AppName, {_SCM, Url, _Branch}}, Regex) ->
     nomatch == re:run(Url, Regex, []);
+is_not_git_dep({_AppName,
+                {git_subdir, Url, {BranchTagOrRefType, _BranchTagOrRef}, _SubDir}},
+               Regex)
+    when BranchTagOrRefType =:= branch;
+         BranchTagOrRefType =:= tag;
+         BranchTagOrRefType =:= ref ->
+    nomatch == re:run(Url, Regex, []);
```
paulo-ferraz-oliveira, line 146: Notice the `branch` being matched against `tag` and `ref` now, too...

test/project_SUITE.erl
```diff
@@ -41,16 +41,16 @@ verify_no_branch_deps(_Config) ->
     Filename = "rebar.config.fail",
     {ok, File} = elvis_test_utils:find_file(SrcDirs, Filename),
 
-    [_, _, _, _] = elvis_project:no_branch_deps(ElvisConfig, File, #{}),
+    [_, _, _, _, _] = elvis_project:no_branch_deps(ElvisConfig, File, #{}),
```
paulo-ferraz-oliveira, line 44: All of these additions show we were already not matching against something we should be.

test/examples/rebar3_2.config.success
elbrujohalcon: I think it should.

elbrujohalcon (approved):
Looks good to me, although I would add support for `git://`, as well.

paulo-ferraz-oliveira:
I'm spelunking into the code: `[0-9]+([.][0-9]+)*)` was introduced in the regex in 0.2.9, but I'm still trying to find why (or what for).

test/examples/rebar.config.fail
paulo-ferraz-oliveira, line 1: Changes allow keeping breakage, as-was.

test/examples/rebar3_2.config.success
paulo-ferraz-oliveira, line 1: Just a few examples to make sure we test it.

doc_rules/elvis_project/protocol_for_deps.md
```diff
@@ -8,7 +8,7 @@ This rule was called `protocol_for_deps_rebar` before
 ## Options
 
 - `regex :: string()`.
-  - default: `(https://.*|[0-9]+([.][0-9]+)*)`.
+  - default: `^(https://|git://|\\d+(\\.\\d+)*)`.
```
paulo-ferraz-oliveira, line 11: Calling your attention to this:

1. it should start with `https`, `git` or `\d` and not just contain these (as it was what it was doing - so `git2://` wouldn't fail, for example)
2. we still support the "Erlang.mk" standard; I don't know what that is, but at least we don't break it much (?)

elbrujohalcon (approved):

PR #374, opened 2024-11-15
Ets fun2ms without ms transform

### LINKED ISSUE #313: Using ets:fun2ms without including ms_transform.hrl should be an error
Using ets:fun2ms without including ms_transform.hrl raises a runtime error, rather than a compile error:

> The parse transform is provided in the ms_transform module and the source must include file ms_transform.hrl in STDLIB for this pseudo function to work. Failing to include the hrl file in the source results in a runtime error, not a compile time error. The include file is easiest included by adding line -include_lib("stdlib/include/ms_transform.hrl"). to the source file.

If elvis could spot this, and raise an error, that'd save some headaches. I understand that it probably can't be exhaustive, but it'd be nice.

bormilan:
# Description

It detects if a module that does not include `ms_transform.hrl` has any `ets:fun2ms` function call.

Closes #313;


src/elvis_style.erl
```diff
@@ -33,6 +33,8 @@
 
 -define(NO_INIT_LISTS_MSG,
         "Do not use a list as the parameter for the 'init' callback at position ~p.").
+-define(MS_TRANSFORM_INCLUDED_MSG,
+        "Missing inclide library: stdlib/include/ms_transform.hrl at position ~p.").
```
elbrujohalcon, line 37:
```suggestion
        "Missing inclide library: stdlib/include/ms_transform.hrl when ets:fun2ms/1 is used at position ~p.").
```

elbrujohalcon (changes requested):
For performance (and semantics) sake, I would do the check in the opposite direction. In other words, instead of first looking for all files that don't include ms_transform and checking each one to see if they use ets:fun2ms… I would look for all files that use ets:fun2ms (which are surely far less than the ones that don't include ms_transform) and then check if they include the hrl or not.

You should also add a test with a file that has the `-include` but not the usage of `fun2ms`, for completeness sake. Like…

```erlang
-module(pass_ms_transform_included_too).

-include_lib("stdlib/include/ms_transform.hrl").

-export([test/0]).

test() -> ok.
```

bormilan:
Oh yes, that is a way better approach. Thanks!

bormilan:
I made some changes based on your suggestions. However, I believe this rule could be more generic. Currently, it only checks for a specific scenario, but we can expand it. 

We could create a rule that detects any function call that requires an include but does not have it. We can add this scenario to the default configuration, and users can also add their own pairs of `{function_call, lib_file_name}` if they wish.

doc_rules/elvis_style/ms_transform_included.md
```diff
@@ -0,0 +1,16 @@
+
+# No Init Lists
```
elbrujohalcon, line 2:
```suggestion
# Include `ms_transform` for `ets:fun2ms/1`
```

src/elvis_style.erl
```diff
@@ -1192,6 +1196,76 @@ is_list_node(#{type := match, content := Content}) ->
...
+-spec get_fun_2_ms_calls(ktn_code:tree_node()) -> [any()].
+get_fun_2_ms_calls(Root) ->
+    IsFun2MsFunctionCall =
+        fun(Node) ->
+           case ktn_code:type(Node) == call of
+               true ->
+                   is_ets_fun2ms(Node);
+               false ->
+                   false
+           end
```
elbrujohalcon, line 1239:
```suggestion
           case ktn_code:type(Node) of
               call ->
                   is_ets_fun2ms(Node);
               _ ->
                   false
           end
```
Or…
```suggestion
           ktn_code:type(Node) == call andalso is_ets_fun2ms(Node)
```

src/elvis_style.erl
```diff
@@ -1192,6 +1196,76 @@ is_list_node(#{type := match, content := Content}) ->
...
+
+    lists:map(ProcessResult, Functions).
+
+-spec is_ets_fun2ms(ktn_code:tree_node()) -> boolean().
+is_ets_fun2ms(Node) ->
+    Fun = ktn_code:node_attr(function, Node),
+    Fun2 = ktn_code:node_attr(function, Fun),
+    Module = ktn_code:node_attr(module, Fun),
+
+    {ets, fun2ms} == {ktn_code:attr(value, Module), ktn_code:attr(value, Fun2)}.
```
elbrujohalcon, line 1253:
```suggestion
    ets == ktn_code:attr(value, Module) andalso fun2ms == ktn_code:attr(value, Fun2).
```
To avoid computing a thing if the other is not the one we're expecting it to be (i.e., short-circuiting).

src/elvis_style.erl
```diff
@@ -1192,6 +1196,76 @@ is_list_node(#{type := match, content := Content}) ->
...
+             ktn_code:type(Node) == include_lib
+             andalso ktn_code:attr(value, Node) == "stdlib/include/ms_transform.hrl"
+          end,
+
+    case elvis_code:find(Fun, Root) of
+        [_] ->
+            true;
+        _ ->
+            false
+    end.
```
elbrujohalcon, line 1267:
```suggestion
    [] /= elvis_code:find(Fun, Root).
```

At least in theory… you can include that file twice. I'm not sure if that code would compile but, as long as it's included **at least once**, it should be fine for our rule.

elbrujohalcon: Maybe try adding a test like…
```erlang
-module(pass_ms_transform_included_twice).

-include_lib("stdlib/include/ms_transform.hrl").
-include_lib("stdlib/include/ms_transform.hrl").

-export([test/0]).

test() -> ets:fun2ms(fun(_) -> ok end).
```

elbrujohalcon (changes requested):
A few tiny improvements in code and docs… and do consider adding the tests I requested in my previous review, @bormilan.
Thanks.

bormilan:
Okay, it sounds

> A few tiny improvements in code and docs… and do consider adding the tests I requested in my previous review, @bormilan. Thanks.

Yes, I will add that test, I didn't forget.

src/elvis_style.erl
bormilan: Oh, its smart, thanks!

src/elvis_style.erl
bormilan: Yes, I was confused about a lot of cases 😅
Your second one is so clean, thanks!

src/elvis_style.erl
bormilan: Oh yeah, this simple assertion is much better.
I will add this test too, thanks!

src/elvis_style.erl
```diff
@@ -1192,6 +1196,64 @@ is_list_node(#{type := match, content := Content}) ->
...
+
+    FunctionCalls = get_fun_2_ms_calls(Root),
+
+    IsIncluded =
+        case length(FunctionCalls) > 0 of
+            true ->
+                is_include_ms_transform(Root);
+            false ->
+                false
+        end,
```
elbrujohalcon, line 1214:
```suggestion
    IsIncluded = FunctionCalls /= [] andalso is_include_ms_transform(Root),
```

src/elvis_style.erl
```diff
@@ -1192,6 +1196,64 @@ is_list_node(#{type := match, content := Content}) ->
...
+-spec is_ets_fun2ms(ktn_code:tree_node()) -> boolean().
+is_ets_fun2ms(Node) ->
+    Fun = ktn_code:node_attr(function, Node),
+    Fun2 = ktn_code:node_attr(function, Fun),
+    Module = ktn_code:node_attr(module, Fun),
+
+    ets == ktn_code:attr(value, Module) andalso fun2ms == ktn_code:attr(value, Fun2).
+
+-spec is_include_ms_transform(ktn_code:tree_node()) -> boolean().
+is_include_ms_transform(Root) ->
```
elbrujohalcon, line 1249: I would call this function `has_include_for_ms_transform/1`, because we're looking for an include, we're not checking if the node _is_ the include itself.

elbrujohalcon (approved):
A little bit of nit-picking, but overall the code is great.
If you feel like ignoring the nit-picky comments, @bormilan … just let me know and I'll merge the PR as-is.

bormilan:
> If you feel like ignoring the nit-picky comments, @bormilan … just let me know and I'll merge the PR as-is.

No, I'm grateful for your time, and I want to bring home as much as I can from these, so please be as nit-picky as you want, I really appreciate it! 🙌

Some say, "God is in the details."
I want to appreciate this project with the best work I can do for the opportunities it provides me on my journey.

test/examples/used_ignored_variable_in_macro.erl
```diff
@@ -0,0 +1,13 @@
+-module(used_ignored_variable_in_macro).
```
elbrujohalcon, line 1: I don't think this file belongs to this PR.

elbrujohalcon (changes requested):

elbrujohalcon (approved):

src/elvis_rulesets.erl
```diff
@@ -113,6 +114,7 @@ rules(beam_files) ->
                invalid_dynamic_call,
                max_anonymous_function_arity,
                max_function_arity,
+               ms_transform_included,
```
paulo-ferraz-oliveira, line 117: @elbrujohalcon, I wonder if, in the future, to ease maintenance, this list should be build like a sub-set of `_strict`, for example...

PR #377, opened 2024-12-01
add forbidden regex

### LINKED ISSUE #151: Extend naming conventions with things to avoid
Currently, our naming convention rules only allow for a single regex that identifies what is _permitted_/_valid_.
It would be nice to add an additional _forbidden_ regex to identify things that are not valid, like function names starting with `do_`.

Ref: WhatsApp/erlfmt#91

bormilan:
# Description

I added the `forbidden_regex' to all "naming_convention" rules.
I made tests for a simple case about making numbers forbidden.

I can't decide in these situations that making a "case in case" is a good implementation. I feel that moving these into another function is not much cleaner, and maybe reading is more frustrating if you need to look into another function just for another case. What do you think about it?

Closes #151.

### DIFF
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -6,7 +6,8 @@ All atoms should be named according to the regular expression provided.
 Except if it matches with a defined `forbidden_regex`.
 Atoms enclosed in apostrophes have special meaning and are thus handled
 by a different configuration option (use `same` if you want the same value as `regex`).
-To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`), use `forbidden_enclosed_regex`(use `same` if you want the same value as `forbidden_regex`).
+To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`),
+use `forbidden_enclosed_regex`(use `same` if you want the same value as `forbidden_regex`).
 
 > Works on `.beam` file? Yes!
 
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -18,7 +18,7 @@ To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for
   - default: `".*"`.
 - `forbidden_regex :: string() | undefined`.
   - default: `undefined`.
-- `forbidden_enclosed_regex :: string() | undefined | same`.
+- `forbidden_enclosed_regex :: string() | undefined`.
   - default: `undefined`.
 
 ## Example
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -6,6 +6,7 @@ All atoms should be named according to the regular expression provided.
 Except if it matches with a defined `forbidden_regex`.
 Atoms enclosed in apostrophes have special meaning and are thus handled
 by a different configuration option (use `same` if you want the same value as `regex`).
+For enclosed atoms we defined a different forbidden regex, called `forbidden_enclosed_regex`.
 
 > Works on `.beam` file? Yes!
 
@@ -16,6 +17,8 @@ by a different configuration option (use `same` if you want the same value as `r
 - `enclosed_atoms :: string()`.
   - default: `".*"`.
 - `forbidden_regex :: string() | undefined`.
+  - default: `undefined`.
+- `forbidden_enclosed_regex :: string() | undefined`.
   - default: `undefined`.
 
 ## Example
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -6,7 +6,7 @@ All atoms should be named according to the regular expression provided.
 Except if it matches with a defined `forbidden_regex`.
 Atoms enclosed in apostrophes have special meaning and are thus handled
 by a different configuration option (use `same` if you want the same value as `regex`).
-For enclosed atoms we defined a different forbidden regex, called `forbidden_enclosed_regex`.
+To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`), use `forbidden_enclosed_regex`.
 
 > Works on `.beam` file? Yes!
 
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -2,8 +2,8 @@
 
 (since [1.0.0](https://github.com/inaka/elvis_core/releases/tag/1.0.0))
 
-All atoms should be named according to the regular expression provided.
-Except if it matches with a defined `forbidden_regex`.
+All atoms should be named according to the provided regular expression,
+except if they match with a defined `forbidden_regex`.
 Atoms enclosed in apostrophes have special meaning and are thus handled
 by a different configuration option (use `same` if you want the same value as `regex`).
 To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`),
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -15,11 +15,11 @@ use `forbidden_enclosed_regex`(use `same` if you want the same value as `forbidd
 
 - `regex :: string()`.
   - default: `"^([a-z][a-z0-9]*_?)*(_SUITE)?$"`.
-- `enclosed_atoms :: string()`.
+- `enclosed_atoms :: string() | same`.
   - default: `".*"`.
 - `forbidden_regex :: string() | undefined`.
   - default: `undefined`.
-- `forbidden_enclosed_regex :: string() | undefined`.
+- `forbidden_enclosed_regex :: string() | undefined | same`.
   - default: `undefined`.
 
 ## Example
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -4,8 +4,8 @@
 
 All atoms should be named according to the regular expression provided.
 Except if it matches with a defined `forbidden_regex`.
-Atoms enclosed in apostrophes have special meaning and are thus handled by a different configuration option (use
-`same` if you want the same value as `regex`).
+Atoms enclosed in apostrophes have special meaning and are thus handled
+by a different configuration option (use `same` if you want the same value as `regex`).
 
 > Works on `.beam` file? Yes!
 
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -6,7 +6,7 @@ All atoms should be named according to the regular expression provided.
 Except if it matches with a defined `forbidden_regex`.
 Atoms enclosed in apostrophes have special meaning and are thus handled
 by a different configuration option (use `same` if you want the same value as `regex`).
-To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`), use `forbidden_enclosed_regex`.
+To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`), use `forbidden_enclosed_regex`(use `same` if you want the same value as `forbidden_regex`).
 
 > Works on `.beam` file? Yes!
 
@@ -18,7 +18,7 @@ To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for
   - default: `".*"`.
 - `forbidden_regex :: string() | undefined`.
   - default: `undefined`.
-- `forbidden_enclosed_regex :: string() | undefined`.
+- `forbidden_enclosed_regex :: string() | undefined | same`.
   - default: `undefined`.
 
 ## Example
--- a/doc_rules/elvis_style/atom_naming_convention.md
+++ b/doc_rules/elvis_style/atom_naming_convention.md
@@ -2,8 +2,9 @@
 
 (since [1.0.0](https://github.com/inaka/elvis_core/releases/tag/1.0.0))
 
-All atoms should be named according to the regular expression provided. Atoms enclosed in
-apostrophes have special meaning and are thus handled by a different configuration option (use
+All atoms should be named according to the regular expression provided.
+Except if it matches with a defined `forbidden_regex`.
+Atoms enclosed in apostrophes have special meaning and are thus handled by a different configuration option (use
 `same` if you want the same value as `regex`).
 
 > Works on `.beam` file? Yes!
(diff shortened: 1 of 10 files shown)


src/elvis_style.erl
```diff
@@ -72,14 +72,23 @@
 -define(FUNCTION_NAMING_CONVENTION_MSG,
         "The function ~p does not respect the format defined by the "
         "regular expression '~p'.").
+-define(FORBIDDEN_FUNCTION_NAMING_CONVENTION_MSG,
+        "The function ~p does written in a forbidden format"
```
elbrujohalcon, line 76: _is written_, not _does written_.
And the same applies to the other texts.

elbrujohalcon (changes requested):

doc_rules/elvis_style/atom_naming_convention.md
```diff
@@ -2,9 +2,10 @@
 
 (since [1.0.0](https://github.com/inaka/elvis_core/releases/tag/1.0.0))
 
-All atoms should be named according to the regular expression provided. Atoms enclosed in
-apostrophes have special meaning and are thus handled by a different configuration option (use
-`same` if you want the same value as `regex`).
+All atoms should be named according to the regular expression provided.
+Except if it matches with a defined `forbidden_regex`.
+Atoms enclosed in apostrophes have special meaning and are thus handled
+by a different configuration option (use `same` if you want the same value as `regex`).
```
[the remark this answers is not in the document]
bormilan: I don't really understand this, what do you mean?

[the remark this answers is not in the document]
elbrujohalcon: we have a regex for accepted atom names (i.e., `regex`), but we also have a regex for enclosed atoms (i.e., `enclosed_atoms`). Now we added `forbidden_regex` that limits what's accepted by `regex`. Should it also limit what's accepted by `enclosed_atoms` or not?
I would say "no" and I would add another property, called `forbidden_enclosed_atoms` to deal with what's forbidden in enclosed atoms.

doc_rules/elvis_style/atom_naming_convention.md
```diff
@@ -6,6 +6,7 @@ All atoms should be named according to the regular expression provided.
 Except if it matches with a defined `forbidden_regex`.
 Atoms enclosed in apostrophes have special meaning and are thus handled
 by a different configuration option (use `same` if you want the same value as `regex`).
+For enclosed atoms we defined a different forbidden regex, called `forbidden_enclosed_regex`.
```
elbrujohalcon, line 9:
```suggestion
To define forbidden enclosed atoms (like the ones in `forbidden_regex` apply for `regex`), use `forbidden_enclosed_regex`.
```

doc_rules/elvis_style/atom_naming_convention.md
```diff
@@ -17,6 +18,8 @@ by a different configuration option (use `same` if you want the same value as `r
   - default: `".*"`.
 - `forbidden_regex :: string() | undefined`.
   - default: `undefined`.
+- `forbidden_enclosed_regex :: string() | undefined`.
```
elbrujohalcon, line 21: According to the description above this field **and also `enclosed_atoms`** should also support the atom `same` and then use the same value as the corresponding parameter for non-enclosed atoms.
If that works, we should have tests for it.

src/elvis_style.erl
```diff
@@ -1162,8 +1163,16 @@ atom_naming_convention(Config, Target, RuleConfig) ->
     ForbiddenRegex = option(forbidden_regex, RuleConfig, atom_naming_convention),
     RegexEnclosed =
         specific_or_default(option(enclosed_atoms, RuleConfig, atom_naming_convention), Regex),
+    ForbiddenEnclosedRegex =
+        specific_or_default(option(forbidden_enclosed_regex, RuleConfig, atom_naming_convention),
+                            Regex),
```
elbrujohalcon, line 1168:
```suggestion
                            ForbiddenRegex),
```
Careful… if `Regex` is the default for `RegexEnclosed`, `ForbiddenRegex` should be the default for `ForbiddenEnclosedRegex`. Since the tests didn't fail with `Regex` here, that means we need more tests to prove that this works as expected.

elbrujohalcon (changes requested):

src/elvis_style.erl
bormilan: That's right; I was inattentive about this one.

doc_rules/elvis_style/atom_naming_convention.md
bormilan: yes, I fix it.

elbrujohalcon: I mean… it should say `string() | undefined | same` here and there.

src/elvis_style.erl
```diff
@@ -292,22 +307,35 @@ default(RuleWithEmptyDefault)
...
+                {match, _} ->
+                    Msg = ?FORBIDDEN_FUNCTION_NAMING_CONVENTION_MSG,
+                    Info = [FunctionNameStr, Regex],
+                    Result = elvis_result:new(item, Msg, Info, 1),
+                    [Result | errors_for_function_names(Regex, ForbiddenRegex, RemainingFuncNames)];
+                nomatch ->
+                    errors_for_function_names(Regex, ForbiddenRegex, RemainingFuncNames);
+                false ->
+                    errors_for_function_names(Regex, ForbiddenRegex, RemainingFuncNames)
+            end
```
elbrujohalcon, line 338:
```suggestion
            case ForbiddenRegex of
                undefined -> errors_for_function_names(Regex, ForbiddenRegex, RemainingFuncNames);
                ForbiddenRegex ->
                    case re:run(FunctionNameStr, ForbiddenRegex, [unicode]) of
                {match, _} ->
                    Msg = ?FORBIDDEN_FUNCTION_NAMING_CONVENTION_MSG,
                    Info = [FunctionNameStr, Regex],
                    Result = elvis_result:new(item, Msg, Info, 1),
                    [Result | errors_for_function_names(Regex, ForbiddenRegex, RemainingFuncNames)];
                nomatch ->
                    errors_for_function_names(Regex, ForbiddenRegex, RemainingFuncNames)
                 end
            end
```

I see what you did there… and… it's clever, but we don't abuse `andalso` in this project (¬_¬)

I think.

src/elvis_style.erl
```diff
@@ -739,7 +769,19 @@ module_naming_convention(Config, Target, RuleConfig) ->
                     Result = elvis_result:new(item, Msg, Info, 1),
                     [Result];
                 {match, _} ->
-                    []
+                    case ForbiddenRegex =/= undefined
```
elbrujohalcon, line 772: Same as above, please don't abuse the poor short-circuit operator.

src/elvis_style.erl
```diff
@@ -1686,9 +1749,27 @@ check_atom_names(Regex, RegexEnclosed, [AtomNode | RemainingAtomNodes], AccIn) -
                 Result = elvis_result:new(item, Msg, Info, Line),
                 AccIn ++ [Result];
             {match, _Captured} ->
-                AccIn
+                case ForbiddenRegex =/= undefined
```
elbrujohalcon, line 1752: Same as all the other places before.

src/elvis_style.erl
```diff
@@ -1714,21 +1795,33 @@ is_atom_node(MaybeAtom) ->
...
         nomatch ->
             Msg = ?VARIABLE_NAMING_CONVENTION_MSG,
             {Line, _} = ktn_code:attr(location, Variable),
             Info = [VariableNameStr, Line, Regex],
             Result = elvis_result:new(item, Msg, Info, Line),
-            [Result | check_variables_name(Regex, RemainingVars)];
+            [Result | check_variables_name(Regex, ForbiddenRegex, RemainingVars)];
         {match, _} ->
-            check_variables_name(Regex, RemainingVars)
+            case ForbiddenRegex =/= undefined
```
elbrujohalcon, line 1812: …aaaand another one.

elbrujohalcon (changes requested):

elbrujohalcon (approved):

elbrujohalcon:
> But what's the problem with `andalso`? or why do we want to avoid it?

The _implicit_ specification for `andalso` (giving its _desired_ semantics) is `boolean() andalso boolean() -> boolean()`, but the actual way in which `andalso` work is (as you noticed in your code) `boolean() andalso Type -> Type`. In other words, while it's _semantically_ expected for the second expression to be boolean… it works anyway if it's not because that's how the VM optimizes this code.

Taking advantage of an optimization to mess up with the expected semantics of an expression is _not a great thing to do_. First of all, nothing prevents the OTP team from changing that behaviour in the future and crash if the second part is not a boolean. But more importantly… for developers reading that code… finding something that's not a boolean there is unexpected and _surprising future devs_ is not usually a good thing in terms of maintainability.

elbrujohalcon:
I'll leave the PR open for another day so @paulo-ferraz-oliveira can review… and if he doesn't request changes by then, I'll merge it.

doc_rules/elvis_style/atom_naming_convention.md
```diff
@@ -2,18 +2,25 @@
 
 (since [1.0.0](https://github.com/inaka/elvis_core/releases/tag/1.0.0))
 
-All atoms should be named according to the regular expression provided. Atoms enclosed in
-apostrophes have special meaning and are thus handled by a different configuration option (use
-`same` if you want the same value as `regex`).
+All atoms should be named according to the regular expression provided.
+Except if it matches with a defined `forbidden_regex`.
```
paulo-ferraz-oliveira, line 6:
```suggestion
All atoms should be named according to the provided regular expression,
except if they match with a defined `forbidden_regex`.
```

doc_rules/elvis_style/function_naming_convention.md
```diff
@@ -1,13 +1,16 @@
 # Function Naming Convention
 
 All functions should be named according to the regular expression provided.
+Except if it matches with a defined `forbidden_regex`.
```
paulo-ferraz-oliveira, line 4:
```suggestion
All functions should be named according to the provided regular expression,
except if they match with a defined `forbidden_regex`.
```

doc_rules/elvis_style/module_naming_convention.md
```diff
@@ -1,13 +1,16 @@
 # Module Naming Convention
 
 All modules should be named according to the regular expression provided.
+Except if it matches with a defined `forbidden_regex`.
```
paulo-ferraz-oliveira, line 4:
```suggestion
All modules should be named according to the provided regular expression,
except if they match with a defined `forbidden_regex`.
```

doc_rules/elvis_style/variable_naming_convention.md
```diff
@@ -1,13 +1,16 @@
 # Variable Naming Convention
 
 All variables should be named according to the regular expression provided.
+Except if it matches with a defined `forbidden_regex`.
```
paulo-ferraz-oliveira, line 4:
```suggestion
All variables should be named according to the provided regular expression,
except if they match with a defined `forbidden_regex`.
```

src/elvis_style.erl
```diff
@@ -72,14 +72,23 @@
 -define(FUNCTION_NAMING_CONVENTION_MSG,
         "The function ~p does not respect the format defined by the "
         "regular expression '~p'.").
+-define(FORBIDDEN_FUNCTION_NAMING_CONVENTION_MSG,
+        "The function ~p is written in a forbidden format"
+        "defined by the regular expression '~p'.").
```
paulo-ferraz-oliveira, line 77: All of these should refer "name" instead of format, because function format is odd to read. I'll probably not be able to comment on the other warnings because they're outside the scope of the pull request, but you could change them all to e.g. "Function ~p's name does not...", "Variable ~p's name, on line ~p, does not...", etc.

test/style_SUITE.erl
```diff
@@ -184,7 +184,16 @@ verify_function_naming_convention(Config) ->
                               elvis_style,
                               function_naming_convention,
                               RuleConfig4,
-                              PathIgnored).
+                              PathIgnored),
+
+    % forbidden
+    PathForbidden = "forbidden_function_naming_convention." ++ Ext,
+    [_, _, _] =
```
paulo-ferraz-oliveira, line 191: It would be nice to have the line numbers here, even if that means that changing the files the tests would fail, because it's easier to review.

test/style_SUITE.erl
```diff
@@ -184,7 +184,16 @@ verify_function_naming_convention(Config) ->
...
+        elvis_core_apply_rule(Config,
+                              elvis_style,
+                              function_naming_convention,
+                              #{regex => "^[a-z](_?[a-z0-9]+)*$", forbidden_regex => "[0-9]"},
```
paulo-ferraz-oliveira, line 195: You could/should be building these from a call to the function that defines them (`default`), otherwise it makes for more complex maintenance.

paulo-ferraz-oliveira (approved):
Generally approving, but did comment on a few things that could be improved.

bormilan:
> > But what's the problem with `andalso`? or why do we want to avoid it?
> 
> The _implicit_ specification for `andalso` (giving its _desired_ semantics) is `boolean() andalso boolean() -> boolean()`, but the actual way in which `andalso` works is (as you noticed in your code) `boolean() andalso Type -> Type`. In other words, while it's _semantically_ expected for the second expression to be boolean… it works anyway if it's not because that's how the VM optimizes this code.
> 
> Taking advantage of an optimization to mess up with the expected semantics of an expression is _not a great thing to do_. First of all, nothing prevents the OTP team from changing that behaviour in the future and crash if the second part is not a boolean (that would be semantically correct, anyway). But more importantly… for developers reading that code… finding something that's not a boolean there is unexpected and _surprising future devs_ is not usually a good thing in terms of maintainability.

Thank you for this explanation. I truly appreciate these "small" things that I can't learn anywhere else. Thanks again for the effort you put into these reviews!

paulo-ferraz-oliveira:
@bormilan, it's also one of the reasons behind https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/no_block_expressions.md. I was seeing (and writing, since Erlang didn't forbid me) code like:

```erlang
Connected andalso begin
  NumberOfEvents = num_events(...)
  NumberOfEvents > 10 orelse log(...)
end
```

While that works, it's generally messy to maintain.

bormilan:
> @bormilan, it's also one of the reasons behind https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/no_block_expressions.md. I was seeing (and writing, since Erlang didn't forbid me) code like:
> 
> ```erlang
> Connected andalso begin
>   NumberOfEvents = num_events(...)
>   NumberOfEvents > 10 orelse log(...)
> end
> ```
> 
> While that works, it's generally messy to maintain.

Thank you for the hint and all the reviews, too!

test/style_SUITE.erl
bormilan: I see, you are right.

test/style_SUITE.erl
bormilan: Yes, you are right, but this rule does not return the line number, only the name of the function and the `regex`.
Now it works like this:
```erlang
            Info = [FunctionNameStr, Regex],
            Result = elvis_result:new(item, Msg, Info, 1),
```
So the number `1` is hardcoded to every Result fun, and we do not add it to the `Info` list.

elbrujohalcon (approved):

bormilan:
If it's okay, I will rebase it to prepare for the merge.

elbrujohalcon:
We can just _squash and merge_, no need for rebasing.

PR #383, opened 2025-02-16
New `elvis_style' rule: `no_boolean_in_comparison`

### LINKED ISSUE #283: New rule: Avoid Comparison to Boolean
### Brief Description

A rule that disallows comparison to booleans.

### Should be on by default?

**YES**

### Reasoning

This is because these expressions evaluate to `true` or `false`, so you could get the same result by using either the variable directly or negating the variable.

### Examples

```erlang
-module bad.

-export [my_fun/2].

my_fun(P1, P2) ->
    case do:something(P1) == false orelse do:something(P2) == false of
        true -> "There is a false";
        false -> "All true"
    end.
```

```erlang
-module good.

-export [my_fun/2].

my_fun(P1, P2) ->
    case (not do:something(P1)) orelse (not do:something(P2)) of
        true -> "There is a false";
        false -> "All true"
    end.

%% … or even …

my_fun(P1, P2) ->
    case do:something(P1) andalso do:something(P2) of
        true -> "All true";
        false -> "There is a false"
    end.
```

### Origin (#281)

Inspired by the `ComparisonToBoolean` rule from [Dogma](https://github.com/lpil/dogma/blob/master/docs/rules.md#comparisontoboolean).

bormilan:
# Description

Finding the `op` nodes that have a child with the atom value `true` or `false`, if the operation is `==`.

Closes #283.


doc_rules/elvis_style/no_boolean_in_comparison.md
```diff
@@ -0,0 +1,19 @@
+# No Boolean In Comparison
+
+Avoid using booleans in comparasions. This rule disallows it.
```
elbrujohalcon, line 3:
```suggestion
Avoid comparing expressions to booleans. Prefer using boolean operators directly.
```

test/examples/fail_no_boolean_in_comparison.erl
```diff
@@ -0,0 +1,9 @@
+-module(fail_no_boolean_in_comparison).
+
+-export [my_fun/2].
+
+my_fun(P1, P2) ->
+    case do:something(P1) == false orelse do:something(P2) == false of
+        true -> "There is a false";
+        false -> "All true"
+    end.
```
elbrujohalcon, line 9: Do you mind adding a few extra examples here? At least one using `true` in the comparison and one having the boolean on the left of it. :)
But maybe as well…

```erlang
-record(you, {can, find = this:stuff() == even_in:record(definitions)}).

other_examples(A) ->
    %% An if clause
    X = if A == true -> "why do you do this?" end,

    %% A regular match operator
    Y = X == true,

    %% A list comprehension (different places)
    [ E == true || E <- build:a_list(), false == check:something_on(E)],

    %% An anonymous function head
    F = fun(X == true, false == Y) -> {shake, my, head} end,
    F(X, Y).
```

elbrujohalcon (changes requested):
Great work, as usual, @bormilan 
I think this rule should be in the default rulesets, not just the strict one.
It should be in all erlang-bound rulesets (i.e., erl_files, hrl_files, beam_files).

elbrujohalcon (changes requested):
With a few extra tests, this will be ready to go, @bormilan

elbrujohalcon:
As a note… Today I found another example of this at my job…
```erlang
%% There was something like this…
case the:result(of, something) of
    Bool when Bool == true; Bool == false -> do:something_with(the, Bool);
    NotBool -> do:something_else_with(this, NotBool, thingy)
end
```

Even when it seems like an exception to this rule, it is not. Because it can be written as…
```erlang
%% There was something like this…
case the:result(of, something) of
    Bool when is_boolean(Bool) -> do:something_with(the, Bool);
    NotBool -> do:something_else_with(this, NotBool, thingy)
end
```

elbrujohalcon (approved):

elbrujohalcon:
Excellent work, as usual, @bormilan !

PR #385, opened 2025-03-09
Add option `count_docs` to rule `max_module_length`

### LINKED ISSUE #384: max_module_length: Add `count_docs` flag

[//]:# (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...])

No, it's just a suggestion.

[//]:# (A clear and concise description of what you want to happen.)

Add a new flag (maybe) called `count_docs` that ignores the `-moduledoc` and `-doc` attributes introduced in OTP-27.

Reaching the default 500 `max_length` with a well-documented module is easy.

[//]:# (A clear and concise description of any alternative solutions or features you've considered.)

It's simple to increase the max length:

```erlang
{elvis_style, max_module_length, #{max_length => 1000}},
```

But it would be nice to just ignore the `-moduledoc` and `-doc` attributes, like:

```erlang
{elvis_style, max_module_length, #{count_docs => false}},
```

The `count_comments` flag only works for the _old_ docs style.

## Additional Context

[//]:# (Add any other context or screenshots about the feature request here.)

I'll suggest it to be `false` as the default.

williamthome:
# Description

This PR adds the `count_docs` option to the `max_module_length` style rule.
When false, ignores `-moduledoc` and `-doc` attributes.
The default is `false`, therefore, documentation will be ignored by default.

Closes #384 

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -301,8 +301,6 @@ default(RuleWithEmptyDefault)
       max_length => integer()}.
 -type function_naming_convention_config() ::
     #{ignore => [ignorable()], regex => string()}.
-%% It's the same as `binary:part/0`, but that type was not exported before
-%% OTP-27 (despite its existence since OTP-R14B).
 -type binary_part() :: {Start :: non_neg_integer(), Length :: integer()}.
 
 -spec function_naming_convention(elvis_config:config(),


test/style_SUITE.erl
```diff
@@ -1434,7 +1452,7 @@ verify_no_successive_maps(Config) ->
                               #{ignore => [Module]},
                               Path).
 
--else.
+- else .
```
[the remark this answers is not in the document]
elbrujohalcon: Are you using Erlang/OTP 27? This misformatting is usually the result of running rebar3_format on a version that it doesn't handle correctly.
Context: https://github.com/AdRoll/rebar3_format/discussions/366

I believe we should start moving projects like this one to `erlfmt`. @paulo-ferraz-oliveira what do you think?

elbrujohalcon (approved):
This rule keeps surprising me (mostly because I don't use it). I thought it belonged to `elvis_text_style` (i.e. rules that treat modules as _just text_), not `elvis_style`.
In any case, I think that the way you implemented it is fine. Although a more _precise_ way (which I don't even know if it's possible, to be fair) would've been to somehow use the AST to extract all the docs and moduledocs (that is, instead of using a regex).
I'm approving anyway since I don't even know if it is possible to use the AST as I think.

williamthome:
> I'm approving anyway since I don't even know if it is possible to use the AST as I think.

AST was my first attempt, but it does not provide enough information to determine exactly the start and end of the attribute.
Is the same problem of the formatter :sweat_smile: 
I cannot distinguish between a single or multi-line of documentation.

I also tried using the `code:get_docs/1` function, but it concatenates the sequence of doc attributes, like `-doc "foo". -doc #{since => "0.1"}.` and requires an extra step which is to check the OTP version >= 27.

Also, both approaches above required `debug_info` in compile options.

The regex was the more easy and simple way of implementing that option.
In summary, this one was more tricky than I thought.

test/style_SUITE.erl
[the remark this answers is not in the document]
williamthome: Yea, OTP-27.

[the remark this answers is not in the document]
elbrujohalcon: Yeah, well… Until we switch to `erlfmt`, I think we'll need to stick to OTP26 for development. 🤷🏻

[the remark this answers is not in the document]
williamthome: Reverted in [5649820](https://github.com/inaka/elvis_core/pull/385/commits/564982047209c38fb5444b0696c7e8de60c9cb40).

src/elvis_style.erl
```diff
@@ -2459,3 +2474,33 @@ get_root(Config, Target, RuleConfig) ->
...
+        {match, Parts} ->
+            [Part || [Part] <- Parts];
+        nomatch ->
+            []
+    end.
+
+-spec ignore_bin_parts(Src, [DocPart]) -> [TextPart]
+    when Src :: binary(),
+         DocPart :: binary:part(),
+         TextPart :: binary:part().
```
elbrujohalcon, line 2493: It seems like Dialyzer doesn't know this type… at least on some versions of Erlang/OTP… 🤷🏻 
Do you mind using the internal definition of it (i.e., `{Start :: non_neg_integer(), Length :: integer()}`) until we no longer support versions where it's not defined?

elbrujohalcon (changes requested):

src/elvis_style.erl
williamthome: Really strange. This type is since OTP R14B (2010). Anyway, I'll change it.

williamthome: There is no dialyzer error in OTP-27 (at least locally).

elbrujohalcon: The type may exist since OTP R14B, but it's possible that it was not exported back then… only after OTP27 ;)

elbrujohalcon: Indeed! https://github.com/erlang/otp/commit/9ecff702048dc67b45e864c16adea1679fa5c982

And, proving once more that the world is a napkin… It was exported by no other than our friend @paulo-ferraz-oliveira / @kivra-pauoli :)

williamthome:
Ok, it seems I'll need to split the tests and run specifically for OTP-27.
I'll take a look at it when have more time to do this.

src/elvis_style.erl
williamthome: Resolved in [ebb45b5](https://github.com/inaka/elvis_core/pull/385/commits/ebb45b50a642f0864ccf9ebabfe87546159cf0eb), but I have to fix the tests now.

src/elvis_style.erl
```diff
@@ -300,6 +301,9 @@ default(RuleWithEmptyDefault)
       max_length => integer()}.
 -type function_naming_convention_config() ::
     #{ignore => [ignorable()], regex => string()}.
+% It's the same as `binary:part/0`, but the dialyzer is not recognizing it below
+% OTP-27 (despite its existence since OTP-R14B).
```
elbrujohalcon, line 305:
```suggestion
%% It's the same as `binary:part/0`, but that type was not exported before
%% OTP-27 (despite its existence since OTP-R14B).
```

williamthome:
@elbrujohalcon Do you know what the CI error means? Running it `DIAGNOSTICS=1` with says nothing:

```bash
$ DIAGNOSTICS=1 rebar3 ex_doc
===> Verifying dependencies...
===> Analyzing applications...
===> Compiling elvis_core
===> Running edoc for elvis_core
edoc: error in doclet 'edoc_doclet_chunks': {'EXIT',error}.
===> An unknown error occurred generating doc chunks with edoc. Run with DIAGNOSTICS=1 for more details.
```

elbrujohalcon:
@williamthome I have no clue. @paulo-ferraz-oliveira do you know what's going on there?

williamthome:
Hmm it's the `binary:part/0` on this comment I've introduced:

```
%% It's the same as `binary:part/0`, but that type was not exported before
%% OTP-27 (despite its existence since OTP-R14B).
```

I'll remove it.

elbrujohalcon (approved):

PR #386, opened 2025-03-09
Add `no_whitespace` to the `line_length` options

williamthome:
# Description

This PR adds the `no_whitespace` option to the `line_length` style rule.
When `false`, allows include items such as long URLs without being forced to break them in the middle.
The default is `true`, therefore, no breaking changes.

### DIFF
--- a/doc_rules/elvis_text_style/line_length.md
+++ b/doc_rules/elvis_text_style/line_length.md
@@ -13,11 +13,15 @@ No line should be longer than a given limit. Comments can be skipped.
   - `any` means _don't emit a warning if the part of the line that goes over `Limit` belongs to a
   comment_
   - `whole_line` means _don't emit a warning if the line that goes over `Limit` is __just__ a comment_
+- `no_whitespace :: boolean()`
+  - default: `true`, means _emit a warning when there is a whitespace beyond the configured `Limit`_.
+    When `false`, allows include items such as long URLs without being forced to break them in the middle
 
 ## Example
 
 ```erlang
 {elvis_text_style, line_length, #{ limit => 100
                                  , skip_comments => false
+                                 , no_whitespace => true
                                  }}
 ```
--- a/src/elvis_text_style.erl
+++ b/src/elvis_text_style.erl
@@ -32,7 +32,9 @@
 
 -spec default(Rule :: atom()) -> DefaultRuleConfig :: term().
 default(line_length) ->
-    #{limit => 100, skip_comments => false};
+    #{limit => 100,
+      skip_comments => false,
+      no_whitespace => true};
 default(no_tabs) ->
     #{};
 default(no_trailing_whitespace) ->
@@ -56,8 +58,9 @@ default(no_redundant_blank_lines) ->
 line_length(_Config, Target, RuleConfig) ->
     Limit = option(limit, RuleConfig, line_length),
     SkipComments = option(skip_comments, RuleConfig, line_length),
+    NoWhitespace = option(no_whitespace, RuleConfig, line_length),
     {Src, #{encoding := Encoding}} = elvis_file:src(Target),
-    Args = [Limit, SkipComments, Encoding],
+    Args = [Limit, SkipComments, Encoding, NoWhitespace],
     elvis_utils:check_lines(Src, fun check_line_length/3, Args).
 
 -spec no_tabs(elvis_config:config(),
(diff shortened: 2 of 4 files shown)


elbrujohalcon:
I'm really really not a fan of this configuration. It's one of the things that I dislike about `erlfmt`: I put a line length in the config and it still leaves my code with some lines that go over that number (¬_¬)
But, as usual, it's a matter of personal preference. Therefore, a config with the default set to _strict mode_ should be fine.

doc_rules/elvis_text_style/line_length.md
```diff
@@ -13,11 +13,15 @@ No line should be longer than a given limit. Comments can be skipped.
   - `any` means _don't emit a warning if the part of the line that goes over `Limit` belongs to a
   comment_
   - `whole_line` means _don't emit a warning if the line that goes over `Limit` is __just__ a comment_
+- `no_whitespace :: boolean()`
```
elbrujohalcon, line 16:
```suggestion
- `skip_long_words :: boolean()`
```
I don't like `no_whitespace` as a name too much (mostly because it's hard to understand what it does without this docs). `skip_long_words` (with the default being `false`) sounds a bit better to me… but I still don't _love_ it. If we were talking about a formatter configuration it would likely be either `break_long_words` or `preserve_long_words`, but Elvis doesn't rewrite code. So… @williamthome … I'm not super-strongly against `no_whitespace`. If you think that's better than my suggestion, just resolve this comment and move forward :)

test/style_SUITE.erl
```diff
@@ -1434,7 +1444,7 @@ verify_no_successive_maps(Config) ->
                               #{ignore => [Module]},
                               Path).
 
--else.
+- else .
```
elbrujohalcon, line 1447: Are you using Erlang/OTP 27? This misformatting is usually the result of running rebar3_format on a version that it doesn't handle correctly.
Context: https://github.com/AdRoll/rebar3_format/discussions/366

I believe we should start moving projects like this one to `erlfmt`. @paulo-ferraz-oliveira what do you think?

elbrujohalcon (approved):
I think the changes are mergeable, but I left some comments.
I would wait for @paulo-ferraz-oliveira's review and, maybe, @bormilan's too, tho.

test/style_SUITE.erl
williamthome: Yes, OTP-27.

doc_rules/elvis_text_style/line_length.md
williamthome: Agree. Rereading, `no_whitespace` does not give the feel of the intention of it.
Maybe `no_whitespace_after_limit`?
I think your suggestions are fine, but I don't have a formed opinion about this right now.

test/style_SUITE.erl
bormilan: I like the idea of using `erlfmt` in this project @elbrujohalcon. I'm also planning on using it in other repos, so I would be happy to see it in `evlis_core` too. (and also sad bc I really liked `rebar3_format`)

doc_rules/elvis_text_style/line_length.md
williamthome: I picked up the `no_whitespace` name and a part of the description of this option from a Markdown linter ([See the complete line length rule](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md013---line-length)):

```md
[...]
This rule has an exception where there is no whitespace beyond the configured line length.
This allows you to still include items such as long URLs without being forced to break them in the middle.
```

williamthome:
> I'm really really not a fan of this configuration. It's one of the things that I dislike about `erlfmt`: I put a line length in the config and it still leaves my code with some lines that go over that number (¬_¬)

IIRC, `erlfmt` does not break long words and uses a similar approach to what I implemented in this PR. But I'm unsure. I'll mix this option with `erlfmt` and post the result.

> But, as usual, it's a matter of personal preference.

I Agree. I really dislike writing a URL in the code or comments and need to break this URL into multiple lines to satisfy the `elvis` rules.

elbrujohalcon:
> > I'm really really not a fan of this configuration. It's one of the things that I dislike about `erlfmt`: I put a line length in the config and it still leaves my code with some lines that go over that number (¬_¬)
> 
> IIRC, `erlfmt` does not break long words and uses a similar approach to what I implemented in this PR. But I'm unsure. I'll mix this option with `erlfmt` and post the result.
> 
Indeed. It does not break long words, which is something I would **like** it to do. Or, at the very least, to show me a warning like "You asked me to keep your lines at 100 characters, but this one I couldn't trim: …" 🤷🏻

doc_rules/elvis_text_style/line_length.md
williamthome: Renamed in https://github.com/inaka/elvis_core/pull/386/commits/b888fbe58d08164d2f98bce882d9782519a4b33b

test/style_SUITE.erl
williamthome: So, well.. what can I do about this on my two opened PRs?

src/elvis_text_style.erl
```diff
@@ -181,28 +184,38 @@ remove_comment(Line) ->
...
+        Len when NoWhitespace ->
             Msg = ?LINE_LENGTH_MSG,
-            Info = [Num, Large],
+            Info = [Num, Len],
             Result = elvis_result:new(item, Msg, Info, Num),
             {ok, Result};
-        _ ->
-            no_result
+        Len ->
+            case binary:match(Line, <<"\s">>, [{scope, {Limit, Len - Limit}}]) of
```
bormilan, line 210: In this line, you match to the whitespace characters, after the limit.
Why does this solution work?
If I have a line containing a string with spaces, but at the end it has a long word, so in the scope: `{Limit, Len - Limit}` it will not contain any whitespace. That line could be broken into multiple lines.

I don't say it's bad, I'm just curious how this works.

test/style_SUITE.erl
elbrujohalcon: Revert the change and push to GitHub without formatting.
If you want to run the tests locally, use OTP26.

elbrujohalcon: Or use two -ifs instead of-if/-else 🫣

src/elvis_text_style.erl
williamthome: It works because I'm matching on the first whitespace after the limit, for example:

```erlang
1 | %% NOTE: The `|` character is at the 100 characters limit.
2 |
3 | % Single line comment with a URL https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_text|_style/line_length.md
4 | example_1() ->
5 |     "This string is 108 characters long and does not have a whitespace after the 100 characters    |limit.".
6 |
7 | example_2() ->
8 |     "This string is 108 characters long and          have a whitespace after the 100 characters    | limit.".
```

Only **line number 8** will emit the warning, all the others are < 100 or do not contain a whitespace after the limit:

```erlang
1> Line = <<"    \"This string is 108 characters long and          have a whitespace after the 100 characters    | limit.\".">>.
<<"    \"This string is 108 characters long and          have a whitespace after the 100 characters    | limit.\".">>
2> Limit = 100.
100
3> Len = string:length(Line).
109
4> binary:match(Line, <<"\s">>, [{scope, {Limit, Len - Limit}}]).
{100,1}
```

The match result in **line number 5**:

```erlang
1> Line = <<"    \"This string is 108 characters long and does not have a whitespace after the 100 characters    |limit.\".">>.
<<"    \"This string is 108 characters long and does not have a whitespace after the 100 characters    |limit.\".">>
2> Limit = 100.
100
3> Len = string:length(Line).
108
4> binary:match(Line, <<"\s">>, [{scope, {Limit, Len - Limit}}]).
nomatch
```

Does this answer your question?

williamthome:
I've fixed a mistake I made on the line length message info and reverted the formatter changes.

test/style_SUITE.erl
williamthome: Reverted in [67ce96d](https://github.com/inaka/elvis_core/pull/386/commits/67ce96dc58b2cd9fcf04dbd1a3505b01172e766a).

src/elvis_text_style.erl
bormilan: I see, so was right about how this works. My question now is that why we want to allow a line to be written in a single line, just because the last word(that hits the 100) has no whitespace?
Or do I misunderstand something?
If I make a line during development like this:
```
example_1() ->
  "This string is 108 characters long and does not have a whitespace after the 100 characters    |limit.".
```
and then running `rebar3 lint` will warn me about this line is longer than my configured max limit.
and then I fix this line to this:
```
example_1() ->
  "This string is 108 characters long and does not have a"
  "whitespace after the 100 characters    |limit.".
```
and this is good in my opinion.

I don't really get it why we want this. If we have a URL that is longer than 100 chars, and we don't want to break it, that's understandable, but in this way I'm confused to be honest.

PR #395, opened 2025-04-16
[Fix #287] New rule: no_operation_on_same_value

### LINKED ISSUE #287: New rule: Operation on Same Value
### Brief Description

The idea is to emit a warning in situations where an infix operation is applied to the same value on both sides of it, like the ones below...

```erlang
X = X
X == X
{A, B, C} /= {A, B, C}
A -- A
…
```

### Should be on by default?

**YES**

### Options

-  `operations :: [atom()]` (default: `['=', '==', '/=', '=:=', '=/=', 'and', 'andalso', 'or', 'orelse', …]`)

For the default, you need to find all Erlang infix operators and figure out which ones make sense in this context and which ones doesn't. For instance, the result of `X -- X` is obvious (`[]`), but the result of `X ++ X` is not.

### Reasoning

> In practice they are likely the result of a debugging session or were made by mistake.

### Origin (#281)

Inspired by the `Warning.OperationOnSameValues` rule from [Credo](https://hexdocs.pm/credo/Credo.Check.Warning.OperationOnSameValues.html).
Also inspired by the `Warning.BoolOperationOnSameValues` rule from [Credo](https://hexdocs.pm/credo/Credo.Check.Warning.BoolOperationOnSameValues.html).

elbrujohalcon:
# Description

Add a new rule: `no_operation_on_same_value`

Closes #287.

> [!NOTE]
> Since I couldn't find a way to traverse the zipper provided by `ktn_code`/`elvis_code` and just _remove all locations_, I had to do that _manually_ through recursion. I hope you don't mind :)


bormilan:
My first thought was to read the source of the file, to assert the Left and Right in a text format, but that's bad because the Left and Right things can be long and stretched over multiple lines. (if it makes sense)
I'm interested in making a traverse for a node to be able to do such things.

elbrujohalcon:
@bormilan https://github.com/inaka/katana-code/issues/86 I created this ticket for the implementation of the missing function. Feel free to take it, if you want.

bormilan (approved):

elbrujohalcon:
@filmor Do you mind reporting a bug with it?

PR #397, opened 2025-04-24
refactor(#389): Error handling at wrong rules in `elvis.config`

### LINKED ISSUE #389: Proper error handling at wrong rules in `elvis.config`
## Give a message about the cause of the error if a non-existing rule added in `rebar.config`

I realized that if you give a non-existent (or make a typo) rule name in `elvis.config` you will get a bunch of error messages coming from elvis_core.

The problem is that if you use Elvis for the first time, you won't know what the problem is.

I would like to add some proper error messages to the user about what he did wrong.

I don't know the alternatives for this problem. The only solution in my head is to display some message about the exact problem.

bormilan:
# Description

I made a little refactor on the `elvis_config:validate` using `maybe`. Feel free to roast it; I'm still experiencing this feature and its capabilities.

I skipped the validation of the rule configs in this step. (Should I implement it too, or is it okay to leave it for later?)

Now this solution can:
    - spot if a non-existent module was given in the `elvis.config`
    - spot is a nonexistent rule(for an existing module) was given in the `elvis.config`
    - spot any other issues that were implemented in the past

I have a test that I commented out because I'm not sure what it was intended to test 😅 (I marked it in the files)
I see that it skips to load a module called `user_defined_rules`, but I'm not sure whether I need to keep this test or change it to expect my new error message.

Also, I want to add more tests, so feel free to suggest some if you have some cases.

Closes #389;.


bormilan:
I'm planning on adding:
    - spot non-existent rulesets
    - spot invalid rule config
    - spot non-existent dirs
    - spot non-existent filter

I don't know whether I should do all in this PR or cut it into smaller tasks

test/elvis_SUITE.erl
```diff
@@ -374,15 +375,15 @@ rock_with_rule_groups(_Config) ->
...
-    meck:unload(elvis_file),
+    % meck:new(elvis_file, [passthrough]),
+    % Dirs = ["../../_build/test/lib/elvis_core/test/examples"],
+    % [File] = elvis_file:find_files(Dirs, "small.erl"),
+    % Path = elvis_file:path(File),
+    % ConfigPath = "../../config/elvis-test-pa.config",
+    % ElvisConfig = elvis_config:from_file(ConfigPath),
+    % ok = elvis_core:rock_this(Path, ElvisConfig),
+    % 0 = meck:num_calls(elvis_file, load_file_data, '_'),
+    % meck:unload(elvis_file),
```
paulo-ferraz-oliveira, line 386: This doesn't seem right. Did you forget to uncomment it, or is it not working?

src/elvis_config.erl
```diff
@@ -1,5 +1,7 @@
 -module(elvis_config).
 
+-feature(maybe_expr, enable).
```
paulo-ferraz-oliveira, line 3: Since our contract is [now](https://github.com/inaka/elvis_core/blob/main/rebar.config#L5) `>= 25` isn't there a way to set this in `rebar.config`, thus make it available for all modules? Or is there a constraint to doing so?

src/elvis_config.erl
```diff
@@ -82,30 +84,57 @@ validate(Config) ->
...
+        {_, true} ?= {missing_rules, maps:is_key(rules, RuleGroup)},
+        {invalid_rules, []} ?= invalid_rules(maps:get(rules, RuleGroup))
+    else
+        {missing_rules, false} ->
+            maybe_missing_rules(maps:is_key(ruleset, RuleGroup), RuleGroup);
+        {invalid_rules, InvalidConfigElements} ->
+            throw({invalid_config, {{invalid_rules, InvalidConfigElements}}});
+        {_, Error} ->
+            throw({invalid_config, {Error, RuleGroup}})
+    end.
```
paulo-ferraz-oliveira, line 101: I'm not sure we have this pattern in the code base already, but tagged tuples are hard to maintain, and read. Check this out: https://hexdocs.pm/credo/Credo.Check.Readability.WithCustomTaggedTuple.html

Maybe, as per what Credo suggests, you could try a similar approach? i.e. 

```
maps:is_key(src_dirs, RuleGroup) or maps:is_key(dirs, RuleGroup)
```

could become a function, say `is_missing_dir`, and then it would either return `ok`, or `{error, _}`. If all the other checks return a similar `ok` | `{error, _}` pattern, you could maybe simplify the structure and improve readability (?) From what I can tell you're always trying to pattern match and then throwing in case of error, so it seems like a good fit.

And maybe we can also take inspiration from that rule and create an issue so that we disallow tagged tuples in `maybe` (?) @elbrujohalcon

src/elvis_style.beam
paulo-ferraz-oliveira, line 1: Not sure you intended on pushing this (?)

test/examples/invalid_rules.elvis.config
```diff
@@ -0,0 +1,47 @@
...
+                filter => "*.erl",
+                rules =>
+                    [
+                        {elvis_style, invalid_dynamic_call, #{ignore => [elvis_core]}},
+                        {elvis_style, dont_repeat_yourself, #{min_complexity => 20}},
+                        {elvis_style, no_debug_call, #{ignore => [elvis_result, elvis_utils]}},
+                        {elvis_style, god_modules, #{ignore => [elvis_style]}},
+                        {elvis_style, no_throw, disable},
+                        {elvis_style, not_existing_rule},
+			{elvis_style, what_is_this_rule, #{ignore => [elvis_style]}},
```
paulo-ferraz-oliveira, line 15: Indentation seems off. Does this mean we're not formatting these files?

paulo-ferraz-oliveira:
> I made a little refactor on the `elvis_config:validate` using `maybe`. Feel free to roast it; I'm still experiencing this feature and its capabilities.

I think it's Ok to keep, since it's now part of the language and inside our support scope, but I did leave a comment on maybe how to improve the current implementation's readability.
 
> I skipped the validation of the rule configs in this step. (Should I implement it too, or is it okay to leave it for later?)

For me it's Ok to leave for later; just open a ticket 😄 
 
> I have a test that I commented out because I'm not sure what it was intended to test 😅 (I marked it in the files) I see that it skips to load a module called `user_defined_rules`, but I'm not sure whether I need to keep this test or change it to expect my new error message.

Yeah, I commented on this. I'd keep the test if it wasn't failing, and add a comment on its intention if you can find it in Git history. Just removing it (or commenting it out) while implementing a new feature seems odd.

I like the general direction, and have left a few comments.

test/elvis_SUITE.erl
```diff
@@ -424,6 +425,26 @@ rock_with_umbrella_apps(_Config) ->
...
 
+-spec rock_with_invalid_rules(config()) -> any().
+rock_with_invalid_rules(_Config) ->
+    ConfigPath = "../../test/examples/invalid_rules.elvis.config",
+    ElvisConfig = elvis_config:from_file(ConfigPath),
+    ExpectedErrorMessage =
+        {{invalid_rules, [
+            {invalid_rule, {elvis_style, not_existing_rule}},
+            {invalid_rule, {elvis_style, what_is_this_rule}},
+            {invalid_module, {not_existing_module}},
```
paulo-ferraz-oliveira, line 436: Seems odd you'd have to wrap the module name in a tuple (?)

test/elvis_SUITE.erl
paulo-ferraz-oliveira: This seems to have come from https://github.com/inaka/elvis_core/pull/72. From what I can tell, from the difference to `rock_this_not_skipping_files`, it tests that a given file was not loaded+parsed unnecessarily. I'd thus keep it.

test/examples/invalid_rules.elvis.config
paulo-ferraz-oliveira: (we seem to be - I checked - so I'm assuming it was unintentional)

src/elvis_config.erl
```diff
@@ -82,28 +84,69 @@ validate(Config) ->
...
+            throw({invalid_config, Error});
+        InvalidRules ->
+            throw({invalid_config, {invalid_rules, InvalidRules}})
+    end.
+
+maybe_boolean_wrapper(Case, Flag) ->
+    case Case of
+        true -> ok;
+        false -> {error, Flag}
+    end.
```
elbrujohalcon, line 115:
```suggestion
maybe_boolean_wrapper(true, _Flag) -> ok;
maybe_boolean_wrapper(false, Flag) -> {error, Flag}.
```

src/elvis_config.erl
```diff
@@ -85,29 +85,48 @@ validate(Config) ->
...
-            {missing_dirs, maps:is_key(src_dirs, RuleGroup) or maps:is_key(dirs, RuleGroup)},
-        {_, false} ?=
-            {missing_filter, (maps:is_key(dirs, RuleGroup) and not maps:is_key(filter, RuleGroup))},
-        {_, true} ?= {missing_rules, maps:is_key(rules, RuleGroup)},
-        {invalid_rules, []} ?= invalid_rules(maps:get(rules, RuleGroup))
+        ok ?=
+            maybe_boolean_wrapper(
+                (maps:is_key(src_dirs, RuleGroup) or maps:is_key(dirs, RuleGroup)),
+                missing_dirs
+            ),
```
paulo-ferraz-oliveira, line 92: I'd prefer all of this to be encapsulated in its own properly named function. This way it's easier to reason on the check being performed.

Same for the other occurrences below.

src/elvis_config.erl
bormilan: I think I need to forget maybe because Erlang/OTP 25 does not support it. It is easier to move back to a case version.

src/elvis_config.erl
bormilan: Yes, that might be a better approach. I will also try a solution in that way. Thanks for your suggestion. I feel it is starting to achieve its best form, but it’s not quite there yet. :D

test/elvis_SUITE.erl
bormilan: I came to the solution to not throw errors when an invalid module is spotted, just print a message to the console.

Because this feature does exactly the opposite of mine, to be able to run with invalid configs, only skip them.

paulo-ferraz-oliveira: Does the original intention (in the PR I mentioned) hold, then?

src/elvis_config.erl
paulo-ferraz-oliveira: `maybe` **is** supported in OTP 25: https://www.erlang.org/doc/system/expressions.html#maybe

src/elvis_config.erl
paulo-ferraz-oliveira: Saw the changes. Looks better, thanks 😄

test/elvis_SUITE.erl
bormilan: I don't deeply understand the topic, but I think it's still relevant.
Maybe I should investigate it more, but first, I want to solve the more important problems, and then I will check on it.

test/examples/invalid_rules.elvis.config
paulo-ferraz-oliveira: This was flagged as resolved, but it's still not indented properly 😄

test/elvis_SUITE.erl
paulo-ferraz-oliveira: This is since reverted. I'll resolve, for now.

test/examples/invalid_rules.elvis.config
elbrujohalcon: Indentation seems _fine_ to me, according to `erlfmt` principles and its hint-taking algorithm 🤷🏻

paulo-ferraz-oliveira: This is what I'm seeing. Not sure we're seeing the same.

image

Do you mean that a mix between tabs (whatever spaces?) and spaces is Ok, in this sense?

elbrujohalcon: 🫤 I see…
image

paulo-ferraz-oliveira: 👍 then I don't know what's up, but either GitHub is really acting up or there's a mix of tabs and spaces 🤷

paulo-ferraz-oliveira: I'm Resolve'ing since I don't want to remove attention from the important parts of the PR.

bormilan:
Currently, I just don't know how to fix builds.

If I make a `vm.args` with the content of `-enable-feature maybe_expr` and run the `ERL_FLAGS="-args_file config/vm.args" rebar3 test` command, the build will run successfully.

But since it's a file path, in the Windows builds, I need to change it, so I need to find a way to change the CI a little bit to execute rebar commands with the proper `ERL_FLAGS`.

elbrujohalcon:
Yeah, @bormilan … I don't know what's the deal with ex_doc and maybe… but I would be super-pragmatic and… just don't run ex_doc in OTP25. Period. Even more, we can just not run ex_doc in anything lower than the latest version we use for CI (i.e. 27).

paulo-ferraz-oliveira:
@elbrujohalcon, should we flag `old_configuration_format` as deprecated in the scope of this change (or later, if you prefer to separate stuff)? We can also remove it while keeping the documentation for historical purposes...

elbrujohalcon:
@paulo-ferraz-oliveira yeah, let's remove it.

test/elvis_SUITE.erl
```diff
@@ -424,6 +425,24 @@ rock_with_umbrella_apps(_Config) ->
...
+            {invalid_rule, {elvis_style, what_is_this_rule}}
+        ]},
+    ok =
+        try
+            ok = elvis_core:rock(ElvisConfig),
+            fail
+        catch
+            {invalid_config, ExpectedErrorMessage} ->
+                ok
+        end.
```
elbrujohalcon, line 444:
```suggestion
    try
        ok = elvis_core:rock(ElvisConfig),
        ct:fail("Elvis should not have rocked with ~p", [ElvisConfig])
    catch
        {invalid_config, ExpectedErrorMessage} ->
            ok
    end.
```

elbrujohalcon (approved):
I think the PR is good as it is, but I left a tiny nit-picky comment for tests.
I also didn't check around to see what we usually use, but if we use pattern-matching everywhere instead of `ct:fail/1,2`, do ignore my comment.

.github/workflows/ci.yml
```diff
@@ -45,9 +45,13 @@ jobs:
       - name: Format check
         if: ${{ matrix.os == 'ubuntu-24.04' && matrix.otp_vsn == '26' }}
         run: rebar3 fmt --check
-      - name: Run test
+      - name: Run test Erlang/OTP 25
```
bormilan, line 48: Change back to "Run test"

elbrujohalcon: This is your PR, @bormilan … you can change it back :)

bormilan: Yes yes, but I need my personal computer for it, so I wanted to note it to myself not to forget 😅

test/elvis_SUITE.erl
bormilan: I like this more, and as I can see, we use pattern match if we match to "ok", and use ct:fail if we expect a fail.

bormilan:
> I'm planning on adding: - spot non-existent rulesets - spot invalid rule config - spot non-existent dirs - spot non-existent filter
> 
> I don't know whether I should do all in this PR or cut it into smaller tasks

I will open a ticket(or more) later in this week, or maybe a discussion first, about these.

src/elvis_config.erl
paulo-ferraz-oliveira: Given the changes to `vm.args` is this still required?

paulo-ferraz-oliveira: (in any case, if CI passes I'm voting for ✅ to move forward; we can always improve later)

paulo-ferraz-oliveira:
@bormilan, you have merge rights, correct? If so, it feels like you can go ahead and merge this.

elbrujohalcon:
Well… now you have to resolve the conflicts, @bormilan … but after that go ahead and (if CI is green) merge the thing.

bormilan:
It's strange that the test, `verify_no_single_match_maybe`, is failing for me locally.
I need to add the ERL_FLAG... to the command to make it work.

paulo-ferraz-oliveira:
> It's strange that the test, verify_no_single_match_maybe, is failing for me locally.

I'm gonna pull this code locally and run CI as GitHub does. I'll report on finding issues, if any.

bormilan:
No, it fails locally. I investigated for a few hours, but I found nothing.

elbrujohalcon:
On OTP25, I assume, right @bormilan ?

If that's the case, I wouldn't mind much… I think it should be fine to use OTP27 for development. In any case OTP28 is already out there… so very very soon we'll stop supporting OTP25 altogether.

paulo-ferraz-oliveira:
Sorry, was going to test this, got caught up with stuff. Retrying now, no promises...

paulo-ferraz-oliveira:
When I run `_build/default/bin/elvis_core`, the same as CI, after `rebar3 escriptize` I get a buncha `Warning: Invalid module: elvis_style`, but it doesn't exit with error. Is that what you're referring to?

paulo-ferraz-oliveira:
Is this supposed to happen? https://github.com/inaka/elvis_core/pull/397/checks#step:8:14

paulo-ferraz-oliveira:
@elbrujohalcon, maybe we could wrap that rule in a macro that'd only add it from OTP 27? (and we'd say so in the doc.)

paulo-ferraz-oliveira:
> Is this supposed to happen? https://github.com/inaka/elvis_core/pull/397/checks#step:8:14

I think you're not loading the module when analysing Elvis on Elvis.

src/elvis_config.erl
```diff
@@ -82,28 +84,71 @@ validate(Config) ->
...
+                RuleName,
+                ExportedRules
+            )
+        )
+    else
         false ->
-            throw({invalid_config, {missing_rules, RuleGroup}});
-        true ->
-            ok
+            elvis_utils:warn_prn("Invalid module: ~p~n", [Module]),
```
paulo-ferraz-oliveira, line 150: When is this supposed to be output? I got it because a module wasn't loaded and it was confusing because I didn't understand where it was coming from.

elbrujohalcon:
> @elbrujohalcon, maybe we could wrap that rule in a macro that'd only add it from OTP 27? (and we'd say so in the doc.)

Not my favorite thing to do, but yeah… it seems like a somewhat reasonable solution.

src/elvis_config.erl
elbrujohalcon: Yeah, this one could benefit from some extra context, so users can understand what's going on.

bormilan:
Okay, I will fix these, thanks!

src/elvis_config.erl
```diff
@@ -82,28 +84,74 @@ validate(Config) ->
...
+                ExportedRules
+            )
+        )
+    else
         false ->
-            throw({invalid_config, {missing_rules, RuleGroup}});
-        true ->
-            ok
+            elvis_utils:warn_prn(
+                "Invalid module was given,in elvis_config. Module: ~p~n",
```
paulo-ferraz-oliveira, line 151:
```suggestion
                "Invalid module (~p) specified in elvis.config.~n",
```

paulo-ferraz-oliveira (approved):
If CI's passing, and locally it works for you, we should be good to go.

src/elvis_config.erl
```diff
@@ -82,28 +84,74 @@ validate(Config) ->
...
-    case maps:is_key(src_dirs, RuleGroup) orelse maps:is_key(dirs, RuleGroup) of
-        false ->
-            throw({invalid_config, {missing_dirs, RuleGroup}});
-        true ->
-            ok
+    maybe
+        ok ?= maybe_missing_dirs(RuleGroup),
+        ok ?= maybe_missing_filter(RuleGroup),
+        ok ?= maybe_missing_rules(RuleGroup),
+        [] ?= maybe_invalid_rules(RuleGroup, maps:is_key(rules, RuleGroup))
```
elbrujohalcon, line 91: I think it would've been nicer if `maybe_invalid_rules/1` would've returned `ok | {error, {invalid_rules, …}}`, matching the other `maybe_*/1` functions here.
Furthermore, it can be implemented as…

```erlang
maybe_invalid_rules(#{rules := Rules}) ->
    case invalid_rules(Rules) of
        [] -> ok;
        InvalidRules -> {error, {invalid_rules, InvalidRules}}
    end;
maybe_invalid_rules(_) -> ok.
```

elbrujohalcon: And then you would not need the second `else` clause below.

src/elvis_config.erl
```diff
@@ -82,28 +84,74 @@ validate(Config) ->
...
+    IsInvalid = fun
+        (_, {_, _}) ->
+            false;
+        (Rule, none) ->
+            {true, {invalid_rule, {Module, Rule}}}
     end,
-    case maps:is_key(rules, RuleGroup) orelse maps:is_key(ruleset, RuleGroup) of
+    maybe
+        {file, _} ?= code:is_loaded(Module),
+        ExportedRules = apply(Module, module_info, [exports]),
```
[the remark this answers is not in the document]
bormilan: Why is it better to use 'erlang:' ?
I use 'apply' instead of 'Module:Function()', because there is an Elvis rule for this to eliminate.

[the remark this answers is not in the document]
elbrujohalcon: Because if you read the code and find a non-qualified function call (e.g. `apply(…)`) it may take you a few seconds to realize that there is `apply(…)` function defined in this module. If you read `erlang:apply(…)` it's immediately obvious.

It's basically the same reasoning as https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/no_import.md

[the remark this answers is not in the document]
bormilan: Oh, I see, thanks!

paulo-ferraz-oliveira:
> It saddens me a little bit that the single-match-maybe rule can't be used before OTP27, but it is what it is. And, since this is a new rule anyway, I accept its faith :)

A year will go by quickly, and when 29 comes out we can move the support range to 27-29 😄

src/elvis_config.erl
```diff
@@ -82,28 +84,74 @@ validate(Config) ->
...
+        IsInvalid(
+            RuleName,
+            proplists:lookup(
+                RuleName,
+                ExportedRules
+            )
+        )
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Simpler yet.

```erlang
        lists:keyfind(RuleName, 1, ExportedRules)
            orelse {true, {invalid_rule, {Module, RuleName}}}
```

src/elvis_config.erl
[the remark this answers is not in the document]
elbrujohalcon: Sounds like a good idea, yes. Although, people who's used to write `element(…, Tuple)`, `setelement(…)` or `length(List)` will not be happy with that rule. Nevertheless, it would be a good addition to the strict ruleset, yes.

src/elvis_config.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Didn't say it was a good idea, just simpler :) And since there's no linter rule for it... :)

src/elvis_config.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: We can add exceptions to the rule, as options 😄 we start with a small subset and keep adding while we get reports (if any)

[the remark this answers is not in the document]
paulo-ferraz-oliveira: https://github.com/inaka/elvis_core/issues/430

PR #412, opened 2025-05-20
Fixes 293: Add rule: No Single Match Maybe

### LINKED ISSUE #293: New rule: No Single Clause Maybe Statements
### Brief Description

The following code...

```erlang
maybe
    {ok, A} ?= a()
end
```

... should actually be written using pattern-matching:

```erlang
{ok, A} = a()
```

The `maybe` statement is unnecessary and the only difference between the two is that the exception raised in the case of error changes from `else_clause` to `badmatch`.

If an `else_clause` is present, though, the rule should not be applied.

### Should be on by default?

**YES**

### Options

- None.

### Reasoning

Single-pattern `maybe` blocks may be written more concisely or removed entirely.

### Origin

Inspired by #290.

elbrujohalcon:
# Description

Fixes #293: Add a new rule: No Single Match Maybe

Closes #&lt;issue&gt;.


elbrujohalcon:
@paulo-ferraz-oliveira / @bormilan : What do I need to do to be able to run the tests on OTP25 and OTP26? Do you know what I'm missing?

bormilan (approved):

elbrujohalcon:
Cool. If you get it running first, I'll cherry-pick your changes.

doc_rules/elvis_style/no_single_match_maybe.md
paulo-ferraz-oliveira, line 1: We need to review this later 😄

src/elvis_rulesets.erl
```diff
@@ -44,6 +44,7 @@ rules(hrl_files) ->
                         no_nested_try_catch,
                         no_operation_on_same_value,
                         no_single_clause_case,
+                        no_single_match_maybe,
```
paulo-ferraz-oliveira, line 47: Oh, default, I like this...

paulo-ferraz-oliveira (approved):
I especially like how the implementation of new rules read so well. Good stuff here.

doc_rules/elvis_style/no_single_match_maybe.md
paulo-ferraz-oliveira: Maybe now that I finished my PR you can sync. the format?

elbrujohalcon: I will, indeed.

.github/workflows/ci.yml
```diff
@@ -45,9 +45,13 @@ jobs:
       - name: Format check
         if: ${{ matrix.os == 'ubuntu-24.04' && matrix.otp_vsn == '26' }}
         run: rebar3 fmt --check
-      - name: Run test
+      - name: Run test Erlang/OTP 25
```
bormilan, line 48: I think it's a mistake by me, first I made two different test steps that's why I renamed it, but the final version contains only one as before.

PR #442, opened 2025-05-30
Propose a few changes to (the codebase's) readability and shared elements

paulo-ferraz-oliveira:
# Description

The goal of this pull request is to try to improve the codebase alone (**edit**: eventually there was a fix around `behaviour`, for found incompleteness). I don't intend on fixing any particular issue with it, other than increasing developer comfort (hopefully making rule implementation even simpler, if only marginally).

I ended up with quite a few changes, and if e.g. this isn't accepted in a similar format, I can always roll them back locally and check what minor changes I did, since I believe some of them will stick around.

This is also an experiment in ergonomics, to see how best we can ease writing new rules.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1842,9 +1842,9 @@ no_catch_expressions(Config, Target, RuleConfig) ->
     [elvis_result:item()].
 no_single_clause_case(Config, Target, RuleConfig) ->
     Root = get_root(Config, Target, RuleConfig),
-    IsSingleClauseCaseStatement = fun (Node) ->
+    IsSingleClauseCaseStatement = fun(Node) ->
         ktn_code:type(Node) =:= 'case' andalso length(case_clauses_in(Node)) =:= 1
-        end,
+    end,
     CaseNodes = elvis_code:find(IsSingleClauseCaseStatement, Root),
     lists:map(
         fun(CaseNode) ->
@@ -1872,7 +1872,7 @@ case_clauses_in(Node) ->
     [elvis_result:item()].
 no_single_match_maybe(Config, Target, RuleConfig) ->
     Root = get_root(Config, Target, RuleConfig),
-    IsSingleMatchMaybeStatement = fun (Node) ->
+    IsSingleMatchMaybeStatement = fun(Node) ->
         ktn_code:type(Node) =:= 'maybe' andalso length(ktn_code:content(Node)) =:= 1
     end,
     CaseNodes = elvis_code:find(IsSingleMatchMaybeStatement, Root),
@@ -1956,12 +1956,12 @@ numeric_format(Config, Target, RuleConfig) ->
 behaviour_spelling(Config, Target, RuleConfig) ->
     Spelling = option(spelling, RuleConfig, behaviour_spelling),
     Root = get_root(Config, Target, RuleConfig),
-    Predicate =
+    IsWronglySpelledBehaviour =
         fun(Node) ->
             (ktn_code:type(Node) =:= behaviour orelse ktn_code:type(Node) =:= behavior) andalso
                 ktn_code:type(Node) =/= Spelling
         end,
-    case elvis_code:find(Predicate, Root) of
+    case elvis_code:find(IsWronglySpelledBehaviour, Root) of
         [] ->
             [];
         InconsistentBehaviorNodes ->
@@ -2040,8 +2040,8 @@ is_function_or_fun(Zipper) ->
 consistent_generic_type(Config, Target, RuleConfig) ->
     TypePreference = option(preferred_type, RuleConfig, consistent_generic_type),
     Root = get_root(Config, Target, RuleConfig),
-    Predicate = consistent_generic_type_predicate(TypePreference),
-    case elvis_code:find(Predicate, Root, #{traverse => all, mode => node}) of
+    IsInconsistentGenType = consistent_generic_type_predicate(TypePreference),
+    case elvis_code:find(IsInconsistentGenType, Root, #{traverse => all, mode => node}) of
         [] ->
             [];
         InconsistentTypeNodes ->
@@ -2058,14 +2058,14 @@ consistent_generic_type(Config, Target, RuleConfig) ->
 always_shortcircuit(Config, Target, RuleConfig) ->
     Operators = #{'and' => 'andalso', 'or' => 'orelse'},
     Root = get_root(Config, Target, RuleConfig),
-    Predicate =
+    IsBadOperator =
         fun(Node) ->
             is_operator_node(Node) andalso
                 lists:member(
                     ktn_code:attr(operation, Node), maps:keys(Operators)
                 )
         end,
-    case elvis_code:find(Predicate, Root, #{traverse => all}) of
+    case elvis_code:find(IsBadOperator, Root, #{traverse => all}) of
         [] ->
             [];
         BadOperators ->


src/elvis_ktn.erl
paulo-ferraz-oliveira, line 1: I hesitated when naming `function`, `module` and `pattern`, since these target `node_attr` not `attr`...

paulo-ferraz-oliveira, line 1: I hesitated when naming function, module and pattern, since these target node_attr not attr...

src/elvis_style.erl
```diff
@@ -1492,9 +1481,8 @@ node_line_limits(FunctionNode) ->
     [elvis_result:item()].
 no_nested_try_catch(Config, Target, RuleConfig) ->
     Root = get_root(Config, Target, RuleConfig),
-    Predicate = fun(Node) -> ktn_code:type(Node) == 'try' end,
     ResultFun = result_node_line_fun(?NO_NESTED_TRY_CATCH),
-    case elvis_code:find(Predicate, Root) of
+    case elvis_code:find(fun is_try_expr/1, Root) of
```
paulo-ferraz-oliveira, line 1485: I personally prefer the simplicity/elegance of the predicate inline, like this, but lemme know...

src/elvis_ktn.erl
paulo-ferraz-oliveira, line 1: I hesitated when naming `function`, `module` and `pattern`, since these target `node_attr` not `attr`...

src/elvis_style.erl
```diff
@@ -1022,11 +1006,11 @@ no_behavior_info(Config, Target, RuleConfig) ->
 
     FilterFun =
         fun(Node) ->
-            case ktn_code:type(Node) of
-                function ->
-                    Name = ktn_code:attr(name, Node),
+            case is_function_node(Node) of
```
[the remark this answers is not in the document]
bormilan: Most of the time, I like this approach, but in this case, I feel like it's not worth making a separate function only for this simple expression. And I prefer the old version for readability reasons, because case clauses like 'function' feel more appropriate, and easier to understand than a boolean.

src/elvis_style.erl
bormilan: It looks better, the only thing I like more in the old one is that way the ' is_try_expr'  will be living only inside the function that uses it.

src/elvis_style.erl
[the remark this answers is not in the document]
bormilan: I like this one ☝️

src/elvis_style.erl
paulo-ferraz-oliveira: Correct, but then:

1. if it's reused code, I moved it (and this expression is being used in other places - so "don't repeat yourself")
2. it wasn't consistent to start with: some functions were outside their scope, some inside their scope
3. if I wanna write a new rule, it might be easier if I have the code inside the function, sure, but I'll also be double-writing some of it, if that makes sense...

elbrujohalcon: Yeah, that's why I think we'll be better of with alternatives to `elvis_code:find/2` that do not require a function to begin with…
When the predicate is complex, yes… by all means, let's keep `find/2`, but if we're just looking for nodes with a particular type… let's add `find_by_type/2` and just pass the type down. Same for `find_by_name/2` and any other simple function like that which can be used everywhere in this module. I wouldn't go too far down that line… maybe the farthest I would go would be `find_by_types([atom(), ...], node()) -> [node()].`, which would receive a list of types and would return everything that has one of those types.

paulo-ferraz-oliveira:
I'm removing `elvis_ktn` and moving the code back to "where it was". I'll also put the predicates inside each function: I'm afraid this'll mean we'll sometimes repeat ourselves, but maybe we can easily live with it... as-was...

src/elvis_style.erl
```diff
@@ -610,7 +610,7 @@ check_variable_casing_consistency({_, [#{name := FirstName, var := FirstVar} | O
         [] ->
             [];
         OtherNames ->
-            {Line, _} = ktn_code:attr(location, FirstVar),
+            {Line, _Col} = ktn_code:attr(location, FirstVar),
```
paulo-ferraz-oliveira, line 613: When I reverted, stuff like this happened (since I was consistently reverting) that unused variables get a name; lemme know if you prefer I revert it to not pollute the PR.

src/elvis_style.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: This change's rationale: we're only interested in functions, so why don't we ask "is this a function?", directly, and go the route "what is this? oh, it's a function", instead? Trying to understand the consistency with the code, overall...

paulo-ferraz-oliveira:
I've reduced the scope a bit (by getting rid of `elvis_ktn` and removing smaller changes like `_` > `_Col`). I'll get on the `case is_...` elements, so the scope will be even further reduced.

src/elvis_style.erl
```diff
@@ -766,7 +766,10 @@ no_nested_hrls(ElvisConfig, RuleTarget, RuleConfig) ->
     ).
 
 is_include(Node) ->
-    ktn_code:type(Node) =:= include orelse ktn_code:type(Node) =:= include_lib.
+    (ktn_code:type(Node) =:= include) orelse is_include_lib_attr(Node).
+
+is_include_lib_attr(Node) ->
```
paulo-ferraz-oliveira, line 771: Used 2 times.

src/elvis_style.erl
```diff
@@ -1040,6 +1048,9 @@ no_behavior_info(Config, Target, RuleConfig) ->
             lists:map(ResultFun, BehaviorInfos)
     end.
 
+is_function_node(Node) ->
```
paulo-ferraz-oliveira, line 1051: Used 5 times.

src/elvis_style.erl
```diff
@@ -1260,6 +1270,12 @@ max_anonymous_function_arity(Config, Target, RuleConfig) ->
         Funs
     ).
 
+is_fun(Node) ->
```
paulo-ferraz-oliveira, line 1273: Used 2 times.

src/elvis_style.erl
```diff
@@ -1609,6 +1624,9 @@ is_relevant_behaviour(Root, RuleConfig) ->
         )
     ).
 
+is_behaviour_attr(Node) ->
```
paulo-ferraz-oliveira, line 1623: Used 3 times. Actually helped find that `is_relevant_behaviour` was incomplete.

src/elvis_style.erl
```diff
@@ -2972,10 +2993,10 @@ wildcard_match(X, Y) ->
 check_nested_try_catchs(ResultFun, TryExp) ->
     lists:filtermap(
         fun(Node) ->
-            case ktn_code:type(Node) == 'try' of
-                true ->
```
paulo-ferraz-oliveira, line 2976: Changed because it would be inconsistent with a previous comment.

src/elvis_ktn.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: No longer relevant.

src/elvis_code.erl
```diff
@@ -135,7 +135,7 @@ find_by_location(Root, Location) ->
     end.
 
 is_at_location(#{attrs := #{location := {Line, NodeCol}}} = Node, {Line, Column}) ->
-    Text = ktn_code:attr(text, Node),
+    Text = elvis_ktn:text(Node),
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Module was removed since, so I think we'll continue the discussion in another comment, or the main PR body...

src/elvis_style.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Reverted.

elbrujohalcon:
> > If you don't mind, let's try with elvis_code:find_by_type/2, elvis_code:find_by_name/2, etc.… for the common scenarios where the predicates are just checking that something is of a specific type or has a specific name (or maybe one in a list of names or types).
> 
> You mean a function that actually does the search, and not just a wrapper around `ktn_code`, right?
> 
> I can try that, now that the code has reverted to some of its previous sanity.
> 
> And this basically means "exploding" `elvis_code:find` into something more specific, right?

Yeah… something like…

```erlang
find_by_type(Type, Node) ->
    find(fun(Node) -> ktn_code:type(Node) =:= Type end, Node).
```

src/elvis_style.erl
elbrujohalcon: Same as above… this may end up being just `elvis_code:find_by_types([behavior, behaviour], Node).` in those 3 places. Or a helper that instead of `is_behaviour_attr` would be `find_behaviour_attrs(Node)`.

And, for completeness, in `elvis_code`…

```erlang
find_by_types(Types, Node) ->
    find(fun(Node) -> lists:member(ktn_code:type(Node), Types) end, Node).
```

src/elvis_style.erl
```diff
@@ -2662,7 +2669,7 @@ is_dynamic_call(Node) ->
             case ktn_code:type(FunctionSpec) of
                 remote ->
                     ModuleName = ktn_code:node_attr(module, FunctionSpec),
-                    var == ktn_code:type(ModuleName);
+                    ktn_code:type(ModuleName) =:= var;
```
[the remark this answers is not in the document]
elbrujohalcon: This is good.

src/elvis_style.erl
elbrujohalcon: Excellent.

src/elvis_style.erl
```diff
@@ -766,7 +766,10 @@ no_nested_hrls(ElvisConfig, RuleTarget, RuleConfig) ->
     ).
 
 is_include(Node) ->
-    ktn_code:type(Node) =:= include orelse ktn_code:type(Node) =:= include_lib.
+    ktn_code:type(Node) =:= include orelse is_include_lib_attr(Node).
```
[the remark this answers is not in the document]
bormilan: Oh, I like this `find_by_any_type` idea

paulo-ferraz-oliveira:
I pushed an alternative on simplification so we can kinda sync. (because I'm convinced even if we have less code, at the end, this might be a somewhat big change - in terms of LoC).

src/elvis_style.erl
```diff
@@ -703,8 +703,7 @@ check_no_macro_calls(Calls) ->
 no_macros(ElvisConfig, RuleTarget, RuleConfig) ->
     TreeRootNode = get_root(ElvisConfig, RuleTarget, RuleConfig),
```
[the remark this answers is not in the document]
elbrujohalcon: Humm… suddenly those seem to be too many params, but… let's give it a try and see how it ends up.

src/elvis_code.erl
```diff
@@ -134,6 +142,14 @@ find_by_location(Root, Location) ->
             {ok, Node}
     end.
 
+find_by_types(Names, Root) ->
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I propose we use it where it's easy to introduce, and leave further refactoring for the future (if we have time), since the goal is to find a good place to allow writing rules faster, not just make it pretty "better" for the sake of it 😄

src/elvis_style.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'm going with this. Resolving since this'll get outdated and the scope will change.

paulo-ferraz-oliveira:
I also wanted to get rid of all the repetitions of `, #{traverse => all, mode => node}` and with this new approach I am 😄 (maybe not all, but a good bunch)

src/elvis_style.erl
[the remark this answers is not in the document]
elbrujohalcon: Cool. Yeah… Let's do it.

src/elvis_style.erl
```diff
@@ -633,13 +633,24 @@ variable_naming_convention(Config, Target, RuleConfig) ->
...
+        case ktn_code:type(Node) of
+            {atom, [_, _], define} ->
+                true;
+            _ ->
+                false
+        end
+    end,
     MacroNodes =
-        elvis_code:find(fun is_macro_define_node/1, Root, #{traverse => all, mode => node}),
+        elvis_code:find(Predicate, Root, #{traverse => all, mode => node}),
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Sure, I can tweak those at the end...

elbrujohalcon (approved):
This is looking great now! 🤩

elbrujohalcon:
I ❤️ negative PRs!!!

bormilan (approved):

paulo-ferraz-oliveira:
⚠️ apart from https://github.com/inaka/elvis_core/pull/442#discussion_r2116000755 (I'm going out now) everything should be in a reviewable state, at this moment.

We'll only know if we can reap what we're sowing on the next rules 😄

elbrujohalcon (approved):
Merge at will, @paulo-ferraz-oliveira

elbrujohalcon:
I'm merging this one.

PR #446, opened 2025-06-06
Fix `operator_spaces` for [`=`, `=>`, `:=`, `<-`, `<=`], while adding `?=` to it, too

### LINKED ISSUE #443: {left, "="} and {right, "="} operator_spaces rules do not work
## Bug Description

`{left, "="}` and  `{right, "="}` `operator_spaces` rules do not work

## To Reproduce

It doesn't work even in the test suite. Notice there's no `#{info := [left, "=" | _]}` nor `#{info := [right, "=" | _]}` item in the pattren list at [verify_operator_spaces](https://github.com/inaka/elvis_core/blob/4.0.0/test/style_SUITE.erl#L478) test case. 

However, it should be reported cause the issue is present in the analysed [sample code](https://github.com/inaka/elvis_core/blob/4.0.0/test/examples/fail_operator_spaces.erl#L135). 

According to the [documentation](https://github.com/inaka/elvis_core/blob/4.0.0/doc_rules/elvis_style/operator_spaces.md#options) `{right, "="}` and `{left, "="}` rules are included by default. I also checked the code, and these rules are indeed [included by default](https://github.com/inaka/elvis_core/blob/4.0.0/src/elvis_style.erl#L175)

## Expected Behavior

missing spaces around `=` operator are reported properly.

## `rebar3` Logs

not relevant

## Additional Context

- It's declared in the documentation, but it doesn't work.
- Moreover, the dedicated test case explicitly verifies that it doesn't work

paulo-ferraz-oliveira:
# Description

Closes #443. Adds `?=` to the rule.


doc_rules/elvis_style/operator_spaces.md
```diff
@@ -41,6 +41,8 @@ compromised without adequate spacing.
 
 `{left, "!"}, {right, "!"}` was added in [4.0.0](https://github.com/inaka/elvis_core/releases/tag/4.0.0)).
 
+`{left, "?="}, {right, "?="}` was added in [4.1.0](https://github.com/inaka/elvis_core/releases/tag/4.1.0)).
```
paulo-ferraz-oliveira, line 44: Not sure 4.1.0 is the correct version, but we can tweak if we're thinking of releasing a patch for this (though adding `?=` makes it a feature, too) 😄

elbrujohalcon: 4.1.0 is what I would use, indeed.

paulo-ferraz-oliveira:
Don't merge, as there's still something wrong with this. CI is Ok, but the test is not validating what it should...

src/elvis_style.erl
```diff
@@ -1915,11 +1924,11 @@ is_match_in_condition(Node) ->
...
 
+is_match(Node) ->
+    lists:member(ktn_code:type(Node), match_operators()).
+
 has_match_child(Node) ->
-    IsMatch = fun(InnerNode) ->
-        ktn_code:type(InnerNode) =:= match orelse ktn_code:type(InnerNode) =:= maybe_match
-    end,
-    lists:any(IsMatch, ktn_code:content(Node)).
+    lists:any(fun is_match/1, ktn_code:content(Node)).
```
paulo-ferraz-oliveira, line 1931: Simplified because of reuse.

src/elvis_style.erl
```diff
@@ -2654,12 +2663,9 @@ check_parent_match_or_macro(Zipper) ->
...
             Parent = zipper:node(ParentZipper),
-            case ktn_code:type(Parent) of
-                match ->
-                    zipper:down(ParentZipper) =:= Zipper;
-                macro ->
-                    zipper:down(ParentZipper) =:= Zipper;
-                maybe_match ->
+            IsMatchOrMacro = lists:member(ktn_code:type(Parent), [macro | match_operators()]),
+            case IsMatchOrMacro of
+                true ->
```
paulo-ferraz-oliveira, line 2668: Simplified because of reuse, but lemme know...

PR #447, opened 2025-06-09
Simplify `elvis_code`

paulo-ferraz-oliveira:
# Description

I try to simplify `elvis_code` a bit:

1. I move to their corresponding modules code that's not that generic
2. in trying to make `get_root` more generic I found the abstraction wasn't worth the result, but found a refactoring opportunity for some default options
3. for commits "Simplify ..." we actually increase the code, because we duplicate it in corresponding rules, but this also seems to make for more independent rule code (lemme know and I can make a generic function in `elvis_style`)

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1671,9 +1671,7 @@ no_boolean_in_comparison(Config, Target, RuleConfig) ->
 
     IsBoolean =
         fun(Node) ->
-            lists:member(
-                ktn_code:attr(value, Node), [true, false]
-            )
+            is_boolean(ktn_code:attr(value, Node))
         end,
 
     IsComparisonWithBoolean =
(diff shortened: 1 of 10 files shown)


src/elvis_style.erl
```diff
@@ -1583,15 +1589,12 @@ is_relevant_behaviour(Root, RuleConfig) ->
     ConfigBehaviors = option(behaviours, RuleConfig, no_init_lists),
     Behaviours = elvis_code:find_by_types([behaviour, behavior], Root),
     lists:any(
-        fun(Elem) -> Elem end,
```
paulo-ferraz-oliveira, line 1586: 🤔... > ...😆

src/elvis_style.erl
```diff
@@ -1668,9 +1671,7 @@ no_boolean_in_comparison(Config, Target, RuleConfig) ->
 
     IsBoolean =
         fun(Node) ->
-            lists:member(
-                ktn_code:attr(value, Node), [true, false]
-            )
+            is_boolean(ktn_code:attr(value, Node))
```
paulo-ferraz-oliveira, line 1674: 🤔

elbrujohalcon (approved):
Nice!!!

PR #448, opened 2025-06-09
Fix `no_boolean_in_comparison`: adds [`=:=`, `/=`, `=/=`]

### LINKED ISSUE #441: bug in `no_boolean_in_comparison`?
Maybe not a bug, but opening it as such, since I don't know the complete details behind the rule's implementation.

Is `no_boolean_in_comparison` supposed to also match on `=:=`?

paulo-ferraz-oliveira:
# Description

This is for the `no_boolean_in_comparison` bug. I think we should also be comparing with `/=` and `=/=`, so I add those.

Closes #441.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -2497,7 +2497,7 @@ check_spaces(Lines, UnfilteredNodes, {Position, Text}, Encoding, {How0, _} = How
 
 %% @private
 maybe_run_regex(undefined = _Regex, _Line) ->
-    false;
+    undefined;
 maybe_run_regex({ok, Regex}, Line) ->
     re:run(Line, Regex).
 
@@ -2524,7 +2524,7 @@ character_at_location(
     TextRegex =
         case TextRegexes of
             [] ->
-                false;
+                undefined;
             _ ->
                 maybe_run_regex(proplists:get_value(Text, TextRegexes), Line)
         end,
@@ -2554,7 +2554,7 @@ character_at_location(
             "";
         _ when How =:= should_have ->
             lists:nth(ColToCheck, TextLineStr);
-        _ when How =:= should_not_have, not TextRegex; TextRegex =:= nomatch ->
+        _ when How =:= should_not_have, TextRegex =:= undefined; TextRegex =:= nomatch ->
             lists:nth(ColToCheck, TextLineStr);
         _ ->
             ""
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -2554,7 +2554,7 @@ character_at_location(
             "";
         _ when How =:= should_have ->
             lists:nth(ColToCheck, TextLineStr);
-        _ when How =:= should_not_have, TextRegex =:= false; TextRegex =:= nomatch ->
+        _ when How =:= should_not_have, not TextRegex; TextRegex =:= nomatch ->
             lists:nth(ColToCheck, TextLineStr);
         _ ->
             ""
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1677,7 +1677,7 @@ no_boolean_in_comparison(Config, Target, RuleConfig) ->
         fun(Node) ->
             Content = ktn_code:content(Node),
             ktn_code:type(Node) =:= op andalso
-                ktn_code:attr(operation, Node) =:= '==' andalso
+                lists:member(ktn_code:attr(operation, Node), ['==', '=:=', '/=', '=/=']) andalso
                 lists:any(IsBoolean, Content)
         end,
     ComparisonsWithBoolean =
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -2497,7 +2497,7 @@ check_spaces(Lines, UnfilteredNodes, {Position, Text}, Encoding, {How0, _} = How
 
 %% @private
 maybe_run_regex(undefined = _Regex, _Line) ->
-    undefined;
+    nomatch;
 maybe_run_regex({ok, Regex}, Line) ->
     re:run(Line, Regex).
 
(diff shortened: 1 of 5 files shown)


src/elvis_style.erl
```diff
@@ -2554,7 +2554,7 @@ character_at_location(
             "";
         _ when How =:= should_have ->
             lists:nth(ColToCheck, TextLineStr);
-        _ when How =:= should_not_have, TextRegex =:= false; TextRegex =:= nomatch ->
+        _ when How =:= should_not_have, TextRegex =:= undefined; TextRegex =:= nomatch ->
```
paulo-ferraz-oliveira, line 2557: 👍 (we were using a boolean when maybe we shouldn't, since `nomatch` - the other possible result - isn't one!)

elbrujohalcon: Maybe instead of `undefined`, we could use `no_regex` or something _clearer_.
Or we could be super-sneaky and use `nomatch` 🫣

src/elvis_style.erl
```diff
@@ -2552,9 +2552,7 @@ character_at_location(
             SpaceChar;
         {_, right, LenLine} when How =:= should_not_have, ColToCheck > LenLine ->
             "";
-        _ when How =:= should_have ->
-            lists:nth(ColToCheck, TextLineStr);
-        _ when How =:= should_not_have, TextRegex =:= false; TextRegex =:= nomatch ->
+        _ when How =:= should_have; TextRegex =:= nomatch ->
```
paulo-ferraz-oliveira, line 2555: This seems sneaky, as mentioned, but works.

elbrujohalcon: I like it.

filmor:
I think this one is a step back. When I write `Val =:= true`, I use this as a shorthand for `is_boolean(Val) andalso Val`. In general, IMO a `=:=` and `=/=` implies a type check, as that is what distinguishes it from `==` and `/=`.

elbrujohalcon:
@filmor Let me see if I understand you point. You're dealing with functions like this one?
```erlang
-spec your_function(boolean() | map()) -> is_true | is_something_else.
your_function(Val) when Val =:= true -> is_true;
your_funciton(_) -> is_something_else.
```
(i.e., a function that does something if it receives `true` and something else in any other case, even if what it receives is not even a boolean)

If that's the case, I would just write…

```erlang
-spec your_function(boolean() | map()) -> is_true | is_something_else.
your_function(true) -> is_true;
your_funciton(_) -> is_something_else.
```

Am I missing something here?

filmor:
This is an actual diff I have now:

```erlang
 -spec is_active() -> boolean().
 is_active() ->
     try
-        etp_config:get_value([?SECTION, <<"active">>]) =:= true
+        case etp_config:get_value([?SECTION, <<"active">>]) of
+            true ->
+                true;
+            _ ->
+                false
+        end
     catch
         _:_ ->
             true
     end.
```
The data comes out of a configuration file.

I'm not saying that this is a huge issue, and I'm aware that there are workarounds. What I try to get across is that `=:=` and `=/=` are semantically not simply comparisons but also type checks.

elbrujohalcon:
Hummm… I disagree on the type-checking nature of `=:=`. I don't think the community as a whole sees it that way.

In any case, I would write that code like…
```erlang
 -spec is_active() -> boolean().
 is_active() ->
     try etp_config:get_value([?SECTION, <<"active">>]) of
        true -> true;
        _ -> false
     catch
         _:_ ->
             true %% BTW, shouldn't this be 'false' ? (o_O)
     end.
```

Yeah, it is larger than the original… but it also moves `=:= true` _outside_ the `try` part and with that you can be sure that the only possible error/exception (if any) will come from `etp_config:get_value/1`. Yeah, I'm aware that `=:=` can't fail, but… in your original code it was part of the `try` anyway.

In any case, what changed now is the addition of `=:=` to the rule. Your code would've raised the warning anyway if you would've written `etp_config:get_value([?SECTION, <<"active">>]) == true`, which is (semantically) the same thing since there is no type-casting for booleans. Or is that what I'm missing here? Does Erlang do some magic type-casting for booleans? Is `0 == true`? I think not… right?

As far as I know, `==` matches everything exactly like `=:=` except for numbers. But maybe I'm wrong.

elbrujohalcon:
Well… [The docs](https://www.erlang.org/doc/system/expressions.html#term-comparisons) are not _super-explicit_ about it, but I think it's reasonable to infer that `==` is different than `=:=` only for numbers…

> The term equivalence operators, =:= and =/=, return whether two terms are indistinguishable. While the other operators consider the same numbers equal even when their types differ (1 == 1.0 is true), the term equivalence operators return whether or not there exists a way to tell the arguments apart.
>
> For example, while the terms 0 and 0.0 represent the same number, we can tell them apart by using the [is_integer/1](https://www.erlang.org/doc/apps/erts/erlang#is_integer/1) function. Hence, =:= and =/= consider them different.
>
> Furthermore, the terms 0.0 and -0.0 also represent the same number, but they yield different results when converted to string form through [float_to_list/1](https://www.erlang.org/doc/apps/erts/erlang#float_to_list/1): when given the former it returns a string without a sign, and when given the latter it returns a string with a sign. Therefore, =:= and =/= consider them different.
> 
> The term equivalence operators are useful when reasoning about terms as opaque values, for example in associative containers or memoized functions where using the equal-to operator (==) risks producing incorrect results as a consequence of mixing up numbers of different types.

kivra-pauoli:
We should improve our documentation, for sure. I also ran into this and believe that doing `is_boolean(Val) andalso Val` (we were doing `fun(V) -> V =:= true end` adds unwarranted implicitness, meaning you have to reason on why that is (on the other hand if you make your code explicit you work toward one of the rules' goals - less implicitness). There's maybe a case to have a list of `true`, `false` where we're looking at the value of the atom, and not its truthiness, but that's an edge case, I believe, since that atom is mostly used for actual truthiness.

PR #450, opened 2025-06-09
Add `{gen_statem, call, 2}`, `{gen_server, call, 2}`, and `{gen_event, call, 3}` to option `caveat_functions` in rule `no_common_caveats_call`

### LINKED ISSUE #408: Missing implementation details for `no_common_caveats` (code not as per doc.)
## Bug Description

This pull requests, https://github.com/inaka/elvis_core/pull/364, seems to have resulted in some wishful thinking, but not a proper implementation. It's missing some of the documented elements, where it shouldn't.

paulo-ferraz-oliveira:
# Description

Add the missing implementation details. As is, it seems to me the test code is rather brittle:

1. I don't know how easy it is to find that reference (that I changed from `6` to `9`)
2. the fact that we were able to accept an update that didn't actually add to the rules is kinda proof of brittleness

Closes #408.

### DIFF
--- a/doc_rules/elvis_style/no_common_caveats_call.md
+++ b/doc_rules/elvis_style/no_common_caveats_call.md
@@ -28,6 +28,9 @@ results in clearer, faster, and more maintainable code.
              , {gen_event, call, 3}
               ]`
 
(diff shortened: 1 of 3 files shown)


elbrujohalcon (approved):

elbrujohalcon:
I agree with your opinion on test brittleness, but I don't mind that much.

PR #453, opened 2025-06-09
New rulesets: `hrl_files_strict` and `beam_files_strict`

paulo-ferraz-oliveira:
# Description

We add two new prebuilt rulesets (for stricter analysis) and take this opportunity to refactor `elvis_rulesets` in a way it's hopefully easier to navigate and maintain.

We also fix a few issues we found in code inconsistencies along the way (or from the fact that we reviewed most of our rules and some issues surfaced).


config/elvis-test-hrl-files.config
```diff
@@ -3,7 +3,7 @@
         {config, [
             #{
                 dirs => ["../../_build/test/lib/elvis_core/test/examples/"],
-                filter => "test-*.hrl",
+                filter => "test_*.hrl",
```
paulo-ferraz-oliveira, line 6: Found by `module_naming_convention` now also applying to `.hrl`.

src/elvis_gitignore.erl
```diff
@@ -1,6 +1,6 @@
 -module(elvis_gitignore).
 
--export([required_patterns/3, forbidden_patterns/3]).
+-export([required_patterns/3, forbidden_patterns/3, default/1]).
```
paulo-ferraz-oliveira, line 3: Required because of the way we're now going to all rulesets consistently.

src/elvis_text_style.erl
```diff
@@ -53,7 +53,9 @@ default(no_tabs) ->
 default(no_trailing_whitespace) ->
     #{ignore_empty_lines => false};
 default(no_redundant_blank_lines) ->
-    #{max_lines => 1}.
+    #{max_lines => 1};
+default(prefer_unquoted_atoms) ->
```
paulo-ferraz-oliveira, line 57: Found by the fact we're not consistently getting the defaults for rules.

paulo-ferraz-oliveira:
I'm leaving the `prefer_unquoted_atoms` stuff to a different pull request. This one should be good to go.

src/elvis_rulesets.erl
```diff
@@ -202,3 +55,196 @@ ensure_clean_table() ->
...
+                param_pattern_matching,
+                private_data_types,
+                state_record_and_type
+            ]
+    ].
+
+erl_files_rules() ->
+    elvis_style_rules() ++ elvis_text_style_rules().
+
+elvis_style_rules() ->
```
bormilan, line 115: For this list, we might name the rule functions in `elvis_style` to all end with `_rule` or something like that, and that way we could dynamically pull them here. Therefore, we don't need to add them here manually; we only care about manually adding the rules if they are special. (For example, don't work on beam files)

src/elvis_rulesets.erl
```diff
@@ -202,3 +55,196 @@ ensure_clean_table() ->
...
+                numeric_format,
+                operator_spaces,
+                param_pattern_matching,
+                private_data_types,
+                used_ignored_variable,
+                variable_naming_convention
+            ]
+    ].
+
+elvis_text_style_rules() ->
```
bormilan, line 162: Same here.

src/elvis_rulesets.erl
```diff
@@ -13,186 +13,39 @@ set_rulesets(RuleSets) ->
...
+            _ ->
+                try
+                    ets:lookup_element(?MODULE, Group, 2)
+                catch
+                    error:badarg ->
+                        []
+                end
+        end,
     lists:map(
         fun({Mod, Rule}) -> {Mod, Rule, apply(Mod, default, [Rule])} end,
```
elbrujohalcon, line 46: Why don't you make this a callback (i.e., `-callback`) and use `-behaviour(…)` in all those modules so you can call `Mod:default(Rule)` here.
If that's not possible for some reason, please use this instead…
```suggestion
        fun({Mod, Rule}) -> {Mod, Rule, erlang:apply(Mod, default, [Rule])} end,
```
or put a `-elvis(…).` on top with a nice comment on why we can't use callbacks here.

src/elvis_rulesets.erl
```diff
@@ -202,3 +55,196 @@ ensure_clean_table() ->
...
+
+gitignore_rules() ->
+    [
+        {elvis_gitignore, Rule}
+     || Rule <-
+            [
+                forbidden_patterns,
+                required_patterns
+            ]
+    ].
```
elbrujohalcon, line 67:
```suggestion
    [{elvis_gitignore, forbidden_patterns}, {elvis_gitignore, required_patterns}].
```

elbrujohalcon: It's… simpler :)

src/elvis_rulesets.erl
```diff
@@ -202,3 +55,196 @@ ensure_clean_table() ->
...
+hrl_only_files_rules() ->
+    [
+        {elvis_style, Rule}
+     || Rule <-
+            [
+                no_nested_hrls,
+                no_specs,
+                no_types
+            ]
+    ].
```
elbrujohalcon, line 81:
```suggestion
hrl_only_files_rules() ->
    [
        {elvis_style, no_nested_hrls},
        {elvis_style, no_specs},
        {elvis_style, no_types}
    ].
```

src/elvis_style.erl
```diff
@@ -1049,8 +1049,16 @@ module_naming_convention(Config, Target, RuleConfig) ->
...
 
     Root = get_root(Config, Target, RuleConfig),
-    [Module] = elvis_code:find_by_types([module], Root),
-    ModuleName = ktn_code:attr(value, Module),
+    ModuleName =
+        case elvis_code:find_by_types([module], Root) of
+            [Module] ->
+                ktn_code:attr(value, Module);
+            _ ->
+                % .hrl, maybe?
```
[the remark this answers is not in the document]
elbrujohalcon: Are we applying a rule for `-module(…).` attributes to hrls?
I mean… if you can make it work, that's fine with me… but I would say that the behaviour of applying such a rule to those files should be entirely unpredictable. Who would put a `-module(…).` attribute on an hrl file 😱 ?

bormilan (commented):
I'm honestly confused about the new rulesets module.  At first, I thought it was super cool, but when I started thinking about what I would do if I needed to add a new rule, I started feeling that it's not too easy to understand.

elbrujohalcon (changes requested):
🤦🏻 I meant to request changes

src/elvis_style.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: We're not 😄. Conceptually, the rule is about the module name (`All modules should conform to the pattern defined by the regex option pattern`): https://github.com/inaka/elvis_core/blob/main/doc_rules/elvis_style/module_naming_convention.md.

It's just that for `.erl` we decided to read from the `-module` attribute, but we could read both from the filename and it'd be just as good, right?

src/elvis_rulesets.erl
paulo-ferraz-oliveira: I think the behaviour route is a good one. I just don't have enough rules yet that this surfaces 😄 but we soon will... Thanks.

src/elvis_rulesets.erl
paulo-ferraz-oliveira: Same stuff, yup; will act on it.

src/elvis_style.erl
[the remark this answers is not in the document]
elbrujohalcon: In any case, I'm fine with your change. Let's keep it.

PR #455, opened 2025-06-09
Fix `prefer_unquoted_atoms` (match better)

### LINKED ISSUE #454: Revisit `prefer_unquoted_atoms`
## Bug Description

These are getting flagged, for example:

```
    - Atom "'=='" on line 458 is quoted but quotes are not needed.
    - Atom "'/='" on line 459 is quoted but quotes are not needed.
    - Atom "'=<'" on line 460 is quoted but quotes are not needed.
    - Atom "'<'" on line 461 is quoted but quotes are not needed.
    - Atom "'>='" on line 462 is quoted but quotes are not needed.
    - Atom "'>'" on line 463 is quoted but quotes are not needed.
    - Atom "'=:='" on line 464 is quoted but quotes are not needed.
    - Atom "'=/='" on line 465 is quoted but quotes are not needed.
    - Atom "'='" on line 468 is quoted but quotes are not needed.
    - Atom "'--'" on line 469 is quoted but quotes are not needed.
```

(from `elvis_style`)

## To Reproduce

On the `elvis_core` root folder: `rebar3 escriptize ; _build/default/bin/elvis_core`

## Expected Behavior

Only flag an atom for "quotes are not required" if that's actually the case.

## `rebar3` Logs

Not applicable.

## Additional Context

Not applicable.

paulo-ferraz-oliveira:
# Description

We were using a simple list of atoms to avoid to know if we should quote or not.

Not we're relying on Erlang/OTP's definition of atom (via a regex) and `erl_scan:f_reserved_word`, as discussed elsewhere, to know if we can actually unquote.

Closes #454.

### DIFF
--- a/src/elvis_text_style.erl
+++ b/src/elvis_text_style.erl
@@ -120,11 +120,11 @@ prefer_unquoted_atoms(_Config, Target, _RuleConfig) ->
     AtomNodes = elvis_code:find_by_types([atom], Tree, #{traverse => all, mode => node}),
     check_atom_quotes(AtomNodes, []).
 
-needs_quoting(AtomStr) ->
-    case re:run(AtomStr, "^'[a-z][a-zA-Z0-9_@]*'$", [{capture, none}]) of
+needs_quoting(AtomName0) ->
+    case re:run(AtomName0, "^'[a-z][a-zA-Z0-9_@]*'$", [{capture, none}]) of
         match ->
-            AtomAsString = string:trim(AtomStr, both, "'"),
-            Atom = list_to_atom(AtomAsString),
+            AtomName = string:trim(AtomName0, both, "'"),
+            Atom = list_to_atom(AtomName),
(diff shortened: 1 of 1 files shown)


src/elvis_style.erl
```diff
@@ -2733,7 +2733,7 @@ has_state_type(Root) ->
 spec_includes_record(Node) ->
     IsTypeRecord =
         fun(Child) ->
-            (ktn_code:type(Child) =:= 'type') andalso (ktn_code:attr(name, Child) =:= record)
+            (ktn_code:type(Child) =:= type) andalso (ktn_code:attr(name, Child) =:= record)
```
paulo-ferraz-oliveira, line 2736: Finding stuff already 😄

bormilan (approved):

paulo-ferraz-oliveira:
👍 

for me this is Ok, and since it's in a `strict` ruleset it'll be less "visible". Merging...

PR #456, opened 2025-06-09
Lint `elvis_core` further

paulo-ferraz-oliveira:
# Description

I start by revising `elvis.config` to use `strict` rulesets (we can always disable stuff, in the future, but at least this way we're exercising more rules) and act on my findings.

### DIFF
--- a/src/elvis_file.erl
+++ b/src/elvis_file.erl
@@ -190,11 +190,11 @@ find_encoding(Content) ->
     Ignore :: [elvis_style:ignorable()],
     Res :: file().
 maybe_add_abstract_parse_tree(
-    #{ruleset := beam_files},
+    #{ruleset := Ruleset},
     #{path := Path} = File,
     Mod,
     Ignore
-) ->
+) when Ruleset =:= beam_files; Ruleset =:= beam_files_strict ->
     AbstractParseTree = get_abstract_parse_tree(Path, Mod, Ignore),
     File#{abstract_parse_tree => AbstractParseTree};
 maybe_add_abstract_parse_tree(_Config, File, _Mod, _Ignore) ->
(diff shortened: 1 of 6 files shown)


src/elvis_file.erl
```diff
@@ -190,11 +190,11 @@ find_encoding(Content) ->
     Ignore :: [elvis_style:ignorable()],
     Res :: file().
 maybe_add_abstract_parse_tree(
-    #{ruleset := beam_files},
+    #{ruleset := Ruleset},
     #{path := Path} = File,
     Mod,
     Ignore
-) ->
+) when Ruleset =:= beam_files; Ruleset =:= beam_files_strict ->
```
paulo-ferraz-oliveira, line 197: This comes from testing further.

src/elvis_gitignore.erl
```diff
@@ -69,7 +69,7 @@ forbidden_patterns(_Config, #{path := Path}, RuleConfig) ->
 
 %% .gitignore
 check_patterns_in_lines(_Lines, [], Results, _Mode) ->
-    {ok, Results};
+    Results;
```
paulo-ferraz-oliveira, line 72: This comes from testing further.

src/elvis_style.erl
```diff
@@ -1044,9 +1044,10 @@ module_naming_convention(Config, Target, RuleConfig) ->
             [Module] ->
                 ktn_code:attr(value, Module);
             _ ->
-                % .hrl, maybe?
+                % .hrl, maybe? or .beam?
```
paulo-ferraz-oliveira, line 1047: I didn't want to change the `.md` promise, and I thought there might be someone only looking at `.beam` so if we don't need to over-avoid it, why not (?)

PR #457, opened 2025-06-09
Allow for OTP behaviours to bypass `export_used_types`

### LINKED ISSUE #308: There are instances where not exporting a type is valid, even if it's used in a spec.

paulo-ferraz-oliveira:
# Description

Closes #308.

Lemme know if this seems like enough, or if e.g. we want a new option to the rule to bypass for other behaviours (user-defined)? It can always be disabled or ignored per module, so by default ignoring what we do now (with this PR) might be enough.

Also, should we have a follow-up task for when dynamic callbacks land?

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1094,7 +1094,7 @@ is_forbidden_module_name(Target, Regex, Message) ->
     [elvis_result:item()].
 state_record_and_type(Config, Target, RuleConfig) ->
     Root = get_root(Config, Target, RuleConfig),
-    case is_otp_module(Root) of
+    case is_otp_behaviour(Root) of
         true ->
             case {has_state_record(Root), has_state_type(Root)} of
                 {true, true} ->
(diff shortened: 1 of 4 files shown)


elbrujohalcon (approved):
I'm not a super-fan of this, but I've to deal with it in the past and the current solution of exporting the `state/0` type is not wonderful either. So, I'm approving.

PR #464, opened 2025-06-11
Add `list_to_atom/1` and `binary_to_atom/1,2` to rule `no_common_caveats_call`'s option `caveat_functions` (default)

### LINKED ISSUE #410: Add `list_to_atom` and `binary_to_atom` to "Common Caveats"?
### Discussed in https://github.com/inaka/elvis_core/discussions/405

<div type='discussions-op-text'>

<sup>Originally posted by **paulo-ferraz-oliveira** May 18, 2025</sup>
Ref.: https://www.erlang.org/doc/system/commoncaveats.html#list_to_atom-1

It's avoidable, if the consumer wants it, but it could be considered harmful by default.</div>

paulo-ferraz-oliveira:
# Description

Closes #410.


src/elvis_style.erl
```diff
@@ -214,8 +214,7 @@
 -define(NO_CALL_MSG, "The call to ~p:~p/~p on line ~p is in the no_call list.").
 -define(NO_DEBUG_CALL_MSG, "Remove the debug call to ~p:~p/~p on line ~p.").
 -define(NO_COMMON_CAVEATS_CALL_MSG,
-    "The call to ~p:~p/~p on line ~p is in the list of "
-    "Erlang Efficiency Guide common caveats."
+    "The call to ~p:~p/~p on line ~p might have performance drawbacks or implicit behavior."
```
paulo-ferraz-oliveira, line 217: I tweaked it since our doc. no longer refers simply to that element and we've thus broadened the scope of our analysis.

elvis.config
```diff
@@ -18,7 +18,10 @@
                         {elvis_style, max_function_length, disable},
                         {elvis_style, max_function_clause_length, disable},
                         {elvis_style, no_macros, disable},
-                        {elvis_style, max_module_length, #{ignore => [elvis_style]}}
+                        {elvis_style, max_module_length, #{ignore => [elvis_style]}},
+                        {elvis_style, no_common_caveats_call, #{
+                            ignore => [{elvis_file, module, 1}, {elvis_style, is_call, 2}]
+                        }}
```
paulo-ferraz-oliveira, line 24: Another solution to avoid these would be to use `string` as the starting point and then do `atom_to_list` comparisons, but I thought it would be Ok to avoid these two cases...

src/elvis_text_style.erl
```diff
@@ -113,7 +113,7 @@ needs_quoting(AtomName0) ->
     case re:run(AtomName0, "^'[a-z][a-zA-Z0-9_@]*'$", [{capture, none}]) of
         match ->
             AtomName = string:trim(AtomName0, both, "'"),
-            Atom = list_to_atom(AtomName),
+            Atom = list_to_existing_atom(AtomName),
```
paulo-ferraz-oliveira, line 116: Fix from dogfooding...

src/elvis_text_style.erl
```diff
@@ -265,7 +265,7 @@ check_no_trailing_whitespace(Line, Num, IgnoreEmptyLines) ->
             no_result;
         {match, [PosLen]} ->
             Msg = ?NO_TRAILING_WHITESPACE_MSG,
-            Info = [Num, size(binary:part(Line, PosLen))],
+            Info = [Num, byte_size(binary:part(Line, PosLen))],
```
paulo-ferraz-oliveira, line 268: Marginally faster? Hurrah for dogfooding (and the `erlang:` choice, which surfaced this)

src/elvis_style.erl
```diff
@@ -2873,8 +2875,24 @@ is_in_call_list(Call, DisallowedFuns) ->
...
 call_mfa(Call) ->
     FunctionSpec = ktn_code:node_attr(function, Call),
-    M = ktn_code:attr(value, ktn_code:node_attr(module, FunctionSpec)),
-    F = ktn_code:attr(value, ktn_code:node_attr(function, FunctionSpec)),
+    M0 = ktn_code:attr(value, ktn_code:node_attr(module, FunctionSpec)),
+    M =
+        case M0 of
+            undefined ->
+                % this is a bare call, e.g. list_to_atom/1; assume 'erlang'
+                erlang;
```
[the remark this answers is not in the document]
elbrujohalcon: Yeah… I agree this is fine.

src/elvis_style.erl
```diff
@@ -2873,8 +2875,24 @@ is_in_call_list(Call, DisallowedFuns) ->
...
+        end,
+    F0 = ktn_code:attr(value, ktn_code:node_attr(function, FunctionSpec)),
+    F =
+        case F0 of
+            undefined ->
+                % this is what we get for local (or erlang:) calls
+                ktn_code:attr(value, FunctionSpec);
+            _ ->
+                F0
+        end,
```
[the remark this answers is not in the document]
elbrujohalcon: Once again, I think this is fine.

src/elvis_text_style.erl
elbrujohalcon: Hummm… This is coming from the regex matching. Are you sure we don't need to, at least, add a try...catch here for atoms that we don't _already know_ ?

elbrujohalcon (commented):
I'm a bit skeptical of the usage of `list_to_existing_atom/1` without catching the possible errors that it may produce.

src/elvis_text_style.erl
paulo-ferraz-oliveira: I mean, maybe (?) The fact that we got here is because we did find an atom in the code (which is now loaded) and we're just manipulating it as a string. For safety, I don't mind moving it to as-was.

paulo-ferraz-oliveira: Changes pushed.

PR #472, opened 2025-06-12
New rule: `no_receive_without_timeout`

### LINKED ISSUE #426: New Rule: `receive_with_timeout`
### Name

`receive_with_timeout`

### Brief Description

All `receive` expressions should be accompanied by an `after` expressions.

### Problematic code

```erlang
receive
    onhook ->
        disconnect(),
        idle();
    {connect, B} ->
        B ! {busy, self()},
        wait_for_onhook()
end
```

### Correct code

```erlang
receive
    onhook ->
        disconnect(),
        idle();
    {connect, B} ->
        B ! {busy, self()},
        wait_for_onhook()
after
    60_000 ->
        exit(nothing_received)
end
```

### Rationale

A `receive` block without a timeout will wait indefinitely if no matching message arrives; by making your timeout explicit you:

* avoid hanging processes
* improve testability (deterministic behaviour under test failure conditions)
* ease debug and recovery
* can implement retry and self-healing
* potentially avoid denial-of-service scenarios where waits are exploited

elbrujohalcon:
# Description

New rule: `no_receive_without_timeout`

Closes #426.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -55,7 +55,8 @@
     no_init_lists/3,
     ms_transform_included/3,
     no_boolean_in_comparison/3,
-    no_operation_on_same_value/3
+    no_operation_on_same_value/3,
+    no_receive_without_timeout/3
 ]).
 
 -export_type([empty_rule_config/0]).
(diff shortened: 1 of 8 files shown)


src/elvis_style.erl
```diff
@@ -1636,6 +1641,32 @@ no_boolean_in_comparison(Config, Target, RuleConfig) ->
...
+            Info = [Line],
+            Msg = ?NO_RECEIVE_WITHOUT_TIMEOUT,
+            elvis_result:new(item, Msg, Info, Line)
+        end,
+
+    lists:map(ResultFun, ReceivesWithoutTimeout).
+
+is_receive_without_timeout(Receive) ->
+    [] == elvis_code:find_by_types([receive_after], Receive).
+
```
elbrujohalcon, line 1669: Look at this code, @paulo-ferraz-oliveira / @bormilan !! It's so… so… simple! It's super-cool 🤩 

I guess… the next step would be to simplify the `ResultFun` (particularly when the only argument for the format string is the line number, like in this case)… If we ever do that, writing new rules would become so easy that even some AI will be able to do it. 🤯

PR #475, opened 2025-06-15
Propose a more consistent output message format (and hopefully simpler coding of future rules)

paulo-ferraz-oliveira:
# Description

**Note**: opening as **Draft**, for discussion.

This pull request proposes:

- increased consistency in message format output
  - it'll also allow us to move forward in non-verbose output like `<file>:<line>:<col>`
- move messages close to where they're used
  - less macros, too (actually, no more macros in `elvis_core`)
  - better more self-contained rules
  - an `io:format/2` -like approach to message and arguments
    - `check_no_call` is the exception to the rule, here, as we're using it in 3 different places with a somewhat complex (?) code surrounding them
      - we could consider moving this code block back to each of the rules' definitions, since it's not a lot and it'd increase their independence
- approaching a generic results' fun, as discussed previously

We got to fix a few tests where we were pattern matching not on `line_num`, but `info`.

# Further considerations

I'm also still gonna:

- [x] make sure all messages have as much information as possible (at least line, column, and threshold)
- [x] review all messages, for consistency
- [ ] ~simplify `ResultFun` implementation to ease rules' repetition~ (postponed to a subsequent PR)


elvis.config
```diff
@@ -17,7 +17,6 @@
                         {elvis_style, no_throw, disable},
                         {elvis_style, max_function_length, disable},
                         {elvis_style, max_function_clause_length, disable},
-                        {elvis_style, no_macros, disable},
```
paulo-ferraz-oliveira, line 20: 👋

src/elvis_config.erl
```diff
@@ -22,8 +22,6 @@
 -type config() :: map().
 -type configs() :: [config()].
 
--define(DEFAULT_FILTER, "*.erl").
```
paulo-ferraz-oliveira, line 25: 👋, single use

src/elvis_core.erl
```diff
@@ -225,7 +225,7 @@ apply_rule({Module, Function, ConfigArgs}, {Result, Config, File}) ->
             end
         catch
             _:Reason:Stacktrace ->
-                Msg = "'~p' while applying rule '~p': ~p.",
+                Msg = "'~p' while applying rule '~p': ~p",
```
paulo-ferraz-oliveira, line 228: `.` is now added automatically at the end of each message.

src/elvis_project.erl
paulo-ferraz-oliveira, line 1: DEP_BRANCH and DEP_NO_GIT are applied to two different rules. We just made it in a single function, before, to save a few bytes, but I think the current approach makes it easier to follow.

src/elvis_result.erl
```diff
@@ -66,10 +82,34 @@ new(warn, Msg, Info) ->
 
 -spec new(item, string(), [term()], integer()) -> item().
 new(item, Msg, Info, LineNum) ->
+    new_item(Msg, Info, #{line => LineNum}).
+
+new(item, Msg0, Info, {Line, Column}, Limit) ->
```
paulo-ferraz-oliveira, line 87: This function is where it all started, basically.

src/elvis_text_style.erl
```diff
@@ -147,8 +138,12 @@ no_redundant_blank_lines(_Config, Target, RuleConfig) ->
     ResultFun =
         fun
             ({Line, BlankLinesLength}) when BlankLinesLength >= MaxLines ->
-                Info = [Line, BlankLinesLength, MaxLines],
-                {true, elvis_result:new(item, ?NO_REDUNDANT_BLANK_LINES_MSG, Info, Line)};
+                {true,
+                    elvis_result:new_item(
+                        "there are too many blank lines",
+                        [],
```
paulo-ferraz-oliveira, line 144: I'm simplify these occurrences with some `elvis_result` smartness.

src/elvis_style.erl
```diff
@@ -548,39 +371,44 @@ function_naming_convention(Config, Target, RuleConfig) ->
     ForbiddenRegex = option(forbidden_regex, RuleConfig, function_naming_convention),
     Root = get_root(Config, Target, RuleConfig),
     Functions = elvis_code:find_by_types([function], Root),
-    FunctionNames0 = lists:map(fun(Node) -> ktn_code:attr(name, Node) end, Functions),
```
paulo-ferraz-oliveira, line 551: We need the nodes, now, for `location` extraction.

src/elvis_style.erl
```diff
@@ -808,7 +671,7 @@ no_space_after_pound(Config, Target, RuleConfig) ->
 -type operator_spaces_config() ::
     #{ignore => [ignorable()], rules => [{right | left, string()}]}.
 
--define(PUNCTUATION_SYMBOLS, [',', ';', dot, '->', ':', '::', '|', '||']).
+punctuation_symbols() -> [',', ';', dot, '->', ':', '::', '|', '||'].
```
paulo-ferraz-oliveira, line 673: Single use macro is now gone. I was under the impression `hank` would recognise these, but maybe it only does so if they're issued from a `.hrl`. (even if they're not strictly dead code, in that case)

src/elvis_style.erl
```diff
@@ -925,11 +792,17 @@ god_modules(Config, Target, RuleConfig) ->
     [elvis_result:item()].
 no_if_expression(Config, Target, RuleConfig) ->
     Root = get_root(Config, Target, RuleConfig),
-    ResultFun = result_node_line_fun(?NO_IF_EXPRESSION_MSG),
```
paulo-ferraz-oliveira, line 928: This'll be next step, to find a pattern for these. (maybe I'll even use the same that was there)

src/elvis_result.erl
```diff
@@ -66,10 +82,34 @@ new(warn, Msg, Info) ->
...
+    Msg =
+        case Limit of
+            -1 ->
+                Msg2;
+            _ ->
+                Msg2 ++ " (limit: " ++ integer_to_list(Limit) ++ ")"
+        end,
     #{
-        message => Msg,
+        message => Msg ++ ".",
```
paulo-ferraz-oliveira, line 130: Shall we be smarter and only add `.` if not present?

src/elvis_style.erl
```diff
@@ -1113,14 +1007,15 @@ dont_repeat_yourself(Config, Target, RuleConfig) ->
             ({Line, Col}, "") ->
                 io_lib:format("(~p, ~p)", [Line, Col]);
             ({Line, Col}, Str) ->
-                io_lib:format("~s, (~p, ~p)", [Str, Line, Col])
+                io_lib:format("~p, (~p, ~p)", [Str, Line, Col])
```
paulo-ferraz-oliveira, line 1010: For testing, locally, I put everything to ~p, but I don't know if we should use more specialised elements (my goal was to move fast and not think about conversions and crashes from those). (?) Lemme know.

src/elvis_style.erl
```diff
@@ -1529,7 +1439,7 @@ no_init_lists(Config, Target, RuleConfig) ->
                     [Init1Fun] ->
                         Content = ktn_code:content(Init1Fun),
                         ListAttrClauses =
-                            lists:filtermap(fun(X) -> filter_list_clause_location(X) end, Content),
+                            lists:filtermap(fun filter_list_clause/1, Content),
```
paulo-ferraz-oliveira, line 1442: Future Elvis will catch these, hopefully.

src/elvis_style.erl
paulo-ferraz-oliveira: ... and this type of change will most likely be found elsewhere, too, in this PR.

src/elvis_style.erl
```diff
@@ -2251,7 +2208,7 @@ check_atom_names(_Regex, _, _RegexEnclosed, _, [] = _AtomNodes, Acc) ->
     Acc;
 check_atom_names(
     Regex,
-    ForbiddenRegexNormal,
+    ForbiddenRegex,
```
paulo-ferraz-oliveira, line 2211: Removing `Normal` for consistency. Found while revamping the code below.

src/elvis_gitignore.erl
```diff
@@ -83,9 +80,21 @@ check_patterns_in_lines(Lines, [Pattern | Rest], Results0, Mode) ->
             true ->
                 Results0;
             false when Mode =:= required ->
-                [elvis_result:new(item, ?REQUIRED_PATTERN, [Pattern]) | Results0];
+                [
```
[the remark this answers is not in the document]
elbrujohalcon: it's the same to me, as long as it's consistent.

src/elvis_style.erl
elbrujohalcon: I believe that `Str` is the accumulator here, so it **needs** to use `~s` or maybe `~ts`. Otherwise you'll end up with `"\"\\\"(1, 2)\\\", (3, 4)\", (5, 6)"`… when we want `"(1, 2), (3, 4), (5, 6)"`.

src/elvis_style.erl
```diff
@@ -1542,10 +1447,11 @@ no_init_lists(Config, Target, RuleConfig) ->
         end,
 
     ResultFun =
-        fun(Location) ->
-            Info = [Location],
-            Msg = ?NO_INIT_LISTS_MSG,
-            elvis_result:new(item, Msg, Info, Location)
+        fun(Node) ->
+            elvis_result:new_item(
+                "an unexpected 'init' callback list argument was found",
```
elbrujohalcon, line 1452:
```suggestion
                "using a list when implementing the 'init' callback should be avoided",
```

test/style_SUITE.erl
```diff
@@ -610,7 +501,7 @@ verify_no_block_expressions(Config) ->
 
     Path = "fail_no_block_expressions." ++ Ext,
 
-    [#{info := [9]}] =
+    [#{line_num := 9}] =
```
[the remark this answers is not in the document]
elbrujohalcon: Excellent

test/style_SUITE.erl
```diff
@@ -8,17 +8,12 @@
     init_per_suite/1,
     end_per_suite/1,
     init_per_group/2,
-    end_per_group/2
+    end_per_group/2,
+    elvis_core_apply_rule/5
```
elbrujohalcon, line 12: Can we move this function to another, utility, module… instead of exporting it?
Like… `elvis_test` or `elvis_test_utils`?

elbrujohalcon:
Also… since we're adding consistent result formatting now (with line, column, etc…) you should add one or more tests to verify that it works as expected. Maybe defining a custom rule that sometimes generates results with line, column, limit, etc… or not and checking that the generated results match the generic format you're adding to this PR.

src/elvis_style.erl
paulo-ferraz-oliveira: Oh, yeah, I meant to change these only in the elvis_result calls. Fail. Thanks for the review; I'll get on it.

src/elvis_project.erl
```diff
@@ -50,7 +36,27 @@ protocol_for_deps(_Config, Target, RuleConfig) ->
...
+                        elvis_result:new_item(
+                            "Dependency '~s' is not using appropriate protocol. "
+                            "Please change this to something like '~s'",
+                            [AppName, Regex]
+                        )
+                    ]
+            end
+        end,
         BadDeps
     ).
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Yeah, maybe; I was caught up with the initial refactoring, so didn't notice it. I'll check what can be improved.

src/elvis_result.erl
```diff
@@ -66,10 +102,34 @@ new(warn, Msg, Info) ->
...
+                Msg0;
+            _ ->
+                "At line " ++ integer_to_list(Line) ++ ", " ++ Msg0
+        end,
+    Msg2 =
+        case Column > 0 of
+            false ->
+                Msg1;
+            _ ->
+                ", column " ++ integer_to_list(Column) ++ ", " ++ Msg1
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Yeah, maybe, I still haven't looked at the printed results; I'm fixing the messages themselves. Let's keep this open for later.

src/elvis_style.erl
```diff
@@ -925,11 +790,16 @@ god_modules(Config, Target, RuleConfig) ->
...
         [] ->
             [];
         IfExprs ->
+            ResultFun = fun(Node) ->
+                elvis_result:new_item(
+                    "an unexpected 'if' expression was found",
+                    #{node => Node}
+                )
+            end,
             lists:map(ResultFun, IfExprs)
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'm still working on finding a pattern to apply to the results. But it'll be something along the lines of what you propose here.

test/style_SUITE.erl
paulo-ferraz-oliveira: Sure.

src/elvis_result.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Pushed a change for this.

elbrujohalcon:
> @elbrujohalcon, as per the pull request's description I still haven't pushed all changes regarding message consistency or the results_fun, so you can hold the review on those. 👍

/me holds on…

test/style_SUITE.erl
paulo-ferraz-oliveira: Changes pushed.

paulo-ferraz-oliveira:
I've now finished the review of the messages and think I'm only missing the result_fun generic thing. I'll act on comments.

src/elvis_project.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: ... or maybe you meant doing it in two steps: first recover the appname, then filter it out? Given there's not a whole lot of `rebar.config` files it should be doable.

src/elvis_style.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'm gonna start with the accumulation function, that can mostly be replaced by `filtermap`, I believe.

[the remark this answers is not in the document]
paulo-ferraz-oliveira: Oh... now I with there was an easy way to detect using `foldl` when `map` could be used. So many places where we're adding to `Acc` for no reason... 😄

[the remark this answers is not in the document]
paulo-ferraz-oliveira: When we do a case before a `lists:` operation to not pass an empty list to it, I'm pretty sure we're not getting a big performance improvement.

So, something like 

```erlang
    case elvis_code:find(fun spec_includes_record/1, Root) of
        [] ->
            [];
        SpecNodes ->
            ResultFun = fun(Node) ->
                elvis_result:new_item(
                    "an unexpected record was found in a spec; prefer creating a type for it and "
                    "using that",
                    #{node => Node}
                )
            end,
            lists:map(ResultFun, SpecNodes)
    end.
```

Is basically a "hidden" single-clause statement.

paulo-ferraz-oliveira:
@elbrujohalcon, I'm looking at the code and wondering if it wouldn't make more sense to separate the ResultFun stuff into a new PR. This one's already quite big, and could become even more cumbersome...

src/elvis_style.erl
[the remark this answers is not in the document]
elbrujohalcon: Yeah, that case statement is useless.

paulo-ferraz-oliveira:
I still pushed a small change, but this is now ready for review...

paulo-ferraz-oliveira:
I forgot to add the test you asked for, but I'll not do it today. Feel free to keep it open so I return to this. Thanks.

src/elvis_style.erl
```diff
@@ -748,12 +589,13 @@ no_specs(ElvisConfig, RuleTarget, RuleConfig) ->
...
     Root = get_root(Config, Target, RuleConfig),
     BeginNodes = elvis_code:find_by_types_in_tokens(['begin'], Root),
-    lists:foldl(
-        fun(BeginNode, Acc) ->
-            {Line, _Col} = ktn_code:attr(location, BeginNode),
-            [elvis_result:new(item, ?NO_BLOCK_EXPRESSIONS_MSG, [Line], Line) | Acc]
+    lists:map(
+        fun(BeginNode) ->
+            elvis_result:new_item(
+                "an avoidable block expression ('begin'...'end') was found",
```
elbrujohalcon, line 595:
```suggestion
                "an avoidable block expression ('begin...end') was found",
```
…or…
```suggestion
                "an avoidable block expression (begin...end) was found",
```
I mean… these are not atoms, right?

src/elvis_style.erl
```diff
@@ -925,13 +770,14 @@ god_modules(Config, Target, RuleConfig) ->
...
-    case elvis_code:find_by_types(['if'], Root) of
-        [] ->
-            [];
-        IfExprs ->
-            lists:map(ResultFun, IfExprs)
-    end.
+    IfExprs = elvis_code:find_by_types(['if'], Root),
+    ResultFun = fun(Node) ->
+        elvis_result:new_item(
+            "an unexpected 'if' expression was found; prefer 'case' expressions",
```
elbrujohalcon, line 776: Oooof… this one is _tricky_ at best. Because the replacement for an `if` is not _necessarily_ a `case`.
Sometimes it might be a function call. Or sometimes it would be best to rethink the logic around the `if` and write it in an if-less way entirely.
It's worth noting that replacing an `if` with a `case` that has a clause for `true` and another one for `false` is… still ugly. It's better than the `if`, but… it's still likely not the best solution.

src/elvis_style.erl
```diff
@@ -1763,12 +1629,13 @@ no_throw(Config, Target, RuleConfig) ->
...
     Root = get_root(Config, Target, RuleConfig),
     ThrowNodes = elvis_code:find(Zipper, Root),
-    lists:foldl(
-        fun(ThrowNode, AccIn) ->
-            {Line, _} = ktn_code:attr(location, ThrowNode),
-            [elvis_result:new(item, ?NO_THROW_MSG, [Line], Line) | AccIn]
+    lists:map(
+        fun(ThrowNode) ->
+            elvis_result:new_item(
+                "an avoidable call to 'throw/1' was found",
```
elbrujohalcon, line 1635: Maybe add something like "If needed, prefer error or exit" ?

src/elvis_style.erl
```diff
@@ -1849,8 +1728,10 @@ no_single_clause_case(Config, Target, RuleConfig) ->
     CaseNodes = elvis_code:find(IsSingleClauseCaseStatement, Root),
     lists:map(
         fun(CaseNode) ->
-            {Line, _Col} = ktn_code:attr(location, CaseNode),
-            elvis_result:new(item, ?NO_SINGLE_CLAUSE_CASE_MSG, [Line], Line)
+            elvis_result:new_item(
+                "an avoidable single-clause 'case' statement was found",
```
elbrujohalcon, line 1732:
```suggestion
                "an avoidable single-clause 'case' expression was found",
```

src/elvis_style.erl
```diff
@@ -1879,8 +1760,10 @@ no_single_match_maybe(Config, Target, RuleConfig) ->
     CaseNodes = elvis_code:find(IsSingleMatchMaybeStatement, Root),
     lists:map(
         fun(CaseNode) ->
-            {Line, _Col} = ktn_code:attr(location, CaseNode),
-            elvis_result:new(item, ?NO_SINGLE_MATCH_MAYBE_MSG, [Line], Line)
+            elvis_result:new_item(
+                "an avoidable single-match 'maybe' statement was found",
```
elbrujohalcon, line 1764:
```suggestion
                "an avoidable single-match 'maybe' expression was found",
```

src/elvis_style.erl
```diff
@@ -1898,8 +1781,11 @@ no_match_in_condition(Config, Target, RuleConfig) ->
     CaseNodes = elvis_code:find(fun is_match_in_condition/1, Root),
     lists:map(
         fun(CaseNode) ->
-            {Line, _Col} = ktn_code:attr(location, CaseNode),
-            elvis_result:new(item, ?NO_MATCH_IN_CONDITION_MSG, [Line], Line)
+            elvis_result:new_item(
+                "an avoidable match condition in a 'case' statement was found; prefer matching "
```
elbrujohalcon, line 1785:
```suggestion
                "an avoidable match condition in a 'case' expression was found; prefer matching "
```

elbrujohalcon (changes requested):
I'll wait for the test, too :)

src/elvis_style.erl
paulo-ferraz-oliveira: I went with `doc_rules/` on this one, so I'll fix it there, too.

src/elvis_style.erl
paulo-ferraz-oliveira: The rule is actually named "No Single-Clause Case Statements", so we'll probably fix that there, too.

paulo-ferraz-oliveira: At a point in time we seem to have diverged titles from rule names. Shall I re-sync them?

This would make "No Single-Clause Case Statements" "No Single-Clause Case" which is explicit enough...

doc_rules/elvis_style/no_nested_hrls.md
```diff
@@ -1,4 +1,4 @@
-# No Types [![](https://img.shields.io/badge/since-4.1.0-blue)](https://github.com/inaka/elvis_core/releases/tag/4.1.0) ![](https://img.shields.io/badge/HRL--only-yes-magenta)
+# No Nested Header Files [![](https://img.shields.io/badge/since-4.1.0-blue)](https://github.com/inaka/elvis_core/releases/tag/4.1.0) ![](https://img.shields.io/badge/HRL--only-yes-magenta)
```
paulo-ferraz-oliveira, line 1: 😄

src/elvis_style.erl
paulo-ferraz-oliveira: They're not (so this was probably a "replace" operation, somehow). I wanted to emphasise when stuff's like 'if' or 'case', so the first suggestion is the one I'll take.

src/elvis_style.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_style.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_style.erl
paulo-ferraz-oliveira: Changes pushed.

doc_rules/elvis_style/god_modules.md
```diff
@@ -1,6 +1,6 @@
 # God Modules ![](https://img.shields.io/badge/BEAM-yes-orange)
 
-The number of functions in a module should be limited to a maximum threshold.
+The number of exported functions in a module should be limited to a maximum threshold.
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: And actually that's not entirely true, since we're not looking at the actual exported functions, but the export nodes (if you're using `compile(export_all)` the count is not the same). Should we also account for this case?

paulo-ferraz-oliveira:
Tests added, and in the meantime I also added `column_num` to the result, since I believe it'll be helpful for implementing the shorter machine-readable output.

RULES.md
```diff
@@ -58,8 +58,8 @@ Most, if not all, of the rules will present (opinionated) documentation sections
 - [No Operator With Same Values](doc_rules/elvis_style/no_operation_on_same_value.md)
 - [No Redundant Blank Lines](doc_rules/elvis_text_style/no_redundant_blank_lines.md)
 - [No `receive` Without Timeout](doc_rules/elvis_style/no_receive_without_timeout.md)
-- [No Single-Clause Case Statements](doc_rules/elvis_style/no_single_clause_case.md)
-- [No Single-Match Maybe Statements](doc_rules/elvis_style/no_single_match_maybe.md)
+- [No Single-Clause Case Expressions](doc_rules/elvis_style/no_single_clause_case.md)
+- [No Single-Match Maybe Blocks](doc_rules/elvis_style/no_single_match_maybe.md)
```
paulo-ferraz-oliveira, line 62: There ya go.

doc_rules/elvis_style/no_single_clause_case.md
```diff
@@ -22,7 +22,7 @@ do:something("else")
 Using a `case` expression with only one clause is unnecessary and reduces code clarity. It adds
 syntactic overhead without providing meaningful branching logic. In such cases, a let-style
 assignment or direct pattern matching is typically more appropriate and idiomatic. Removing
-single-clause case statements also improves readability and simplifies the control flow.
+single-clause case expressions also improves readability and simplifies the control flow.
```
paulo-ferraz-oliveira, line 25: There ya go.

doc_rules/elvis_style/no_single_match_maybe.md
```diff
@@ -19,7 +19,7 @@ end
 {ok, A} = do:something()
 ```
 
-Note that `maybe` expressions with an `else` are perfectly acceptable, too:
+Note that `maybe` blocks with an `else` are perfectly acceptable, too:
```
paulo-ferraz-oliveira: There ya go.

src/elvis_style.erl
```diff
@@ -1736,14 +1736,14 @@ no_catch_expressions(Config, Target, RuleConfig) ->
...
+    IsSingleClauseCaseExpression = fun(Node) ->
         ktn_code:type(Node) =:= 'case' andalso length(case_clauses_in(Node)) =:= 1
     end,
-    CaseNodes = elvis_code:find(IsSingleClauseCaseStatement, Root),
+    CaseNodes = elvis_code:find(IsSingleClauseCaseExpression, Root),
     lists:map(
         fun(CaseNode) ->
             elvis_result:new_item(
-                "an avoidable single-clause 'case' statement was found",
+                "an avoidable single-clause 'case' expression was found",
```
paulo-ferraz-oliveira: There ya go.

src/elvis_style.erl
```diff
@@ -1768,14 +1768,14 @@ case_clauses_in(Node) ->
...
+    IsSingleMatchMaybeBlock = fun(Node) ->
         ktn_code:type(Node) =:= 'maybe' andalso length(ktn_code:content(Node)) =:= 1
     end,
-    CaseNodes = elvis_code:find(IsSingleMatchMaybeStatement, Root),
+    CaseNodes = elvis_code:find(IsSingleMatchMaybeBlock, Root),
     lists:map(
         fun(CaseNode) ->
             elvis_result:new_item(
-                "an avoidable single-match 'maybe' statement was found",
+                "an avoidable single-match 'maybe' block was found",
```
paulo-ferraz-oliveira: There ya go.

src/elvis_style.erl
```diff
@@ -1796,7 +1796,7 @@ no_match_in_condition(Config, Target, RuleConfig) ->
     lists:map(
         fun(CaseNode) ->
             elvis_result:new_item(
-                "an avoidable match condition in a 'case' statement was found; prefer matching "
+                "an avoidable match condition in a 'case' expression was found; prefer matching "
```
paulo-ferraz-oliveira: There ya go.

src/elvis_style.erl
```diff
@@ -773,7 +773,7 @@ no_if_expression(Config, Target, RuleConfig) ->
     IfExprs = elvis_code:find_by_types(['if'], Root),
     ResultFun = fun(Node) ->
         elvis_result:new_item(
-            "an unexpected 'if' expression was found; prefer 'case' expressions",
+            "an unexpected 'if' expression was found",
```
paulo-ferraz-oliveira: There ya go.

elbrujohalcon (changes requested):
Just a tiny change in a message.

doc_rules/elvis_style/god_modules.md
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Cool, then we might need to adjust the documentation, just...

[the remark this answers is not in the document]
paulo-ferraz-oliveira: Change pushed. (for clarity)

src/elvis_style.erl
```diff
@@ -2137,13 +2024,17 @@ export_used_types_in(TreeRootNode) ->
...
+    Locations = map_type_declarations_to_location(TreeRootNode),
 
     % Report
     lists:map(
         fun({Name, Arity} = Info) ->
-            Line = maps:get(Info, LineNumbers, unknown),
-            elvis_result:new(item, ?EXPORT_USED_TYPES_MSG, [Name, Arity, Line], Line)
+            {Line, Column} = maps:get(Info, Locations, {-1, -1}),
+            elvis_result:new_item(
+                "type '~p/~p' is used by an exported function; prefer to also export the it",
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Cool beans. I wanted "export it", but that also works and is maybe clearer.

paulo-ferraz-oliveira:
It's applied (2912f50b66647904b4cdb6818f0e60f8ce945c48): merging.

paulo-ferraz-oliveira:
Or rather, updating branch and enabling auto-merge.

src/elvis_style.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: > I would ❤️ to see a function like…

This'd only work in the very generic case where the message doesn't take arguments (the `DataOrAttrs` portion or `elvis_result:new_item/2,3`).

If only we had string interpolation we could generate the resulting message more easily, but alas we'll also have to resort to a helper fun.

PR #489, opened 2025-06-16
Attempt to fix crash in export_used_types

filmor:
# Description

`export_used_types` crashes when a spec is defined that returns an atom, which in turn is taken from a macro:

```erlang
-define(TYPE_CONSTANT, type_constant).

-spec fun_using_type_constant() -> ?TYPE_CONSTANT.
fun_using_type_constant() ->
    ?TYPE_CONSTANT.
```


src/elvis_style.erl
```diff
@@ -2113,7 +2113,9 @@ export_used_types_in(TreeRootNode) ->
     ExportedSpecs =
         lists:filter(
             fun(#{attrs := #{arity := Arity, name := Name}}) ->
-                lists:member({Name, Arity}, ExportedFunctions)
+                    lists:member({Name, Arity}, ExportedFunctions);
+                (_) ->
```
paulo-ferraz-oliveira, line 2117: Is it possible to be more explicit here? Or do the results vary wildly?

elbrujohalcon:
And I would not be myself if I don't say this…

Why do people write code like this? I hardly see the benefit of using a macro just for an atom.
People told me in the past that it's a way to avoid typos (in the sense that the pre-processor will detect a typo in compile time vs. run-time), but… if you write tests that cover your code, they should catch any typo like this. And even more if you use this technique for specs! If you make a typo on a spec for a type that's a single atom, that's something that Dialyzer will 100% warn you about.

Yeah, I know I'm preaching to the choir here, but… I _physically_ can't see code like that and not feel sad about the poor devs that need to maintain it.

filmor:
Well, I'm myself the dev that wrote it and has to maintain it :)

The original code is a pubsub using `pg` and this atom is the scope. I prefer compile errors over tests.

PR #493, opened 2025-06-21
Add ruleset for tests(#435)

### LINKED ISSUE #435: Ruleset for tests

Some rules are not useful in a test module. For me, `dont_repeat_yourself` is a rule that I ignore most of the time.

I want to use a ruleset designed for test modules, named `erl_test_files`.

Just ignore it every time.

bormilan:
# Description

I just added the new ruleset, same as `erl_files`, except `dont_repeat_yourself`.

Closes #435.


elbrujohalcon (approved):

paulo-ferraz-oliveira (approved):
You should add a reference to `RULES.md`. Other than that, it seems Ok as a starting point.

paulo-ferraz-oliveira:
I'd probably also check how this could apply to our test suite (check `elvis.config`) because maybe there's simplification we can do (?)

paulo-ferraz-oliveira:
> We have no config section for test files, should I add one?

Yup 😄 Hopefully there won't be a lot of things to touch, but if there are leave it. Our tests are really crazy as ways to explore Elvis.

.tool-versions
paulo-ferraz-oliveira, line 1: I'm Ok with this addition, but if we're gonna have it we should make sure CI supports it. `setup-beam`, for example, supports it out-of-the-box. Then again it's an extra something to maintain, so I'm torn 😕

paulo-ferraz-oliveira:
The Windows with OTP 28 check is `rebar3` waiting for a fix; not on us. (https://github.com/erlang/rebar3/pull/2954)

---

> elvis_style, god_modules - we should make a higher default for the test ruleset

Or outright `disable` it. Pretty sure (famous last words) the most common pattern is to export all functions in test files. And then it allows you to do the trick we do 

```erlang
all() ->
    Exports = ?MODULE:module_info(exports),
    [F || {F, _} <- Exports, not lists:member(F, elvis_test_utils:excluded_funs_all())].
```

> elvis_style, no_debug_call - I think we should keep this one in the test ruleset too.

👍, as long as we make sure the functions we target are actually debug, e.g., `ct:` and friends

> elvis_style, variable_naming_convention - I would keep this one too

👍 

> elvis_style, no_throw - this could be ignored, I don't think it matters during testing

I'm Ok but I do think that generally you shouldn't use `throw` because there's better mechanism.

> elvis_style, atom_naming_convention - I don't know about this. In Elvis, we use a lot of "bad" atom names to test if Elvis fails on them or not, but in any other projects, this shouldn't be in use, I guess.

👍 I'd keep it in the ruleset and disable it as you did

---

I think we could/should go through all the rules, yes, but mind you that from now on every rule we add to `erl_files` we should also remember to either tweak or disable in `test`, so I'd prefer to have helper functions that'd ease it. e.g.,:

* `without_rules(Rules, Ruleset)` (removes rules by key)
* `updated_rules(RulesWithDefaults, Ruleset)` (updated rules with new defaults - dropping all of them from the original rule)
* `add_rules(Rules, Ruleset)` (adds rules by key)

test/elvis_SUITE.erl
```diff
@@ -245,6 +245,7 @@ rock_without_errors_has_no_output(_Config) ->
     %% and CT will capture it.
     %% Thus, we remove it from the list of captures before doing the actual check
     RemoveSearchPattern = "fail_non_parsable_file.erl",
+    ct:pal("Output=~p~n", [Output]),
```
paulo-ferraz-oliveira: I'd prefer to not have debug calls in our tests (mostly they'll pollute output for no reason), but I'm not sure this was intended.

src/elvis_ruleset.erl
```diff
@@ -219,3 +225,9 @@ elvis_config_rules() ->
     [
         {elvis_project, old_configuration_format}
     ].
+
+-spec without_rules(Exceptions, Rules) -> Rules when
+    Exceptions :: [{atom(), atom()}],
+    Rules :: [{atom(), atom()}].
+without_rules(Exceptions, Rules) ->
+    Rules -- Exceptions.
```
paulo-ferraz-oliveira: You can't do this in the most recent version of `elvis_ruleset`.

bormilan: Yes, I changed to use the 'trim' function.

test/elvis_SUITE.erl
bormilan: I'm thinking about removing all throws and debug printing from the tests. Now I just put it back to not change anything until I fix the checks on gh.

PR #500, opened 2025-06-23
Tweak: further simplification in the code base

paulo-ferraz-oliveira:
# Description

It started with me trying to find a pattern in the way we write rules, and adapting the code to simplify for that pattern.

The most generic way I could find to ease rule implementation is to have:

1. something to find inside of something (most commonly inside the root node)
2. something to filter results (as node or zipper)
3. something to output errors (as LC or `lists:*map`, in most functions)

Many typespecs were removed because many inconsistencies were found; and also I think default already documents them well enough.

There's also a potentially unnecessary `lists:reverse(` in `max_function_clause_length`, since we're already sorting the output (tests might need adjustment).

# Further considerations

This is not yet complete, but I'm pushing for early feedback.

At the moment of writing I'd removed around ~400+~ ~300 lines of code (specs included) while trying to keep it readable and maintainable.

Look at an example of a rule:

```erlang
no_nested_hrls(RuleCfg) ->
    IncludeNodes = elvis_code:find(#{
        of_types => [include, include_lib],
        inside => root(RuleCfg)
    }),

    [
        elvis_result:new_item(
            "unexpected nested '-include[_lib]' attribute ('~s') was found; "
            "avoid including .hrl files in .hrl files",
            [ktn_code:attr(value, IncludeNode)],
            #{node => IncludeNode}
        )
     || IncludeNode <- IncludeNodes
    ].
```

1. find `include` and `include_lib` nodes in the tree's root
2. print an error for each of the found nodes (we take the location from the node)

Another one:

```erlang
used_ignored_variable(RuleCfg) ->
    {zippers, IgnoredVarZippers} = elvis_code:find(#{
        of_types => [var],
        inside => root(RuleCfg),
        filtered_by => fun is_ignored_var/1,
        filtered_from => zipper
    }),

    [
        elvis_result:new_item(
            "an unexpected use of an ignored variable was found",
            #{zipper => IgnoredVarZipper}
        )
     || IgnoredVarZipper <- IgnoredVarZippers
    ].
```

1. find `var` zipper in the tree's root
2. filter them by "is ignored variable?"
3. print an error for each of the found zippers (we take the location from the zipper)

# Bugfixes

* `ignore` added to `no_branch_deps.md`
* `ignore` added to `protocol_for_deps.md`
* `allow` added to default for `no_macros`
* all `unicode`, regexp, and stuff made consistent via naming, functions and execution (and pre-compiled/reused where possible)
* `prefer_unquoted_atoms` gets location in results
* (not a bug) `option` now gets `?FUNCTION_NAME` as last argument, everywhere
* messages that refer numbers also output those (results changed in tests)
* moved a couple of comments around that formatting had put in wrong places
* removed `report` from `rebar.config`
* reviewed all `~...` formatters in messages: some where broken
* (not a bug) some functions are exported but not used; we're not using Xref to the fullest

### DIFF
--- a/doc_rules/elvis_project/no_branch_deps.md
+++ b/doc_rules/elvis_project/no_branch_deps.md
@@ -29,7 +29,7 @@ branch heads.
 
 ## Options
 
-- `ignore :: [string()]`
+- `ignore :: [atom()]`
   - default: `[]`
 
 ## Example configuration
--- a/elvis.config
+++ b/elvis.config
@@ -22,7 +22,7 @@
                             ignore => [
                                 {elvis_file, module, 1},
                                 {elvis_style, is_call, 2},
-                                {elvis_style, no_macros, 1},
+                                {elvis_style, is_allowed_macro, 2},
                                 {elvis_text_style, doesnt_need_quotes, 1}
                             ]
                         }}
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -23,18 +23,18 @@
 
 -export_type([find_options/0, tree_node/0, tree_node_type/0, tree_node_zipper/0]).
 
--spec find(Options) -> [NodeOrZipper] when
+-spec find(Options) -> [Node] | {zippers, [Zipper]} when
     Options :: #{
         % undefined means "all types"
         of_types := [tree_node_type()] | undefined,
         inside := Node,
         % undefined means "don't filter"
-        filtered_by => fun((NodeOrZipper) -> boolean()),
+        filtered_by => fun((Node | Zipper) -> boolean()),
         filtered_from => node | zipper,
         traverse => content | all
     },
-    NodeOrZipper :: Node | tree_node_zipper(),
-    Node :: tree_node().
+    Node :: tree_node(),
+    Zipper :: tree_node_zipper().
 find(Options) ->
     OfTypes = maps:get(of_types, Options),
     Inside = maps:get(inside, Options),
@@ -61,11 +61,19 @@ find(Options) ->
         Traverse
     ),
 
-    case FilteredBy of
-        undefined ->
-            NonFilteredResults;
-        _ ->
-            [Result || Result <- NonFilteredResults, FilteredBy(Result)]
+    Results =
+        case FilteredBy of
+            undefined ->
+                NonFilteredResults;
+            _ ->
+                [Result || Result <- NonFilteredResults, FilteredBy(Result)]
+        end,
+
+    case FilteredFrom of
+        node ->
+            Results;
+        zipper ->
+            {zippers, Results}
     end.
 
 %% @doc Find all nodes in the tree for which the predicate function returns
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -35,9 +35,7 @@
     },
     Node :: tree_node(),
     Zipper :: tree_node_zipper().
-find(Options) ->
-    OfTypes = maps:get(of_types, Options),
-    Inside = maps:get(inside, Options),
+find(#{of_types := OfTypes, inside := Inside} = Options) ->
     FilteredBy = maps:get(filtered_by, Options, undefined),
     FilteredFrom = maps:get(filtered_from, Options, node),
     Traverse = maps:get(traverse, Options, content),
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -103,13 +103,10 @@ zipper(Root) ->
     zipper(Root, content).
 
 -spec zipper(tree_node(), content | all) -> tree_node_zipper().
-zipper(Root, Traverse) ->
-    case Traverse of
-        content ->
-            content_zipper(Root);
-        all ->
-            all_zipper(Root)
-    end.
+zipper(Root, content) ->
+    content_zipper(Root);
+zipper(Root, all) ->
+    all_zipper(Root).
 
 -spec content_zipper(tree_node()) -> tree_node_zipper().
 content_zipper(Root) ->
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -3,16 +3,7 @@
 %% General
 -export([
     find/1,
-    find/2,
-    find/3,
-    find_by_location/2,
-    find_by_types/2,
-    find_by_types/3,
-    find_by_types/4,
-    find_by_types_in_tokens/2,
-    find_token/2,
-    code_zipper/1,
-    code_zipper/2
+    code_zipper/1
 ]).
 %% Specific
 -export([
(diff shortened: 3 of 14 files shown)


doc_rules/elvis_project/no_branch_deps.md
```diff
@@ -29,10 +29,11 @@ branch heads.
 
 ## Options
 
-- None.
+- `ignore :: [string()]`
+  - default: `[]`
```
paulo-ferraz-oliveira: This might've been simplified in the past, because `ignore` (the option) is named the same way as `ignore` the module avoider...

elvis.config
paulo-ferraz-oliveira, line 1: Not a lot changed here, or in the tests. I wanted to maintain that as the base references for the change, while keeping sanity.

src/elvis_code.erl
```diff
@@ -2,43 +2,85 @@
...
-    find/3,
-    find_by_location/2,
-    find_by_types/2,
-    find_by_types/3,
-    find_by_types_in_tokens/2,
-    find_token/2,
-    code_zipper/1,
-    code_zipper/2
+    find/1,
+    code_zipper/1
```
paulo-ferraz-oliveira, line 6: I hesitated about moving `code_zipper` to `elvis_style` since that's the only place it's used from. Here, though, we have just a single function to get all stuff. It hopefully covers all use cases and is not overly complex.

doc_rules/elvis_project/no_branch_deps.md
```diff
@@ -29,10 +29,11 @@ branch heads.
 
 ## Options
 
-- None.
+- `ignore :: [atom()]`
+  - default: `[]`
```
paulo-ferraz-oliveira, line 33: This might've been simplified in the past, because `ignore` (the option) is named the same way as `ignore` the module avoider...

src/elvis_code.erl
```diff
@@ -2,43 +2,85 @@
...
+-export_type([find_options/0, tree_node/0, tree_node_type/0, tree_node_zipper/0]).
+
+-spec find(Options) -> [Node] | {zippers, [Zipper]} when
+    Options :: #{
+        % undefined means "all types"
+        of_types := [tree_node_type()] | undefined,
+        inside := Node,
+        % undefined means "don't filter"
+        filtered_by => fun((Node | Zipper) -> boolean()),
+        filtered_from => node | zipper,
```
paulo-ferraz-oliveira, line 33: `filtered_from` says that:

1. the function returns nodes or zippers, and
2. correspondingly the function filters on nodes or zippers

(there was a somewhat confusing mix of nodes or zippers before)

The function also returns {zipper, Zippers} to 1. be sure you understand what you're matching on, 2. trick Dialyzer.

test/style_SUITE.erl
```diff
@@ -1926,7 +1926,7 @@ verify_no_successive_maps(Config) ->
                     elvis_test_utils:elvis_core_apply_rule(
                         Config, elvis_style, no_successive_maps, #{}, Path
                     ),
-                [#{line_num := 10}, #{line_num := 26}, #{line_num := 32}] =
+                [#{line_num := 11}, #{line_num := 27}, #{line_num := 33}] =
```
paulo-ferraz-oliveira, line 1929: We now point to the succession, not the original. As-is it's simpler, but I can always go back.

src/elvis_style.erl
```diff
@@ -3,108 +3,70 @@
...
-    ms_transform_included/3,
-    no_boolean_in_comparison/3,
-    no_operation_on_same_value/3,
-    no_receive_without_timeout/3
+    no_init_lists/1,
+    ms_transform_included/1,
+    no_boolean_in_comparison/1,
+    no_operation_on_same_value/1,
+    no_receive_without_timeout/1,
+    root/1
```
paulo-ferraz-oliveira, line 60: Mind this addition. It's reused in `elvis_text_style` but maybe we can move it to `elvis_core`, too, where the types definitions lie?

src/elvis_style.erl
```diff
@@ -340,267 +305,295 @@ default(RuleWithEmptyDefault) when
...
+    FunctionName = ktn_code:attr(name, FunctionNode),
+    unicode:characters_to_list(atom_to_list(FunctionName)).
 
--spec consistent_variable_casing(
-    elvis_config:config(),
-    elvis_file:file(),
-    consistent_variable_casing_config()
-) ->
-    [elvis_result:item()].
-%% @todo Use maps:groups_from_list/2 when we specify OTP25 as the minimum OTP version
```
paulo-ferraz-oliveira, line 423: Handled it. First time using the function.

src/elvis_code.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_code.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_style.erl
paulo-ferraz-oliveira: Changes pushed.

paulo-ferraz-oliveira:
> This is a library after all… Maybe they're used by rebar3_lint, elvis, or elvis.mk ?

And it makes sense, which is why (macros, I know), a little trick I do these days is

```erlang
-define(API, [fun1/0, fun2/0]).
-export(?API). -ignore_xref(?API).
```

I thus can Xref-ignore the actual API I consider so.

src/elvis_code.erl
```diff
@@ -50,29 +91,27 @@ find(Pred, Root) ->
...
-code_zipper(Root, Mode) ->
-    case Mode of
+-spec zipper(tree_node(), content | all) -> tree_node_zipper().
+zipper(Root, Traverse) ->
+    case Traverse of
         content ->
             content_zipper(Root);
         all ->
             all_zipper(Root)
     end.
```
elbrujohalcon, line 112:
```suggestion
zipper(Root, content) ->
    content_zipper(Root);
zipper(Root, all) ->
    all_zipper(Root).
```

src/elvis_core.erl
```diff
@@ -23,8 +23,9 @@
 -type rule() ::
     {Module :: module(), Function :: atom(), RuleConfig :: rule_config()}
     | {Module :: module(), Function :: atom()}.
+-type rule_cfg_with(Rule) :: {elvis_config:config(), elvis_file:file(), Rule}.
```
elbrujohalcon, line 26: Where is this type used? I only see it in this definition… 🤔

elbrujohalcon: Also… if we're not removing it, maybe consider calling it `rule_context(Rule)` or something like that.

src/elvis_style.erl
elbrujohalcon: I see the function in `elvis_code`… but it was not removed from here 🤔

src/elvis_style.erl
```diff
@@ -2986,3 +2734,21 @@ bin_parts_to_iolist(Src, Parts) when is_binary(Src), is_list(Parts) ->
...
+    re_compile(Regexp, [unicode]).
+
+re_compile(undefined, _Options) ->
+    undefined;
+re_compile(Regexp, Options) ->
+    {ok, MP} = re:compile(Regexp, Options),
+    MP.
+
+root(RuleCfg) ->
+    elvis_code:root(RuleCfg).
```
elbrujohalcon, line 2754: I would just remove this function and call `elvis_code:root/1` everywhere, but even if you keep it… Please don't export it.

elbrujohalcon (changes requested):

src/elvis_style.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_core.erl
paulo-ferraz-oliveira: I was used in `:root/1` at one point (when it had a typespec), so maybe it can be removed.

src/elvis_code.erl
paulo-ferraz-oliveira: Changes pushed. Discussion started.

src/elvis_style.erl
paulo-ferraz-oliveira: Changes pushed.

paulo-ferraz-oliveira:
Brujo, feel free to merge tomorrow, if you're happy. I know I am. If not, we can always do the smaller changes later in a follow-up, if they're details.

elbrujohalcon (approved):

PR #505, opened 2025-06-28
Rename some of the rules

### LINKED ISSUE #496: Rename `macro_names` as `macro_naming_convention`
~**And** adding a `forbidden_regex` to it.~ This is done in #492.

### LINKED ISSUE #504: Rename some of the existing rules
### Discussed in https://github.com/inaka/elvis_core/discussions/497

<div type='discussions-op-text'>

<sup>Originally posted by **paulo-ferraz-oliveira** June 22, 2025</sup>
Now that we have a few years of experience with creating and naming rules, some might be Ok to revisit (I'm not attached to any of them - and I'm pretty sure this'll potentially happen in the future, so we should try to have a simple function that would translate them internally, but still warn on the wrong name):

The `no_` "category":

* `god_modules` > `no_god_modules`
* `nesting_level` > `no_deep_nesting`
* `invalid_dynamic_call` > `no_invalid_dynamic_calls`
* `used_ignored_variable` > `no_used_ignored_variable`

The `consistent_` category: 

* `param_pattern_matching` > `consistent_param_pattern_matching`
* `numeric_format` > `consistent_numeric_format`
* `operator_spaces` > `consistent_operator_spaces`

Those are the ones that pop in my head, atm.</div>

paulo-ferraz-oliveira:
# Description

We rename some modules according to the linked issue (closes #504, closes #496).

## Changes

- [x] rename `god_modules` as `no_god_modules`
- [x] rename `nesting_level` as `no_deep_nesting`
- [x] rename `invalid_dynamic_call` as `no_invalid_dynamic_calls`
- [x] rename `used_ignored_variable` as `no_used_ignored_variables`
- [x] rename `macro_names` as `macro_naming_convention`
- [x] rename `consistent_generic_type` as `generic_type `
- [x] rename `consistent_variable_casing` as `variable_casing`
- [x] deprecate `old_configuration` (and with it ruleset `elvis_config`)
- [x] remove pre-1.0 normalisation of config. (`dirs` used to be `src_dirs`, `[#{}]` used to be `#{}`)
- [x] tweak README on newer output visuals
- [x] move reusable code for `option` into `elvis_ruleset`
- [x] remove dead code (run Xref and take `elvis` and `rebar3_lint` into account)
- [x] export, for tests, stuff that is only for tests

---

Opening as draft for potential feedback as I go.


paulo-ferraz-oliveira:
I'm basically missing "implement `warnings_as_errors`" but I've paved the code-way for it with the changes until now.

paulo-ferraz-oliveira:
I'm putting this to ready for review as I started working on validation configuration and it'll be not only a bit out-of-scope but out-of-review focus, also...

src/elvis_core.erl
```diff
@@ -21,8 +21,8 @@
 -type target() :: source_filename() | module().
 -type rule_config() :: #{atom() => term()}.
 -type rule() ::
-    {Module :: module(), Function :: atom(), RuleConfig :: rule_config()}
-    | {Module :: module(), Function :: atom()}.
+    {Ruleset :: module(), Rule :: atom(), RuleConfig :: rule_config()}
```
paulo-ferraz-oliveira, line 24: `Ruleset` is a `Module`, but semantically I think it makes more sense to rename it.

elbrujohalcon:
We had a _migration/upgrade guide_ somewhere… We need to update it with the changes needed to move from 4.1.0 to 4.2.0. Particularly considering all the renames.

paulo-ferraz-oliveira:
I think ruleset and filegroups are different things, sure, but where do you see those?

src/elvis_core.erl
```diff
@@ -21,8 +21,8 @@
...
+    | {Ruleset :: module(), Rule :: atom()}.
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: What is ODT? Opaque Definition Type?

paulo-ferraz-oliveira:
> We had a migration/upgrade guide somewhere… We need to update it with the changes needed to move from 4.1.0 to 4.2.0. Particularly considering all the renames.

I think if we consider that:

1. 4.2.0 will have a warnings_as_erros
2. the renames will be warnings (not errors) before moving forward

there's actually no need for a migration guide, at this moment. We nudge consumers in the right direction, (the one for which we give support), while adding warnings and only crashing if people want to (thus the option). How does that sound?

paulo-ferraz-oliveira:
We could also call the module ruledefs and the rulesets we keep as is, because they're more part of the interface than the former.

src/elvis_core.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'll check if anything would make sense at this moment... and try to act on it.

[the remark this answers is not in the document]
paulo-ferraz-oliveira: On the other hand I don't want to have to rename all `Rule` as `RuleName`. The rule is, above all else, the name. The rule def. is the name, config., etc.

elbrujohalcon:
> > We had a migration/upgrade guide somewhere… We need to update it with the changes needed to move from 4.1.0 to 4.2.0. Particularly considering all the renames.
> 
> I think if we consider that:
> 
> 1. 4.2.0 will have a warnings_as_erros
> 2. the renames will be warnings (not errors) before moving forward
> 
> there's actually no need for a migration guide, at this moment. We nudge consumers in the right direction, (the one for which we give support), while adding warnings and only crashing if people want to (thus the option). How does that sound?

OK. I see now that MIGRATION.md is for major versions (i.e. 3.x to 4.x, and not 4.1.x to 4.2.x).
👌🏻

src/elvis_core.erl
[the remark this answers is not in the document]
elbrujohalcon: > What is ODT? Opaque Definition Type?

Opaque Data Type. 
[Reference](https://youtu.be/Q7FWm3R3dtk?si=3ehJtP6SWR49gcrc) - Although I learned later that Opaque was a better term for it than Abstract… but the concept is the same.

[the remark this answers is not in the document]
elbrujohalcon: > On the other hand I don't want to have to rename all `Rule` as `RuleName`. The rule is, above all else, the name. The rule def. is the name, config., etc.

Maybe the ODT can be…

```erlang
-module elvis_rule_def.

-type config() :: #{atom() => term()}.
-record(elvis_rule_def, {namespace :: module(), rule :: atom(), config :: config()}).

-opaque t() :: #elvis_rule_def{}.
-export_type [config/0, t/0].

-spec new(module(), atom(), config()) -> t().
new(…) -> #elvis_rule{…}.

-spec config(t()) -> config().
config(#elvis_rule{config = Config}) -> Config.

…etc…
```

src/elvis_code.erl
```diff
@@ -164,13 +164,13 @@ find_with_zipper(Pred, Zipper, Results, Mode) ->
             find_with_zipper(Pred, zipper:next(Zipper), NewResults, Mode)
     end.
 
--spec root({Ruleset, Config, Target, RuleConfig}) -> Res when
-    Ruleset :: module(),
+-spec root({RuleNamespace, Config, Target, RuleConfig}) -> Res when
```
elbrujohalcon, line 167: Not a change request, but… Following """""standard practice""""" from other languages, you might want to consider calling these variables `RuleNS`.

paulo-ferraz-oliveira: Sure. I thought about it, then I "heard" you saying the contrary, that `NS` would not be readable 😄

src/elvis_core.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: The change (85720c5) ended up being a bit bigger than expected but I think it's worth it.

New files: `elvis_rule`. All exposed types opaque, and then we derived all the changes from that, and fixed until all tests passed.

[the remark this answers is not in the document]
paulo-ferraz-oliveira: And we still remove around 300 lines of code, at the moment.

src/elvis_code.erl
```diff
@@ -164,15 +164,13 @@ find_with_zipper(Pred, Zipper, Results, Mode) ->
...
-    Target :: elvis_file:file(),
-    RuleConfig :: (Options :: #{atom() => term()}),
+-spec root(Rule, ElvisConfig) -> Res when
+    Rule :: elvis_rule:t(),
+    ElvisConfig :: elvis_config:config(),
     Res :: ktn_code:tree_node().
-root({_RuleNamespace, Config, Target, RuleConfig}) ->
-    {Root0, File0} = elvis_file:parse_tree(Config, Target, RuleConfig),
-    case maps:get(ruleset, Config, undefined) of
+root(Rule, ElvisConfig) ->
```
paulo-ferraz-oliveira, line 171: This simplification allowed:

1. to move all the options into the new type
2. to reduce the API
3. to make sure we're using `ElvisConfig` basically everywhere

src/elvis_config.erl
```diff
@@ -250,56 +250,52 @@ ignore_to_regexp(A) when is_atom(A) ->
     "/" ++ atom_to_list(A) ++ "\\.erl$".
 
 %% @doc Merge user rules (override) with elvis default rules.
--spec merge_rules(UserRules :: list(), DefaultRules :: list()) -> [elvis_core:rule()].
+-spec merge_rules(UserRules :: list(), DefaultRules :: list()) -> [elvis_rule:t()].
```
paulo-ferraz-oliveira, line 253: Now everything's a rule.

README.md
```diff
@@ -133,10 +127,13 @@ The `dirs` key is a list that tells `elvis_core` where it should look for the fi
 
 If you want to override the [pre-defined rules](#pre-defined-rules), for a given ruleset, you need
 to specify them in a `rules` key which is a list of items with the following structure
-`{Module, Function, RuleConfig}`, or `{Module, Function}` - if the rule takes no configuration
+`{RuleNamespace, Rule, RuleConfig}`, or `{RuleNamespace, Rule}` - if the rule takes no configuration
 values. You can also `disable` certain rules if you want to, by specifying them in the `rules` key
 and passing `disable` as a third argument.
 
+`RuleNamespace` is an Erlang module.
```
elbrujohalcon, line 134:
```suggestion
`RuleNamespace` is an Erlang module that implements the `elvis_rule` behaviour.
```

src/elvis_rule.erl
```diff
@@ -0,0 +1,153 @@
...
+        name = Name,
+        def = Def,
+        ignores = maps:get(ignore, Def, [])
+    }.
+
+-spec from_tuple(NSName | NSNameDef) -> t() when
+    NSName :: {NS :: module(), Name :: atom()},
+    NSNameDef :: {NS :: module(), Name :: atom(), Def :: disable | map()}.
+from_tuple(Rule) when is_record(Rule, rule) ->
+    Rule;
```
elbrujohalcon, line 56: 🤔  This… doesn't match the spec… 🤔

src/elvis_rule.erl
```diff
@@ -0,0 +1,153 @@
...
+defmap(Map) ->
+    Map.
+
+-spec ignorable(module() | {module(), atom()} | {module(), atom(), arity()}) -> ignorable().
+ignorable(Ignorable) ->
+    Ignorable.
+
+-spec same(t(), t()) -> boolean().
+same(RuleL, RuleR) ->
+    % we make a loose comparison for a specific use case
```
elbrujohalcon, line 152:
```suggestion
-spec same(t(), t()) -> boolean().
%% @doc Loose equality comparison for a specific use case
same(RuleL, RuleR) ->
```

src/elvis_rule.erl
paulo-ferraz-oliveira: Yeah, right? Dialyzer ftw.

src/elvis_file.erl
```diff
@@ -40,54 +40,52 @@ src(#{path := Path} = File) ->
...
-parse_tree(_Config, File, _RuleConfig) ->
-    throw({invalid_file, File}).
+    {ktn_code:tree_node(), t()}.
+parse_tree(File, ElvisConfig) when is_map(File) ->
+    Rule = elvis_rule:new(no_namespace, no_rule, elvis_rule:defmap(#{})),
+    parse_tree(elvis_rule:file(Rule, File), ElvisConfig);
+parse_tree(Rule, ElvisConfig) ->
+    File0 = elvis_rule:file(Rule),
+    case File0 of
+        #{parse_tree := ParseTree0} ->
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Changes pushed. The above probably has a `badmatch` error somewhere.

src/elvis_rule.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_rule.erl
paulo-ferraz-oliveira: Changes pushed.

src/elvis_file.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: If you like the result `case` feel free to merge.

PR #522, opened 2025-08-05
New Rules for anonymous function lengths

### LINKED ISSUE #161: Rule idea: allow only short/trivial inline funs
Take [an example from MongooseIM (actually a legacy leftover from ejabberd)](https://github.com/esl/MongooseIM/blob/4211e4ca6e11d60cccfc0da16f08ba0c0d8e4a94/apps/ejabberd/src/mod_muc_room.erl#L2225-L2281) - this _inline fun_ is 55 lines long!
This leads to the following issues:
- the fun code has problems with indentation, since the body of the inline fun should be indented more than the outer code and using `case`s and `if`s increases the indentation level even more,
- the inline fun is not easily traceable, because we don't know its real name (at least without referring to `erlc -S`) - in case of a fun this long, its logic might not be obvious, so tracing and seeing the return value might be really helpful.

An inline fun could be classified as too long/complex by counting the number of lines or indentation levels (maybe the same method as in inaka/elvis#82).

Elvis could even suggest a solution like this (possibly parameterized with the names of functions/arguments in question):

``` erlang
%% Before refactoring
do_something_on_a_list_of_items(ListOfItems) ->
    SomeVar1 = get_some_var1(),
    SomeVar2 = get_some_var2(),
    SomeVar3 = get_some_var3(),
    lists:map(fun (Elem) ->
                      %% very long closure using variables
                      %% SomeVar1, SomeVar2, SomeVar3 closed over
                      %% from the outer environment
              end, ListOfItems).

%% After refactoring
do_something_on_a_list_of_items(ListOfItems) ->
    SomeVar1 = get_some_var1(),
    SomeVar2 = get_some_var2(),
    SomeVar3 = get_some_var3(),
    lists:map(mk_step(SomeVar1, SomeVar2, SomeVar3), ListOfItems).

mk_step(SomeVar1, SomeVar2, SomeVar3) ->
    fun (Elem) ->
            do_something_with_one_item(SomeVar1, SomeVar2, SomeVar3, Elem)
    end.

do_something_with_one_item(SomeVar1, SomeVar2, SomeVar3, Elem) ->
    %% still a long function using variables
    %% SomeVar1, SomeVar2, SomeVar3
    %% passed in as arguments
    ...
```

elbrujohalcon:
# Description

A brief description of your changes.

Closes #161.


bormilan (approved):
I like it, and I didn't find anything to suggest (sorry, I'm so below your level lol), so I wanted to wait for @paulo-ferraz-oliveira's review.
I read it so many times, it feels perfect (as my level of knowledge and exp).

PR #529, opened 2025-09-28
New Rule: Macro Definition Parentheses

### LINKED ISSUE #152: New Rule: Parentheses in Macro Definitions
Ref: WhatsApp/erlfmt#91

* Omit `()` in macro names only if they represent constants
```erlang
%% Bad
-define(NOT_CONSTANT, application:get_env(myapp, key)).
-define(CONSTANT(), 100).
```
```erlang
%% Good
-define(NOT_CONSTANT(), application:get_env(myapp, key)).
-define(CONSTANT, 100).

### PR DESCRIPTION
**Pull request by bormilan**
# Description

Prevent using parentheses in macro definitions that can be misleading:
  - use parentheses if the macro defines a function
  - do not use, if the macro defines a constant value

Closes #152;.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -2552,7 +2552,8 @@ macro_attrs({_, Attrs, _, _}) ->
     Attrs.
 
 is_stringified_function(Tree) ->
-    re:run(Tree, "^[a-zA-Z_][a-zA-Z0-9_]*\\([^)]*\\)$", [{capture, none}]) =:= match.
+    re:run(Tree, "^[a-zA-Z_][a-zA-Z0-9_]* ?:? ?[a-zA-Z0-9_]*\\([^)]*\\)$", [{capture, none}]) =:=
+        match.
 
 get_tree_content({tree, text, _, Content}) ->
     Content.
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -64,13 +64,8 @@
     simplify_anonymous_functions/2,
     prefer_include/2,
     strict_term_equivalence/2,
-    parentheses_in_macro_defs/3
+    macro_definition_parentheses/2
 ]).
-
-% -define(PARENTHESES_IN_MACRO_DEFS,
-%     "Invalid parenthesis at a macro in line ~p"
-%     "Functions should contain parenthesis, constants shouldn't"
-% ).
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %% Default values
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: test/examples/fail_macro_definition_parentheses.erl [hunk 1]
```diff
@@ -0,0 +1,10 @@
+-module(fail_macro_definition_parentheses).
+
+-export([get_attr/1]).
+
+-define(CONSTANT(), 5).
+-define(GET_NAME, get_attr(name)).
+-define(GET_AGE, get_attr(age)).
```
elbrujohalcon, line 7: Another one that should fail, but… I'm not sure if Elvis will be able to catch it…

```erlang
-define(THE_MACRO, io:format("~p\n", [).
```

This ☝🏻 is sadly perfectly valid Erlang code…

```erlang
-module(x).

-export([x/0]).

-define(THE_MACRO, io:format("~p\n", [).

x() ->
    ?THE_MACRO
    bananas]).
```

# 🤷🏻

Personallly, I would try to add some test like that… and then:
* if the rule as-it-is works, amazing 🎆 
* if Elvis crashes when applying the rule, I would fix that
* if the rule doesn't report a warning, but it also doesn't crash… let's keep it as-is.

elbrujohalcon (changes requested):

test/examples/fail_macro_definition_parentheses.erl
bormilan: It doesn't report a warning, but also doesn't crash. I will look into it to see if it's possible to fix or not with a simple solution, but my only idea is to check with a regular expression or something similar, which is not the best approach.

elbrujohalcon (changes requested):

src/elvis_style.erl
```diff
@@ -2500,6 +2501,63 @@ is_public_data_type_in(TypesToCheck, TypeAttrNode) ->
...
+    }),
+
+    lists:map(
+        fun(#{attrs := #{value := [Elem1, _Elem2]}}) ->
+            {Line, Column} = proplists:get_value(location, macro_attrs(Elem1)),
+            elvis_result:new_item(
+                "Invalid parenthesis at a macro definition in line ~p"
+                "Functions should contain parenthesis, constants should not",
+                [],
+                #{line => Line, column => Column}
```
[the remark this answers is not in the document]
bormilan: You're right!

src/elvis_style.erl
```diff
@@ -2500,6 +2501,53 @@ is_public_data_type_in(TypesToCheck, TypeAttrNode) ->
...
+    {nodes, InvalidMacroNodes} = elvis_code:find(#{
+        of_types => [define],
+        inside => elvis_code:root(Rule, ElvisConfig),
+        filtered_by => TypeMismatchingDefine
+    }),
+
+    lists:map(
+        fun(Node) ->
+            elvis_result:new_item(
+                "Invalid parenthesis at a macro definition in line ~p"
```
elbrujohalcon, line 2530: I guess you still need to remove the `~p` from here and, in fact, the reference to a line.

PR #534, opened 2025-10-07
Fix #533: New rule: max_record_fields

### LINKED ISSUE #533: New `erl_files_strict` (+hrl +beam) rule: `max_record_fields`
### Name

`max_record_fields`

### Brief Description

Do not declare records with more than a fixed number of keys, the default being `25`.

### Problematic code

```erlang
-record t, {
  k1 :: pos_integer(),
  k2 :: pos_integer(),
  …
  k25 :: pos_integer()
}.
```

### Correct code

```erlang
-record t, {
  k1 = pos_integer(),
  k2 = pos_integer(),
  ...
  k20 = pos_integer(),
  a_substructure :: #sub_t{}
}.
-record sub_t, {
  k21 :: pos_integer(),
  ...
  k25 :: pos_integer()
}.
```

### Rationale

It is considered an anti-pattern to have a record definition that declares more than 25 fields. It's very likely that many of them can be grouped in substructures that would provide better semantics to the whole.

elbrujohalcon:
# Description

Add a new rule: `max_record_fields`.

Closes #533.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1415,7 +1415,7 @@ max_record_fields(Rule, ElvisConfig) ->
     [
         elvis_result:new_item(
             "record ~p has ~p fields, which is higher than the configured limit",
-            [ktn_code:attr(name, LargeRecordNode)],
+            [ktn_code:attr(name, LargeRecordNode), length(ktn_code:content(LargeRecordNode))],
             #{node => LargeRecordNode, limit => MaxFields}
         )
      || LargeRecordNode <- LargeRecordNodes


bormilan (approved):

doc_rules/elvis_style/max_record_fields.md
```diff
@@ -0,0 +1,23 @@
+# Max Record Fields
```
paulo-ferraz-oliveira: This has no `since | x.y.z` or `BEAM | yes` (if it works for beam). Should be taken into account by the issue I created recently.

PR #536, opened 2025-10-08
Revisit strict from an options' perspective

### LINKED ISSUE #495: Revisit `strict` from an options' perspective
Some options, like, `private_data_types` might benefit from strict including all `apply_to` types. Check what rules this makes sense to change for and allow for a generic configuration mechanism.

bormilan:
# Description

I added all types to the default of `private_data_types `.
Revisited all the rules, but I didn't find any others like this.

Closes #495.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -79,6 +79,7 @@ Most, if not all, of the rules will present (opinionated) documentation sections
 - [Operator Spaces](doc_rules/elvis_style/operator_spaces.md)
 - [Param Pattern-Matching](doc_rules/elvis_style/param_pattern_matching.md)
 - [Prefer Include](doc_rules/elvis_style/prefer_include.md)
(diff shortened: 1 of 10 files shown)


elbrujohalcon:
Actually… it's an opaqueness that I missed in bormilan/elvis_core#1 … If you can't fix it on your own, let me know and I can help :)

bormilan:
Thank you; I will try it with the help of your description. At least I can test if I understand it clearly.🤓

bormilan:
I think this might be a good solution, as it helps to visualize that the type `t()` belongs to `elvis_file`, so it feels good to separate this operation into `elvis_file` even if it's just a simple `maps:get`. What do you think @elbrujohalcon ?

elbrujohalcon:
Yeah… it's perfect, @bormilan. Nice work!

PR #541, opened 2025-10-24
Strict module layout

### LINKED ISSUE #288: Strict Module Layout
### Brief Description

Provide module parts in the required order.

```erlang
-module(…).

-include[_lib](…).

-dialyzer(…).
-elvis(…).
-format #{…}.
-hank(…).
-mixin(…).
…

-type(…).
-type(…).
…

-export_type(…).

-export(…).

exported_function_1(…) ->
…

exported_function_2(…) ->
…

non_exported_function_1(…) –>
…
```

### Should be on by default?

**YES**

### Options

-  `order :: [Part] when Part :: module | includes | types | … | everything_else` (default: ??)

We need to come up with a list of all possible parts and a reasonable default for them. We can inspire ourselves in Credo (link below) but Erlang has some extra attribute types and we need to be able to correctly deal with `-if`, `-ifdef`, and all its friends.

### Reasoning

> Like all `Readability` issues, this one is not a technical concern. But you can improve the odds of others reading and liking your code by making it easier to follow.

### Origin (#281)

Inspired by the `Consistency.Readability.StrictModuleLayout` rule from [Credo](https://hexdocs.pm/credo/Credo.Check.Readability.StrictModuleLayout.html)

bormilan:
# Description

A new rule for making a strict layout order for Erlang modules.
Now, the user can define an order by creating a list of ot atoms like this: `[module, export, function]`

This rule will examine the modules and verify if the actual order is correct or not.

Three open questions I have:
  - Is the default list good enough? We could define a simpler one. (that contains just a basic 3-4 types)
  - Do we want to make result error messages at every wrong node (now we have), or do we want just one error message pre module that describes the module is incorrect in its order of parts?
  - How to handle `-ifdef`? I have two possible solutions, one of which is working with this code, and the other would require some additional logic. I describe these two below with examples.

Closes #288;.

### DIFF
--- a/doc_rules/elvis_style/strict_module_layout.md
+++ b/doc_rules/elvis_style/strict_module_layout.md
@@ -1,4 +1,4 @@
-# Strict Module Layout ![](https://img.shields.io/badge/BEAM-yes-orange)
+# Strict Module Layout [![](https://img.shields.io/badge/since-4.2.0-blue)](https://github.com/inaka/elvis_core/releases/tag/4.2.0) ![](https://img.shields.io/badge/BEAM-yes-orange)
 
 Prefer sticking an exact order of the different parts in a module.
 
--- a/doc_rules/elvis_style/strict_module_layout.md
+++ b/doc_rules/elvis_style/strict_module_layout.md
@@ -11,6 +11,7 @@ Prefer sticking an exact order of the different parts in a module.
 
 -include[_lib](…).
 ```
+
 > with a configuration that describes an order: [module, include, dialyzer]
 
 ## Prefer
@@ -22,11 +23,13 @@ Prefer sticking an exact order of the different parts in a module.
 
 -dialyzer(…).
 ```
+
 > with a configuration that describes an order: [module, include, dialyzer]
 
 ## Rationale
 
-This is a readability issue. It can improve the odds of others reading and liking your code by making it easier to follow.
+This is a readability issue. It can improve the odds of others reading and
+liking your code by making it easier to follow.
 
 ## Options
 
@@ -39,5 +42,6 @@ This is a readability issue. It can improve the odds of others reading and likin
 ## Example configuration
 
 ```erlang
-{elvis_style, strict_module_layout, #{order => [module, inclue, dialyzer, type, export_type, export, function_call]}}
+{elvis_style, strict_module_layout,
+  #{order => [module, inclue, dialyzer, type, export_type, export, function_call]}}
 ```
--- a/doc_rules/elvis_style/strict_module_layout.md
+++ b/doc_rules/elvis_style/strict_module_layout.md
@@ -1,0 +1,43 @@
+# Strict Module Layout ![](https://img.shields.io/badge/BEAM-yes-orange)
+
+Prefer sticking an exact order of the different parts in a module.
+
+## Avoid
+
+```erlang
+-module(…).
+
+-dialyzer(…).
+
+-include[_lib](…).
+```
+> with a configuration that describes an order: [module, include, dialyzer]
+
+## Prefer
+
+```erlang
+-module(…).
+
+-include[_lib](…).
+
+-dialyzer(…).
+```
+> with a configuration that describes an order: [module, include, dialyzer]
+
+## Rationale
+
+This is a readability issue. It can improve the odds of others reading and liking your code by making it easier to follow.
+
+## Options
+
+- order :: [atom]
+  - default: [module, include, dialyzer, elvis, mixin, type_attr, export_type, export, function]
+
+- ignore_mod_parts :: [atom]
+  - default :: []
+
+## Example configuration
+
+```erlang
+{elvis_style, strict_module_layout, #{order => [module, inclue, dialyzer, type, export_type, export, function_call]}}
+```
--- a/doc_rules/elvis_style/strict_module_layout.md
+++ b/doc_rules/elvis_style/strict_module_layout.md
@@ -43,5 +43,5 @@ liking your code by making it easier to follow.
 
 ```erlang
 {elvis_style, strict_module_layout,
-  #{order => [module, inclue, dialyzer, type, export_type, export, function_call]}}
+  #{order => [module, include, dialyzer, type, export_type, export, function_call]}}
 ```
--- a/src/elvis_ruleset.erl
+++ b/src/elvis_ruleset.erl
@@ -140,7 +140,6 @@ elvis_style_rules() ->
         elvis_rule:new(elvis_style, variable_naming_convention),
         elvis_rule:new(elvis_style, guard_operators),
         elvis_rule:new(elvis_style, simplify_anonymous_functions)
-        % elvis_rule:new(elvis_style, strict_module_layout)
     ].
 
 erl_files_test_rules() ->
@@ -183,7 +182,8 @@ elvis_style_stricter_rules() ->
         elvis_rule:new(elvis_style, state_record_and_type),
         elvis_rule:new(elvis_style, prefer_include),
         elvis_rule:new(elvis_style, strict_term_equivalence),
-        elvis_rule:new(elvis_style, macro_definition_parentheses)
+        elvis_rule:new(elvis_style, macro_definition_parentheses),
+        elvis_rule:new(elvis_style, strict_module_layout)
     ].
 
 elvis_text_style_stricter_rules() ->
--- a/src/elvis_ruleset.erl
+++ b/src/elvis_ruleset.erl
@@ -139,8 +139,8 @@ elvis_style_rules() ->
         elvis_rule:new(elvis_style, no_used_ignored_variables),
         elvis_rule:new(elvis_style, variable_naming_convention),
         elvis_rule:new(elvis_style, guard_operators),
-        elvis_rule:new(elvis_style, simplify_anonymous_functions),
-        elvis_rule:new(elvis_style, strict_module_layout)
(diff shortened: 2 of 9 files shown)


test/examples/pass_strict_module_layout.erl
```diff
@@ -0,0 +1,38 @@
...
+-ifdef(TEST).
+-export_type([name/0]).
+-endif.
+
+-export([some_fun/0, some_fun_2/0, some_fun_3/0, some_fun_4/0]).
+
+% -ifdef(TEST).
+% -type name() :: definition.
+% -export_type([name/0]).
+% -endif.
```
bormilan: The second solution comes into play if we want to define a layout where all the -ifdef things are in one place, before the functions. If we want this, the rule would ignore anything inside the -ifdef, and refer to this section with all content as it's just one -ifdef part.

bormilan:
I didn't forget to add a description of the possible attributes that can be added to this rule's config, working on it.

doc_rules/elvis_style/strict_module_layout.md
```diff
@@ -0,0 +1,47 @@
+# Strict Module Layout ![](https://img.shields.io/badge/BEAM-yes-orange)
```
elbrujohalcon: You're missing the "since" badge.

elbrujohalcon (changes requested):
There are a bunch of things missing in the tests, like…
* `-ignore_xref`
* `-hank`
* `-import`
* `-compile`
* `-doc`
* `-define`
* other preprocessor macros, like. `-else.`
* `-opaque` and `-nominal`
* maybe others I had not thought about yet…

elbrujohalcon:
I would like to hear the opinion of @pablocostass in this topic, too.

elbrujohalcon:
> `[module, includes, custom_attributes, defines, records, types, export_types, exports, body]`,

I wonder if we could come up with a comprehensive list of _first level things in a module_. I know that _technically_ the list is `[function, attribute]` but that doesn't say much. And the trick here is to be able to split the list of possible attributes in a way that…
1. makes senes; but also…
2. includes every possible attribute.

src/elvis_ruleset.erl
```diff
@@ -140,6 +140,7 @@ elvis_style_rules() ->
         elvis_rule:new(elvis_style, variable_naming_convention),
         elvis_rule:new(elvis_style, guard_operators),
         elvis_rule:new(elvis_style, simplify_anonymous_functions)
+        % elvis_rule:new(elvis_style, strict_module_layout)
```
elbrujohalcon: We should add this to the strict ruleset, I think. Not to the default one.

src/elvis_style.erl
```diff
@@ -2759,6 +2766,77 @@ macro_definition_parentheses(Rule, ElvisConfig) ->
...
+        end,
+
+    {_, ResultNodes} = lists:foldl(NotInPlace, {nil, []}, Nodes),
+
+    lists:map(
+        fun({CorrectPrev, ActualPrev, NodeTypeCategory, Node}) ->
+            elvis_result:new_item(
+                "A ~p type element was found after a ~p type element,"
+                "but it should follow a type ~p",
+                [NodeTypeCategory, ActualPrev, CorrectPrev],
```
elbrujohalcon:
```suggestion
                "A ~p type element was found after a ~p type element."
                " That doesn't respect the expected module layout: ~p",
                [NodeTypeCategory, ActualPrev, Order],
```

src/elvis_style.erl
```diff
@@ -2759,6 +2766,77 @@ macro_definition_parentheses(Rule, ElvisConfig) ->
...
+        CorrectPrev =:= nil orelse
+            CorrectPrev =:= Prev orelse
+            Category =:= Prev
+    of
+        true -> {Category, Result};
+        false -> {Category, [{CorrectPrev, Prev, Category, Node} | Result]}
+    end.
+
+ignore(Type, Category, Order) ->
+    lists:member(Type, [comment, ifdef, ifndef, 'else', 'if', elif, endif]) orelse
```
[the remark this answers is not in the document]
elbrujohalcon: And… in that case… Does `no_if_expression` emit a warning for `-if`s, too? That looks like a bug to me.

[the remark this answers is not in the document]
elbrujohalcon: I just cheked and, indeed `-if` has type `'if'`, which is concerning for `no_if_expression` (I'll verify that rule and write a bug for this if it emits warnings), but `else` is the one for `maybe`. `-else` has type `else_attr`.

test/style_SUITE.erl
```diff
@@ -3210,6 +3211,40 @@ verify_macro_definition_parentheses(Config) ->
...
+        ),
+
+    [] =
+        elvis_test_utils:elvis_core_apply_rule(
+            Config,
+            elvis_style,
+            strict_module_layout,
+            #{order => [module, exports, body]},
+            FailPath
+        ).
```
elbrujohalcon: I think you should add examples of failures with different layouts. Say, using `[type, export_type]` (which is the inverse order as the fail module).

You should also somehow test that this _always_ fails:

```erlang
…
-type bananas() :: bananas.
-export_type([bananas/0]).
-type tomatoes() :: tomatoes.
–
```

Well… It always fails unless `types` or `exports` are not part of the `order` parameter.

We should also define and test what should happen if `#{order => []}`.
What would that mean? Would it be the same as `disable` (i.e., any order is fine… never emit a warning)? Or something else?

Also… what should we do if somebody specifies `#{order => [types, exports, types, body]}` (which has duplicates) or `#{order => [body, module]}` (which is incompatible with the Erlang compiler)?

src/elvis_style.erl
```diff
@@ -2759,6 +2766,77 @@ macro_definition_parentheses(Rule, ElvisConfig) ->
...
+        {function, body},
+        {include, includes},
+        {include_lib, includes},
+        {type_attr, types},
+        {opaque, types},
+        {nominal, types},
+        {function, body},
+        {spec, body},
+        {export, exports},
+        {export_type, exports},
```
elbrujohalcon:
```suggestion
        {export_type, export_types},
```

I'm pretty sure that there will be people who prefer putting `-export_type([…]).` away from `-export([…]).`.

elbrujohalcon (changes requested):
A lot of good progress, but we still need to improve a few things around this rule.

src/elvis_style.erl
bormilan: Yes you right

test/style_SUITE.erl
bormilan: Interesting ideas; I will write a lot of tests, as you suggested in the first section.
Yes, this regex-like definition could be good, but I agree that it might be too complex. If someone argues about the current version, they can simply disable it. But I think this is flexible enough to satisfy most people.

doc_rules/elvis_style/strict_module_layout.md
```diff
@@ -0,0 +1,47 @@
+# Strict Module Layout [![](https://img.shields.io/badge/since-4.2.0-blue)](https://github.com/inaka/elvis_core/releases/tag/4.2.0) ![](https://img.shields.io/badge/BEAM-yes-orange)
```
paulo-ferraz-oliveira: The `since` will probably have be to revisited once this is ready to be tagged.

doc_rules/elvis_style/strict_module_layout.md
```diff
@@ -0,0 +1,47 @@
...
+
+Prefer sticking an exact order of the different parts in a module.
```
paulo-ferraz-oliveira:
```suggestion
Prefer a consistent and well-defined order of module "sections".
```

doc_rules/elvis_style/strict_module_layout.md
```diff
@@ -0,0 +1,47 @@
...
+
+-dialyzer(…).
+```
+
+> with a configuration that describes an order: [module, include, dialyzer]
+
+## Rationale
+
+This is a readability issue. It can improve the odds of others reading and
+liking your code by making it easier to follow.
```
paulo-ferraz-oliveira:
```suggestion
A standardized layout allows developers to quickly locate declarations such as exports, records, macros, and function definitions. It also simplifies code reviews and reduces unnecessary diffs caused by reordering unrelated sections.
```

doc_rules/elvis_style/strict_module_layout.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Options
+
+- order :: [atom]
+  - default: [module, include, dialyzer, elvis, mixin, type_attr, export_type, export, function]
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'd go with:
`-module`
-behaviour declarations (if any)
-export and -export_type attributes
-compile and other module attributes
-include and -include_lib directives
Macro definitions (-define)
Record definitions (-record)
Type definitions (-type, -opaque)
Public function definitions
Internal (non-exported) function definitions

paulo-ferraz-oliveira:
Good stuff here. I won't look at all the implementation details at this moment, but gave my 5c in the documentation with some guidance on potential choices.

elbrujohalcon:
> @elbrujohalcon, once we agree on the grouping/default order of the elements, maybe that could be mentioned in the issue this PR derived from, to improve guidance and later documentation for release?

Yeah… More importantly, it has to super-clear in the rule docs, but adding it to the original issue is a good idea, too.

PR #548, opened 2026-02-13
Optimise rules with heavy computation

NelsonVides:
# Description

After the upgrade from 4.1 to 4.2, linting https://github.com/dnsimple/erldns with absolutely all rules enabled went from <1s to 15s with parallelism of 1. This takes it back to 2s. If using parallelism from #547 then it takes it back to <1s again.


NelsonVides:
Elvis is failing with a line I did not change myself so I'm confused 🥲

elbrujohalcon (approved):

PR #515, opened 2025-07-15
Deeper configuration parsing and validation

paulo-ferraz-oliveira:
# Description

This pull request is all about:

1. configuration validation (check https://github.com/inaka/elvis_core/discussions/508)
2. preparation for some of the upcoming ergonomics/UX changes we have identified in Issues

- [x] I have read and understood the [contributing guidelines](/inaka/elvis_core/blob/main/CONTRIBUTING.md)

First of all, I want to acknowledge this is a big change and some (much?) stuff can change before it's considered for merging.

Two things stand out, to me:

1. I rebased on the latest 6 commits to the main branch and had 0 conflicts, which shows it's somewhat self-contained (even if BIG)
2. I want to update `rebar3_lint` and `elvis` in tandem with this, before it's considered for merging, because this a. touches on the API the aforementioned use, b. hopefully simplifies and gives utilities for a more unified experience in the Elvis ecosystem.

## Required changes

- [ ] I have updated `rebar3_lint` here (TBD: link will come when I do it)
- [ ] I have updated `elvis` here (TBD: link will come when I do it)
- [ ] I have written tests for this
- [ ] I have written tests that touch on structure and type and as many exceptions as I can remember
- [ ] I have tweaked `rebar3_lint`'s code to benefit from 1. the default config. (mention in README), 2. the `throw` elements
- [ ] I have tweaked `elvis`'s code to benefit from 1. the default config. (mention in README), 2. the `throw` elements
- [ ] I have decided on showing results without `throw`ing (kind of the current behaviour which prints sometimes and throws sometimes), and adapted tests to that

Closes #438.
Closes #439.

---

## Further considerations

As always:

1. I'm opening this as draft but Ok to get comments on it
2. I haven't written tests yet, but have imagined how I'll do it (I basically adapted, up to now, the code for the tests to pass, trying to change them slightly, but not fundamentally)
3. I'll self-review my choices to allow for further discussion points, and these might also lead to further changes I'll eventually push

PR #551, opened 2026-02-14
Fix: Allow glob dir patterns that match nothing

### LINKED ISSUE #543: Default elvis config produces an error if the project is not an umbrellla one
## Bug Description

Using the config from `elvis_config:default()` on a minimal rebar3 project fails because there is no `apps/**/src` directory.

## To Reproduce

Run `elvis_core:rock(elvis_config:default()).` on a simple Erlang project, with folders such as `src`, `include`, etc… without an `apps` folder.

This error is reported:

```erlang
in 'elvis.config', at list position number 1, in 'dirs', the following elements are not (or don't contain) directories: ["apps/**/src"].
```

## Expected Behavior

Elvis runs with its default configuration.

## Additional Context

- OS: MacOS
- Erlang OTP28
- rebar3 26.0

NelsonVides:
Fixes #543

The dirs validation introduced in #515 rejected glob patterns like
`"apps/**/src"` when they didn't match any existing directories,
breaking non-umbrella projects using the default config.

Remove the per-element directory existence check entirely, and relax the
dir+filter combo validation to require that at least one combination
yields files, rather than all of them.

### DIFF
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -626,10 +626,9 @@ config_is_valid(CustomRulesetNames, Config) ->
         % and the fact it knows about elvis_core's internals, but we
         % should shortly revisit this
         ok ?= map_keys_are_in(Config, [dirs, filter, ignore, ruleset, rules, files]),
-        Dirs = get_config_opt(dirs, Config),
-        ok ?= is_nonempty_list_of_dirs(dirs, Dirs),
         Filter = get_config_opt(filter, Config),
         ok ?= is_nonempty_string(filter, Filter),
+        Dirs = get_config_opt(dirs, Config),
         ok ?= all_dirs_filter_combos_are_valid(Dirs, Filter),
         Ignore = get_config_opt(ignore, Config),
         ok ?= is_list_of_ignorables(ignore, Ignore),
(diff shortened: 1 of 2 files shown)


src/elvis_config.erl
```diff
@@ -657,29 +656,6 @@ map_keys_are_in(Map, Keys) ->
                 ])}
     end.
 
-is_nonempty_list_of_dirs(What, List) when not is_list(List) orelse List =:= [] ->
-    {error, io_lib:format("'~s' is expected to be a non-empty list.", [What])};
```
kivra-pauoli: Does this mean you no longer check for empty lists? Or where's the other (newer) check?

src/elvis_config.erl
```diff
@@ -689,30 +665,15 @@ is_nonempty_string(What, String) ->
     end.
 
 all_dirs_filter_combos_are_valid(Dirs, Filter) ->
```
paulo-ferraz-oliveira: I think this function is better named `at_least_one_dirs_filter_combo_is_valid`. I'll propose it.

src/elvis_config.erl
NelsonVides: The problem is that the default config contained both `apps/**/...` and `src/**` and chances are that almost for sure one of them is empty and then the default config would almost surely be invalid, which is a really bad practice. My change here was to ensure that at least one file can be found, so `dirs => []` would fail.

In any case, all of this is irrelevant now because we just reworked everything to stop using `dirs` so whatever, let's have this discussion in the current state :D

Issue #267, opened 2023-02-27
Consistent Variable Naming


elbrujohalcon:
### Brief Description

This is a more sophisticated version of #266, but it's also more _heuristic_ (thus the possibility of emitting false positives).
The idea is for Elvis to use some _sound-alike_ algorithm to find variables with _similar_ names, like `MyVar` and `My_Var`, or `MyVar` and `MyVariable`, or `MyID` and `My_id`, etc…

### Should be on by default?

**NO**, unless we find a suitable default with a very low false positive rate.

### Reasoning

The reasoning here is, again, similar to #266:
> If two variables have the same name, it's very likely that they refer to the same concept and, therefore, it would make the lives of code readers easier if they share the exact same name, including casing and all.

### Refactoring Proposal

Rename one or more of the variables.

### Configuration

The rule must have some `soundness_level` parameter to adjust how much _alike_ two variables need to be for them to be considered _the same_.

### Examples

```erlang
-module bad.

-export [my_fun/1, my_fun/2].

my_fun(MyVar) -> my_fun(MyVar, too).

my_fun(My_Var, too) -> {My_Var, too}.
```

```erlang
-module good.

-export [my_fun/1, my_fun/2].

my_fun(MyVar) -> my_fun(MyVar, too).

my_fun(MyVar, too) -> {MyVar, too}.
```

```erlang
-module also_good.

-export [my_fun/1, my_fun/2].

my_fun(MyVar) -> my_fun(MyVar, too).

my_fun(MyVar, MyOtherVar) -> {MyVar, MyOtherVar}.
```

paulo-ferraz-oliveira:
Were you thinking about something like `string:jaro_similarity` with a given interval, for this?

PR #554, opened 2026-02-16
New `erl_files` rule: `consistent_variable_naming`

NelsonVides:
Detect variables with the same logical name written in different styles within a module (e.g. TypeVar vs Type_var, MyVar vs My_Var). Variables are grouped by canonical token form and each group with inconsistent surface names is reported once.

Unlike variable_casing (which groups by uppercased name), this rule tokenises names by splitting on casing boundaries and underscores, catching differences across naming conventions (CamelCase vs snake_case).

Included in the strict ruleset, not enabled by default.

Closes inaka/elvis_core#267


elbrujohalcon (approved):
@NelsonVides maybe a hot take, but… If this rule supersedes `consistent_variable_casing`, can we…
* Remove/deprecate `consistent_variable_casing` indicating users that they should use this one instead?
* Make existing `consistent_variable_casing` tests use this new rule and produce the same (or more) warnings? (I mean… instead of removing them)

NelsonVides:
> We're not going to do this for `4.3.0`, I added #556 for doing that in `5.0.0`. Nevertheless, we can add this one to the default ruleset and remove `consistent_variable_casing` from it, can we?

Well that's also a breaking change 😅

Somebody (me for example xD) with some config with that rule, running in CI with a dependency declared as `{rebar3_lint, "~>  4.0"}` would magically get 4.2 installed and get CI failing because the config is invalid, even if the PR introduced nothing that would break regarding that rule. That's the contract with the semantic versioning, I can use `"~> 4.0"` and I'd automatically get any bugfixes and new features without manual upgrades, but no breaking changes.

NelsonVides:
> > But if it's a renaming, won't that show up as a warning, and actually not break anything?
> 
> That was our usual assumption before @NelsonVides joined the team, yes. But he has a point, I think.

I originally joined the team to fix all the breaking changes you introduced into my CI because I had `~> 4.0` in my projects, and here we are now 😜

PR #566, opened 2026-02-27
Replace options' pair `dirs`/`filter` with single option `files` (globs)

### LINKED ISSUE #516: Move from `dirs`+`filter` to `files`
At the moment, we have `dirs`, `files` (not exposed), and `filter` (that are "concatenated" with `dirs`) to identify files.

Feature request: get rid of all but `files` (expose it), since we can express everything with that alone.

Also, it's common practice in other libraries so the concept shouldn't be foreign to users and we can always implement it (initially) in a backward compatible way.

NelsonVides:
# Description

Config now uses a single `files` key: a list of glob patterns (e.g. ["src/*.erl", "apps/**/src/*.erl"]) instead of `dirs` and `filter`. This is a breaking change for 5.0.0; `dirs` and `filter` are no longer supported.

- elvis_file: new find_files(Globs) and filter_files(Files, Globs, Ignore); from_path/1 for building opaque t(); old find_files/2 and filter_files/4 removed.
- elvis_config: resolved file list is stored under `resolved_files`; user-facing `files` holds the glob list. Getters are file_globs/1 and files/1 (from resolved_files). resolve_files/1 resolves via globs; resolve_files/2 filters a given file list by globs. Validation allows only [files, ignore, ruleset, rules] and requires at least one glob to match. default/0 and defaults use the new files format.

Closes #516.


paulo-ferraz-oliveira:
Maybe worth taking into consideration: either here or the next pull request (we can create an issue). If `elvis_core` decides to use the default config, that should potentially output an info. At the moment I struggle to understand why a specific `elvis.config` is not used as a config. since it's:

1. present
2. seemingly valid

so the warning for 

```
===> Invalid config: in 'elvis.config', at list position number 2, no '<dir>' + '<filter>' combo in ["include", "apps/**/include", "src", "apps/**/src"] + '*.hrl' yielded any files to analyse.
```

is cryptic because I can't find that in my `elvis.config`.

paulo-ferraz-oliveira:
I also think a nice way to move this forward would be to indicate deprecation of the previous options, while allowing for them, and then in a future version break it, but if you feel it's not worth it we can move as-is.

README.md
```diff
@@ -129,19 +126,15 @@ To look at what is considered the "current default" configuration, do:
...
+        "src/*.hrl","include/*.hrl"],
    ruleset => hrl_files},
- #{filter => "rebar.config",
-   dirs => ["."],
+ #{files => ["rebar.config"],
    ruleset => rebar_config},
- #{filter => ".gitignore",
-   dirs => ["."],
+ #{files => [".gitignore"],
    ruleset => gitignore}]
```
paulo-ferraz-oliveira: If you decide to adapt the above you could then revisit if it makes a diff. here (but I think the CLI results care not about how the code is written).

README.md
```diff
@@ -151,10 +144,10 @@ documentation was updated.
 
 ### Files, rules and rulesets
 
-The `dirs` key is a list that tells `elvis_core` where it should look for the files that match
-`filter`, which will be run through each of the pre-defined rules in the specified `ruleset`.
-`filter` can contain `**` for further matching (it uses
-[`filelib:wildcard/1`](https://erlang.org/doc/man/filelib.html#wildcard-1) under the hood).
+The `files` key is a list of glob patterns that tell `elvis_core` which files to analyse. Each
+pattern uses [`filelib:wildcard/1`](https://erlang.org/doc/man/filelib.html#wildcard-1), so you can
```
paulo-ferraz-oliveira: I think (don't know by heart) there's a simpler way to write this and ExDoc will translate it to a link to the doc.s

README.md
NelsonVides: I'm not sure to understand what did you mean here 🤔

README.md
NelsonVides: https://github.com/inaka/elvis_core/issues/573

paulo-ferraz-oliveira:
> [#566 (comment)](https://github.com/inaka/elvis_core/pull/566#issuecomment-3978836544) and [#566 (comment)](https://github.com/inaka/elvis_core/pull/566#issuecomment-3978841563), are they related to the changes in this PR? Perhaps work on those points deserve its own follow-up issue? 🤔

We can always revisit later. If you think this implementation is not piling on that issue, fine; otherwise, I was mentioning it in case you'd seen it and wanted to take care of it. But less is more, so I'd postpone to a newer PR; I don't think we're in any hurry to tag a new release.

src/elvis_config.erl
```diff
@@ -688,24 +665,22 @@ map_keys_are_in(Map, Keys) ->
                 ])}
     end.
 
-is_nonempty_string(What, String) ->
-    case io_lib:char_list(String) andalso length(String) > 0 of
+all_files_globs_are_valid(FileGlobs) when not is_list(FileGlobs) orelse FileGlobs =:= [] ->
+    {error, "'files' is expected to be a non-empty list."};
```
paulo-ferraz-oliveira: Basically, what I mean is I don't see this covered, but maybe I missed it 😄 

Conditions:

1. is a list
2. is a non-empty list

The "being compulsory" would be tested by some other element.

elbrujohalcon:
> I also think a nice way to move this forward would be to indicate deprecation of the previous options, while allowing for them, and then in a future version break it, but if you feel it's not worth it we can move as-is.

We decided to create release `5.0.0` specifically to avoid stuff like that. Personally, I prefer breaking changes and good docs in `MIGRATION.md` over support for old+new ways of configuring stuff.

test/elvis_config_SUITE.erl
```diff
@@ -156,42 +155,6 @@ validate(_Config) ->
                     "in 'elvis.config', at list position number 1, the following keys are unknown: [<<\"key\">>]."
                 )},
 
-            % invalid elvis.config > dirs (does not exist)
```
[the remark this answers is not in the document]
NelsonVides: Fair enough, easy to re-add 💪🏽

elbrujohalcon:
> maybe all PRs that break stuff should just end up changing `MIGRATION.md` and that's it

I second that motion!

PR #575, opened 2026-03-01
Honor `.gitignore`'s patterns to skip unnecessary paths

### LINKED ISSUE #165: Support honoring the entries in a .gitignore file
It'd be great if Elvis could be configured to ignore files matching entries in a .gitignore file.  This would be useful to ignore auto-generated files without having to duplicate the content of .gitignore in elvis.config.

Tentative configuration example:

``` erlang
[{elvis,
  [{config,
    [#{dirs => ["src", "test"],
       filter => "*.erl",
       honor_gitignore => true,
       ruleset => erl_files
      }
    ]
   }]
 }] 
```

paulo-ferraz-oliveira:
# Description

In addition to custom ignore rules, honor `.gitignore` during execution to skip unnecessary paths.

Closes #165.

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -20,7 +20,7 @@
                                 {elvis_core, main, 1}
                             ]
                         }},
-                        {elvis_style, no_god_modules, #{ignore => [elvis_style]}},
+                        {elvis_style, no_god_modules, #{ignore => [elvis_style, elvis_config]}},
                         {elvis_style, prefer_strict_generators, disable},
                         {elvis_style, no_throw, disable},
                         {elvis_style, max_function_length, disable},


src/elvis_config.erl
```diff
@@ -259,9 +263,10 @@ dirs(#{}) ->
 ignore(Config) when is_list(Config) ->
     lists:flatmap(fun ignore/1, Config);
 ignore(#{ignore := Ignore}) ->
-    lists:map(fun ignore_to_regexp/1, Ignore);
+    lists:map(fun ignore_to_regexp/1, Ignore) ++ ignore(#{});
 ignore(#{}) ->
-    [].
+    % Base state.
+    gitignored().
```
paulo-ferraz-oliveira: Fairly straightforward, we populate `ignore` with elements from `.gitignore`, by default.

src/elvis_config.erl
```diff
@@ -796,8 +801,30 @@ flag({_Option, _What} = Obj) ->
...
+gitignored() ->
+    _ = case check_flag(gitignore_checked()) of
+        true ->
+            % Already "in the ETS"
+            ok;
+        false ->
+            % Flag it as "in the ETS"
+            flag(gitignore_checked()),
+            % Populate it
+            GitIgnored = string:lexemes(os:cmd("git check-ignore *"), "\n\r"),
```
paulo-ferraz-oliveira: `git check-ignore *` comes from the issue description; then we split it for newline terminations.

src/elvis_config.erl
```diff
@@ -796,8 +801,40 @@ flag({_Option, _What} = Obj) ->
...
+                flag(gitignore_checked()),
+                % Populate it
+                flag({gitignored, git_check_ignore()})
+        end,
+    case ets_lookup(gitignored) of
+        [] ->
+            [];
+        [{gitignored, GitIgnored}] ->
+            GitIgnored
+    end.
```
NelsonVides:
```suggestion
    case ets:lookup_element(?TABLE, gitignored, 2, undefined) of
        undefined ->
            [];
        GitIgnored ->
            GitIgnored
    end.
```
And slightly faster, as it skips creating a list and a tuple in the process heap and also skips comparing the atom of the first element in the tuple, which by the lookup we already know it is what it is with certainty.

src/elvis_config.erl
NelsonVides: Why do we need this ensure_ets_table to begin with? Why isn't created just once by the first process? 🤔

src/elvis_config.erl
```diff
@@ -796,8 +801,40 @@ flag({_Option, _What} = Obj) ->
...
+        case check_flag(gitignore_checked()) of
+            true ->
+                % Already "in the ETS"
+                ok;
+            false ->
+                % Flag it as "in the ETS"
+                flag(gitignore_checked()),
+                % Populate it
+                flag({gitignored, git_check_ignore()})
+        end,
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Yeah, naming is misleading but I didn't want to change the logic that much. We flag that we've started (`gitignore_checked/0`), and if so don't populate the table again. Then we add the actual content of `.gitignore` (`{gitignored, git_check_ignore()}`). If we don't flag in e.g. tests, we keep calling the function to populate. Also, given how `elvis` and e.g. `rebar3_lint` might call `elvis_core` (on top of which I try to make little assumptions) this is a minor protection mechanism against race conditions.

812 is the semaphore for 806. 814 is a value (previously called `flag` in name, already)

paulo-ferraz-oliveira:
> Wouldn't it be faster and easier to use `persistent_term` instead of an ets table for this?

I didn't make assumptions on performance; I already had the ETS at hand, so went with it. I'd start with a process then figure out that another process is actually holding the state, e.g. during tests, and... I didn't wanna go that route. Went with what was easier for me... Using the process dictionary is more obscure, to me, than an ETS table.

Is the "request for changes" to actually block the merge because it uses ETS? Or a generic comment?

src/elvis_config.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I took the time to create a new function, `ets_insert`, that does the insurance and then inserts. I've used it from `flag`, and directly, to give more semantic meaning to the implementation.

src/elvis_config.erl
paulo-ferraz-oliveira: Something similar is now in place. Thanks for the suggestion.

src/elvis_config.erl
```diff
@@ -796,8 +801,40 @@ flag({_Option, _What} = Obj) ->
...
+                os:cmd("git check-ignore *")
+        end,
+    string:lexemes(GitCheckIgnore, "\r\n").
+
+gitignore_checked() ->
+    {gitignore, checked}.
+
 check_flag({Option, _What} = Obj) ->
-    try ets:lookup(elvis_config, Option) of
+    try ets_lookup(Option) of
```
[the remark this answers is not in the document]
NelsonVides: I find this much more verbose than reusable, it's wrapping a one-liner not doing complex pattern-matching or anything else, which points to the most standard library of them all (`ets`), I'd just repeat all `ets:lookup_element/4` where is needed any number of times, much easier to grep later. Also the name is now even more misleading, because `ets_lookup` is not calling `ets:lookup`, but `ets:lookup_element`.

src/elvis_config.erl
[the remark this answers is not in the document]
elbrujohalcon: I believe that some, if not all, of this will be far simpler using `persistent_term` instead of an `ets` table.

paulo-ferraz-oliveira:
So:

1. https://github.com/inaka/elvis_core/pull/575#discussion_r2881104219 - I'll drop the function and copy the code everywhere it's used; it's not my preference, but a. I don't own `elvis_core`, b. I don't care enough to keep going back and forth on details of this nature, especially since my time is limited

2. https://github.com/inaka/elvis_core/pull/575#discussion_r2876616382 - I'll remove implementations of ets, in a different PR, then rebase this one (I hate to mix implementation details otherwise); also mind you, `ets` was introduced in https://github.com/inaka/elvis_core/pull/453, also by me (not that I didn't know `persistent_term` didn't exist, but it wasn't commented on) so I kept consistent throughout

paulo-ferraz-oliveira:
Mind you I changed the base for this.

src/elvis_config.erl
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Resolve'ing since not relevant for current implementation.

NelsonVides:
My main issue was around how ets tables were used. That a persistent term would be a much better solution (which I totally think so but didn't discover that until this PR) could be moved to a different PR, even to one coming after this one and not before. The issue was that not only `ets_lookup` is a bad name to wrap `ets:lookup_element/4` but that I'd never wrap that to begin with. If in the helper module we were doing complex pattern-match on the result and then running some transformations and other computation, but, literally just `ets_lookup(Val) -> ets:lookup(table, Val)` and nothing else, I'd seriously never let that pass 🥲

CI decided to complain, I have some time now to debug and fix 👌🏽

src/elvis_config.erl
elbrujohalcon: It seems to me that `gitignored/0` returns `ok`, but `ignore/1` returns a list. How does this work?

PR #580, opened 2026-03-03
Add migration notes, rule badges, and ruleset update

### LINKED ISSUE #567: Add documentation-level info labels
For rules added after 4.1.1, add information like

```
BEAM | yes
since | x.y.z
HRL-only | yes
until | x.y.z
```

where appropriate, at the top next to the rule name.

Most rules have it, the ones that don't were released before the first public version.

NelsonVides:
# Description

After comments at https://github.com/inaka/elvis_core/pull/564#issuecomment-3988146550, https://github.com/inaka/elvis_core/pull/555#discussion_r2864645369, and other conversations. Adding badges to rules, updating migrations, and recovering a new rule into the default ruleset.

Closes #567.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -787,9 +787,10 @@ is_operator_node(Node) ->
         map_field_assoc,
         map_field_exact
     ],
-    NodeOperators = ktn_code:content(Node),
-    IsNotSingleNodeOperator = length(NodeOperators) > 1 andalso lists:member(NodeType, OpOrMatch),
-    IsNotSingleNodeOperator orelse lists:member(NodeType, ExtraOpsTypes).
+    case ktn_code:content(Node) of
+        [_, _ | _] -> lists:member(NodeType, OpOrMatch);
+        _ -> lists:member(NodeType, ExtraOpsTypes)
+    end.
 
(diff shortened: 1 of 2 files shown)


src/elvis_style.erl
```diff
@@ -2482,7 +2483,10 @@ no_single_match_maybe(Rule, ElvisConfig) ->
     ].
 
 is_single_match_maybe(MaybeNode) ->
-    length(ktn_code:content(MaybeNode)) =:= 1.
+    length_equal(1, ktn_code:content(MaybeNode)).
+
+length_equal(1, [_]) -> true;
```
elbrujohalcon: I wrote a longer comment somewhere, but it's lost now…
Basically… can we not _slightly bend_ our own rules like this? (¬_¬)
Maybe do…

```erlang
has_a_single_element([_]) -> true;
has_a_single_element(_) -> false
```

…if you really think this extra function is needed. Or… just write case statements in the two places were it's used :)

NelsonVides: Totally, I'm fine with full case statements, pushed a patch for that :D

elbrujohalcon (approved):

PR #583, opened 2026-03-03
Extend `ignore` documentation around using (string) filenames

### LINKED ISSUE #574: Improve documentation around "### The `ignore` option"
## Bug Description

"### The `ignore` option" (`RULES.md`) does not mentioned strings as possible input.

## To Reproduce

N/A

## Expected Behavior

The documentation should be updated, as simply as possible, to reflect currently implemented behavior.

elbrujohalcon:
# Description

Extend docs for ignoring files by path.

Closes #574.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -190,6 +190,7 @@ It suffices to add the `ignore` list to your rules, as per the example below.
 
 You can add the exceptions using the following syntax:
 
+- file name: `ignore => ["path/to/header.hrl"]`
 - whole module: `ignore => [mod]`
 - functions (by name only): `ignore => [{mod, fun}]` (available for **`elvis_style`-based rules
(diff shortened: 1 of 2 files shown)


RULES.md
```diff
@@ -190,6 +190,7 @@ It suffices to add the `ignore` list to your rules, as per the example below.
 
 You can add the exceptions using the following syntax:
 
+- file name: `ignore => ["path/to/header.hrl"]`
```
paulo-ferraz-oliveira: If I'm not mistaken this can even be more generic, since we're checking regexes, right? Or do we wanna be relaxed in the implementation and more strict here?

paulo-ferraz-oliveira (approved):

PR #584, opened 2026-03-04
Remove `ets` (replace with `persistent_term`) where appropriate

paulo-ferraz-oliveira:
# Description

We simplify it, according to discussions in https://github.com/inaka/elvis_core/pull/575.

### DIFF
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -39,6 +39,10 @@
 -ignore_xref([from_rebar/1, from_file/1, default/0, resolve_files/2, apply_to_files/2]).
 -ignore_xref([set_output_format/1, set_verbose/1, set_no_output/1, set_parallel/1]).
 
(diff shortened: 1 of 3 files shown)


src/elvis_config.erl
```diff
@@ -411,45 +415,41 @@ elvis_control_opts() ->
     [output_format, verbose, no_output, parallel].
 
 do_validate({elvis = Option, Elvis}) ->
-    case check_flag(validation_started(Option)) of
-        true ->
+    maybe
+        {v, false} ?= {v, has_validation_started_for(Option)},
```
paulo-ferraz-oliveira: Why use a prior case if we can rely on `maybe`, right?

I tag this value to safely match it in `else`. (if I were to just match it to `false`, for example, any function below that'd also match would produce a similar result (return this function scope with `ok`) and that's undesirable.

PR #591, opened 2026-03-07
Add considerations on versioning to README

### LINKED ISSUE #590: README should state SemVer as our way to version
SemVer is a standard, but not the only one. If it's made explicit, consumers can have expectations on breakage between versions. Lets add a note to the README, along with how we propose migrations between majors.

paulo-ferraz-oliveira:
# Description

Closes #590.


NelsonVides (approved):
💪🏽

PR #603, opened 2026-03-10
Add expression_can_be_simplified rule (issue #577)

### LINKED ISSUE #577: New `erl_files` rule: `expression_can_be_simplified`
### Name

`expression_can_be_simplified`

### Brief Description

This rule identifies expressions that are mathematically or logically redundant. While the Erlang compiler often optimizes these away at bytecode level, keeping them in the source code increases cognitive load and can signal "dead" logic or incomplete refactoring.

### Problematic code

```erlang
-module(main).
foo(X) ->
    Total = X + 0,
    Items = [] ++ [1, 2, 3],
    is_valid(Total) andalso true.
```

### Correct code

```erlang
-module(main).
foo(X) ->
    Total = X,
    Items = [1, 2, 3],
    is_valid(Total).
```

### Rationale

This check because it focuses on readability, enhancing code clarity and removing redundant operations that often linger after refactoring. While the compiler handles the performance side, `X ++ []` suggests the developer might have intended to append a variable that was later hardcoded to `[]`, leaving behind "code noise."

### Operations on which simplification is possible

* `[] ++ X` | `X ++ []`
* `[] -- X`
* `X -- []`
* `X + 0` | `X - 0`
* `0 - X`
* `X * 1` | `1 * X`
* `X div 1`
* `X rem 1`
* `true andalso X`
* `false orelse X`
* `not true`
* `X band -1`
* `X bor 0`
* `X bxor 0`
* `not false`

Inspiration: https://whatsapp.github.io/erlang-language-platform/docs/erlang-error-index/w/W0019/

NelsonVides:
# Description

New elvis_style rule that reports expressions that can be simplified using identity/absorbing constants (e.g. X + 0, [] ++ X, true andalso X). Configurable via simplifications list; each pattern has a label so users can enable or disable specific simplifications.

Closes #577 

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -4,7 +4,6 @@
     {elvis_style, abc_size, #{ignore => [{elvis_style, default, 1}]}},
     {elvis_style, code_complexity, #{ignore => [{elvis_style, default, 1}]}}
 ]).
--hank([{unnecessary_function_arguments, [{true, 1}]}]).
 
 -behaviour(elvis_rule).
 -export([default/1]).
@@ -2088,13 +2087,15 @@ expression_can_be_simplified(Rule, ElvisConfig) ->
     {nodes, OpNodes} = elvis_code:find(#{
         of_types => [op],
         inside => elvis_code:root(Rule, ElvisConfig),
-        filtered_by => fun(OpNode) -> is_tuple(can_be_simplified(OpNode, Simplifications)) end,
+        filtered_by => fun(OpNode) ->
+            is_tuple(recommended_simplification(OpNode, Simplifications))
+        end,
         traverse => all
     }),
     UniqOpNodes = lists:uniq(OpNodes),
     lists:map(
         fun(OpNode) ->
-            {Label, Msg} = can_be_simplified(OpNode, Simplifications),
+            {Label, Msg} = recommended_simplification(OpNode, Simplifications),
             elvis_result:new_item(
                 "expression can be simplified (~s): ~s",
                 [Label, Msg],
@@ -2104,8 +2105,8 @@ expression_can_be_simplified(Rule, ElvisConfig) ->
         UniqOpNodes
     ).
 
--spec can_be_simplified(ktn_code:tree_node(), [atom()]) -> false | {atom(), string()}.
-can_be_simplified(OpNode, Simplifications) ->
+-spec recommended_simplification(ktn_code:tree_node(), [atom()]) -> false | {atom(), string()}.
+recommended_simplification(OpNode, Simplifications) ->
     case ktn_code:content(OpNode) of
         [Operand] ->
             Op = ktn_code:attr(operation, OpNode),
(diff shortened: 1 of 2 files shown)


doc_rules/elvis_style/expression_can_be_simplified.md
```diff
@@ -0,0 +1,92 @@
...
+| `1 * X`             | `X`      | `multiply_by_one_left`        |
+| `X div 1`           | `X`      | `div_by_one`                  |
+| `X rem 1`           | `0`      | `rem_by_one`                  |
+| `true andalso X`    | `X`      | `andalso_true`                |
+| `false orelse X`    | `X`      | `orelse_false`                |
+| `not true`          | `false`  | `not_true`                    |
+| `not false`         | `true`   | `not_false`                   |
+| `X band -1`         | `X`      | `band_neg_one`                |
+| `X bor 0`           | `X`      | `bor_zero`                    |
+| `X bxor 0`          | `X`      | `bxor_zero`                   |
```
elbrujohalcon: It never ceases to amaze me how most (if not all and even other similar ones) of these expressions are 100% not equal to the `Prefer` solution in Javascript.
How can that people write code, I wonder… 🤦🏻

src/elvis_style.erl
```diff
@@ -4,6 +4,7 @@
     {elvis_style, abc_size, #{ignore => [{elvis_style, default, 1}]}},
     {elvis_style, code_complexity, #{ignore => [{elvis_style, default, 1}]}}
 ]).
+-hank([{unnecessary_function_arguments, [{true, 1}]}]).
```
elbrujohalcon: Please write a comment above this line explaining why it's necessary.

src/elvis_style.erl
```diff
@@ -2057,6 +2082,126 @@ no_operation_on_same_value(Rule, ElvisConfig) ->
...
+    try_binary_patterns_loop(Pats, Op, Left, Right, Simplifications).
+
+try_binary_patterns_loop(
+    [{OpPat, LeftOk, RightOk, Label, Msg} | Rest], Op, Left, Right, Simplifications
+) ->
+    case
+        Op =:= OpPat andalso
+            lists:member(Label, Simplifications) andalso
+            LeftOk(Left) andalso
+            RightOk(Right)
```
elbrujohalcon: To avoid the need for a function like `true/1`, you may just allow these to be at atom like `no_check` or `all` and do…
```suggestion
            (LeftOk =:= no_check orlese LeftOk(Left)) andalso
            (RightOk =:= no_check orelse RightOk(Right))
```
I'm not sure which one is better… I don't _hate_ the `true/1` function enough to request this change (the change request is due to the lack in comment above if you decide to keep the function), but I would've written this code as I suggest here.

elbrujohalcon (changes requested):
Excellent code, as usual… just one or two nit-pickings from me.

test/examples/fail_expression_can_be_simplified.erl
```diff
@@ -0,0 +1,22 @@
...
+    _ = 1 * X,
+    _ = X div 1,
+    _ = X rem 1,
+    _ = true andalso X,
+    _ = false orelse X,
+    _ = not true,
+    _ = not false,
+    _ = X band -1,
+    _ = X bor 0,
+    _ = X bxor 0,
```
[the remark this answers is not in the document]
NelsonVides: All applied :D

PR #617, opened 2026-03-25
Replace zipper traversal with direct recursion

williamthome:
# Description

`elvis_code:find/1` uses the [zipper library](https://github.com/inaka/zipper) to traverse ASTs. Every node visit allocated a map-based breadcrumb trail (`lefts`, `rights`, `parent_node`, `parent_info`). With 35+ rules per file, each doing a full tree walk, this dominated lint time.

This PR replaces the zipper with direct recursion for all rule traversals. On a 51-file project, total lint time dropped from **72s to 0.83s** (87x speedup).

## Benchmark (51 Erlang files)

| Rule | Before | After |
|---|---|---|
| `variable_naming_convention` | 35s | 0.24s |
| `consistent_variable_naming` | 32s | 0.22s |
| `no_used_ignored_variables` | 24s | < 0.1s |
| `atom_naming_convention` | 7.8s | < 0.1s |
| **Total** | **72s** | **0.83s** |

## What changed

**Core: `elvis_code.erl`**
- `find/1` with `filtered_from => node` now uses body-recursive `find_nodes` - no zipper, no dedup map, no allocation per node
- New `filtered_from => node_and_ancestors` mode passes `{Node, [Parent, Grandparent, ...]}` to predicates, replacing `zipper:up/1` usage
- Zipper path kept for backward compatibility; all internal callers now use `node` or `node_and_ancestors` mode

**Rules: `elvis_style.erl`**
- All 9 zipper-mode rules converted to `node` or `node_and_ancestors`
- `is_var` token lookup: linear scan (242M calls to `is_at_location`) replaced with a pre-built `#{Location => Token}` index
- `exported_functions`/`exported_types` membership: `lists:member` on lists replaced with `sets:is_element`
- `export_used_types` and `state_record_and_type`: multiple AST traversals merged into single pass
- `dont_repeat_yourself`: all zipper usage replaced with direct recursion (`node_size`, `fold_content`, `remove_attrs`, `is_children`)
- `node_line_limits`: `zipper:map` replaced with direct `max_line` recursion
- Redundant `length/1` calls eliminated across several rules

**Infra**
- `elvis_task`: worker tracking changed from `[{Pid, MRef}]` list to `#{Pid => MRef}` map
- Added `bench.escript` for profiling (per-rule timing + eprof)

---


src/elvis_code.erl
```diff
@@ -14,6 +14,8 @@
 
 % These are local debug functions.
 -ignore_xref([print_node/1, print_node/2]).
+% Kept for backward compatibility; no longer used internally.
+-ignore_xref([zipper/1]).
```
elbrujohalcon: I don't think we need to provide this type of backwards compatibility. If it's not used in this project, just remove it altogether, please.

bench.escript
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env escript
```
elbrujohalcon: I'm not 100% sure, but I believe we might benefit from this file being in the `priv` folder. 🤔 
I just find it odd that it's in the root folder of the project… Maybe adding a `scripts` folder, if `priv` is not the place for it?

elbrujohalcon (changes requested):
I like this! I'm not sure in which release/milestone to include it, tho.
I'll leave that decision to @paulo-ferraz-oliveira and @NelsonVides.

elbrujohalcon (approved):

paulo-ferraz-oliveira (approved):
❤️ muito obrigado, cara!

PR #626, opened 2026-05-07
fix consistent_ok_error_spec to ignore behaviours

### LINKED ISSUE #623: `consistent_ok_error_spec` should ignore behaviour callbacks
## Bug Description

If a module implements a behaviour that requires functions to return `{ok, …}` among other options, and the module only ever returns `{ok, ...}`, adding a proper (concise) spec to those functions will raise a warning from `consistent_ok_error_spec`.

## To Reproduce

```erlang
-module my_module.
-behaviour gen_server.
...
-spec init(noargs) -> {ok, state()}.
init(noargs) ->
  {ok, #{some => state}}.
```

## Expected Behavior

No warnings.

## Additional Context

- OS: N/A
- Erlang version: OTP28
- rebar3 version: 27.0

bormilan:
# Description

I added a filter to `elvis_code:find` for the behaviour callbacks.

Closes #623;.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1194,7 +1194,7 @@ known_behaviours() ->
     ].
 
 callbacks_of_behaviour(Behaviour) ->
-    {module, Behaviour} ?= code:ensure_loaded(Behaviour),
+    {module, Behaviour} = code:ensure_loaded(Behaviour),
     erlang:apply(Behaviour, behaviour_info, [callbacks]).
 
 is_behaviour_callback_spec(SpecNode, BehaviourCallbacks) ->
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1170,7 +1170,7 @@ behaviour_callbacks_from_node(BehaviourNode) ->
     maybe
         {module, Behaviour} ?= code:ensure_loaded(Behaviour),
         true ?= erlang:function_exported(Behaviour, behaviour_info, 1),
-        Callbacks = catch Behaviour:behaviour_info(callbacks),
+        Callbacks = catch apply(Behaviour, behaviour_info, [callbacks]),
         true ?= is_list(Callbacks),
         Callbacks
     else
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1130,10 +1130,17 @@ is_opaque_state(TypeAttrOrOpaqueNode) ->
 
 -spec consistent_ok_error_spec(elvis_rule:t(), elvis_config:t()) -> [elvis_result:item()].
 consistent_ok_error_spec(Rule, ElvisConfig) ->
+    Root = elvis_code:root(Rule, ElvisConfig),
+    BehaviourCallbacks = behaviour_callbacks(Root),
+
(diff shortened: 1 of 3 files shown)


elbrujohalcon (changes requested):
I see what you did here, @bormilan ... and, intuitively, it should work, but...
As we saw in https://github.com/erlang/otp/issues/4847 this approach (i.e., checking the _known_ callbacks of a behaviour) won't work in all cases.
I think, at least for now, it's better to do what we did in Hank with `unnecessary_function_arguments`: https://hexdocs.pm/rebar3_hank/unnecessary_function_arguments.html. That is... If the module implements one of the _known_ OTP behaviours that (crucially) **do not have dynamic callbacks** (i.e., `gen_server`, `gen_event`, `supervisor`, and `application`), then it does kind of what you did... otherwise... as soon as the module is detected to implement a different behaviour (Even if it _also_ implements one of the known ones)... the rule is entirely disabled.

bormilan:
You are right, sorry for my superficial work. I used a macro to define these known behaviours. I don't know whether it should be a default parameter. It sounds odd because we don't want parameters for this rule, for sure, but in my head it feels a little bit right, bc that way we can easily understand that this list belongs to this rule, now it's just hanging there at the top of the module.

bormilan:
oh Im acually a dummy. We have a rule called 'no macros'.🫠

src/elvis_style.erl
```diff
@@ -1130,22 +1132,84 @@ is_opaque_state(TypeAttrOrOpaqueNode) ->
...
+callbacks_of_behaviour(Behaviour) ->
+    maybe
+        {module, Behaviour} ?= code:ensure_loaded(Behaviour),
+        true ?= erlang:function_exported(Behaviour, behaviour_info, 1),
+        Callbacks = apply(Behaviour, behaviour_info, [callbacks]),
+        true ?= is_list(Callbacks),
+        Callbacks
+    else
+        _ -> []
+    end.
```
elbrujohalcon: Since we're only executing this for known OTP behaviours, this function can simply be....

```suggestion
callbacks_of_behaviour(Behaviour) ->
    {module, Behaviour} ?= code:ensure_loaded(Behaviour),
    erlang:apply(Behaviour, behaviour_info, [callbacks]).
```

...or...

```suggestion
callbacks_of_behaviour(Behaviour) ->
    Behaviour:behaviour_info(callbacks).
```

...and then you add this on top of the module:

```erlang
-elvis([{elvis_style, invalid_dynamic_call, #{ignore => [{elvis_style, callbacks_of_behaviour, 1}]).
````

Whatever you like the most.

Comment (elbrujohalcon) [reply src/elvis_style.erl]: Oh, and... with that... you can remove the feature toggle for `maybe`,

**Review by elbrujohalcon — CHANGES_REQUESTED** [review]

File: src/elvis_style.erl [hunk 2]
```diff
@@ -1,5 +1,7 @@
 -module(elvis_style).
 
+-feature(maybe_expr, enable).
```
elbrujohalcon: This is no longer needed, isn't it?

elbrujohalcon (changes requested):

elbrujohalcon (approved):

PR #630, opened 2026-05-15
Fix #627: Add SAFE to CI

### LINKED ISSUE #627: Add SAFE to CI
Coming from https://github.com/inaka/elvis_core/discussions/521

elbrujohalcon:
# Description

Add [SAFE](https://safe-docs.erlang-solutions.com) to our CI pipelines.

Closes #627.


elbrujohalcon:
We're _safe_ now, @paulo-ferraz-oliveira @bormilan @NelsonVides !!

bormilan (approved):
Nice! I think we should make an issue to fix the problematic lines that you ignored now.

elbrujohalcon:
> Nice! I think we should make an issue to fix the problematic lines that you ignored now.

I don't think so. I reviewed each one of them and they are, in fact, safe... Even when using functions that _can_ be abused, we know that they are not.

The calls to `list_to_atom` come from parsing Erlang modules that are not compiled. So, those atoms 100% not existing, but we do want to create them for our analysis.
The calls to `file:consult/1` are needed in order to analyse config files and such... We could, in theory, check the size or whatever attribute from those files before running `file:consult/1` on them... but... we'd end up calling `file:consult/1`, anyway... and I'm pretty sure that SAFE doesn't know if we _checked the files before calling the function_.

So... unavoidable ignores, all of them.

PR #633, opened 2026-05-24
Speed up space-check rules with O(1) line/character lookups

williamthome:
# Description

`operator_spaces`, `no_space`, and `no_space_after_pound` fetched the source line and character with `lists:nth/2` (O(n)) and re-decoded the line on every token, which is O(tokens × lines) per file. `lines_in/1` now builds a per-line index once (`{line, chars-tuple, length}`), and `character_at_location` uses `element/2`. No behaviour change; `lists:nth` leaves the profile's hot path.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -769,13 +769,22 @@ no_space_after_pound(Rule, ElvisConfig) ->
         filtered_by => fun is_text_node/1
     }),
 
-    {Lines, Encoding} = lines_in(elvis_rule:file(Rule)),
-    generate_space_check_results({Lines, Encoding}, TextNodes, {right, "#"}, {should_not_have, []}).
+    Lines = lines_in(elvis_rule:file(Rule)),
+    generate_space_check_results(Lines, TextNodes, {right, "#"}, {should_not_have, []}).
 
 lines_in(File) ->
     {Src, File1} = elvis_file:src(File),
     Encoding = elvis_file:encoding(File1),
-    {elvis_utils:split_all_lines(Src), Encoding}.
+    Lines = elvis_utils:split_all_lines(Src),
+    %% Index lines (and their characters) by 1-based line number so the space
+    %% checks can use O(1) element/2 lookups instead of lists:nth/2 over the line
+    %% list and the character list once per token.
+    list_to_tuple([index_line(Line, Encoding) || Line <- Lines]).
+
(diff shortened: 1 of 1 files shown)


elbrujohalcon (commented):
Certainly not my favorite type of optimization, since I think that _semantically_, it makes way more sense for `Lines` to be a list. But I'd defer the review to @paulo-ferraz-oliveira @NelsonVides and @bormilan .

elbrujohalcon:
...and we also need to fix the SAFE issue from #634 and #635.

bormilan:
I agree with @elbrujohalcon . I liked it for the first check, but after considering the change from a list to a tuple, I'm not really a fan. I haven't done any calculations or benchmarks, but does this optimization really provide a significant improvement?

williamthome:
Hey! Thanks for the review and for asking whether the speedup is real. Fair point, I should have brought numbers from the start, so here they are. I ran the repo's own `scripts/bench.escript` on `main` and on this branch (only `elvis_style.erl` changed, OTP 28), taking the fastest run each time.

**On elvis_core itself** (all rules, 36 files, including the big 4251-line `elvis_style.erl`):

| | main | this PR |
|---|---:|---:|
| Total lint | 1229 ms | 1030 ms (16% faster) |
| the 3 space rules together | 264 ms | 67 ms (~4x faster) |

That whole 200 ms gain comes from these 3 rules. `eprof` also shows `lists:nth/2` going from 94,504 calls down to 0, and each line is now decoded once per line instead of once per token.

The gap grows on bigger files, because the old code got slower fast as files grew while the new one grows in a straight line. Pointing the same bench at a single large module (only the 3 rules):

| Lines | main | this PR | faster by |
|---:|---:|---:|---:|
| 4,003 | 287 ms | 81 ms | 3.5x |
| 16,003 | 5,182 ms | 464 ms | 11x |
| 28,003 | 16,205 ms | 1,069 ms | 15x |

The results are exactly the same on both branches (checked on a file full of spacing mistakes), so behavior doesn't change.

About `Lines` becoming a tuple: honestly I'm not a fan of `list_to_tuple` either, but in this particular case I think it earns its place. The slow part that grows with file size is `lists:nth(LineNo, Lines)` looking up the line, and turning the line list into a tuple is what makes that fast. Happy to try another shape if the tuple still bothers you, but we need indexed access somewhere to keep the speedup.

elbrujohalcon:
Very thorough, @williamthome . You convinced me.

PR #634, opened 2026-05-24
Dedup `find_nodes_all` results with `lists:uniq` instead of a per-node `Seen` map

williamthome:
# Description

`elvis_code:find/1` with `traverse => all` tracked visited nodes in a `Seen` map keyed by the *whole node map*, which contains its subtree, so every node got deep-hashed (O(n²)-ish). Drop the map and dedup the (small) result list with `lists:uniq/1` instead. The AST is acyclic, so the walk still terminates; results are unchanged. `find_nodes_all` drops from ~20% of a large lint to ~2%.

### DIFF
--- a/src/elvis_code.erl
+++ b/src/elvis_code.erl
@@ -115,8 +115,11 @@ find_nodes_all_ancestors(_Pred, _Node, _Ancestors, Seen) ->
 find_nodes(Pred, Node, content) ->
     find_nodes_content(Pred, Node);
 find_nodes(Pred, Node, all) ->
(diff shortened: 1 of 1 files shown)


elbrujohalcon (approved):

elbrujohalcon:
This PR presents the same SAFE issue as #635. We need to find a way to fix that and/or remove the SAFE check from CI. Let's see if @robertfiko, @paulo-ferraz-oliveira , or @NelsonVides can figure this out, @williamthome .

PR #635, opened 2026-05-24
Count filtered lines directly instead of building a list to `length` it

williamthome:
# Description

The `max_*_length` rules built a filtered list of a node's lines only to call `length/1` on it. Count with a fold (`count_matching/2`) instead, with no intermediate list. Pure refactor: counts and reported results are unchanged.

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1286,18 +1286,18 @@ max_module_length(Rule, _ElvisConfig) ->
 
     LineIsCommentRegex = line_is_comment_regex(),
     LineIsWhitespaceRegex = line_is_whitespace_regex(),
-    Lines = lists:filter(
+    LineCount = count_matching(
+        Lines0,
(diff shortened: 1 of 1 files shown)


elbrujohalcon:
@NelsonVides / @paulo-ferraz-oliveira: This is failing the new SAFE check. Do you think that's happening because we only have the `SAFE_LICENSE` secret in our fork of this repo and this thing is trying to use @williamthome's fork for running the action?

elbrujohalcon:
@robertfiko what do you think? Maybe you can help us here.

elbrujohalcon (approved):

robertfiko:
@elbrujohalcon Yeah, I see the issue, and I also saw #636 & #637.
I am sad to see that you had to remove SAFE because of this, but it is understandable, the UX is not great with forks...
I will look into what can be done with this, I will let you know when I have something

PR #636, opened 2026-05-26
Only run SAFE on local repository

elbrujohalcon:
# Description

As seen in #633, #634, and #635... we can't run SAFE for contributions coming from external repositories. They lack the secrets that this repo has.

Closes #&lt;issue&gt;.


elbrujohalcon:
Welll… This didn't make the trick 😢

Issue #460, opened 2025-06-10
Should the `erl_files_strict` complain that you're using `type` instead of `nominal`?


paulo-ferraz-oliveira:
We'll wait on gathering community feedback and our own use of the newer types before we move forward on rules around them.

### Discussed in https://github.com/inaka/elvis_core/discussions/428

<div type='discussions-op-text'>

<sup>Originally posted by **paulo-ferraz-oliveira** May 27, 2025</sup>
Or is that too strict, even for Elvis?</div>

PR #638, opened 2026-05-28
Support OTP 28 -nominal types in style rules

williamthome:
# Description

`-nominal` ([EEP-69](https://www.erlang.org/eeps/eep-0069)) is OTP 28's third type-defining attribute alongside `-type` and `-opaque`. Two style rules special-cased `-opaque` but not `-nominal`; this teaches both to treat them alike.

- **`state_record_and_type`**: now recognizes a `state()` type declared with `-nominal`, so it no longer false-flags valid OTP 28 modules as "missing a `state()` type"
- **`max_map_type_keys`**: now also checks `-nominal` map types, mirroring its existing `-opaque` handling

All OTP-28-only syntax in the test examples and suite is gated behind `-if(?OTP_RELEASE >= 28)`, so the build/tests stay green on OTP 26 and 27.

Relates to #460. This only makes existing rules nominal-aware (parity with `-type`/`-opaque`).

### DIFF
--- a/src/elvis_style.erl
+++ b/src/elvis_style.erl
@@ -1746,7 +1746,8 @@ max_record_fields(Rule, ElvisConfig) ->
 max_map_type_keys(Rule, ElvisConfig) ->
     Root = elvis_code:root(Rule, ElvisConfig),
     MaxFields = elvis_rule:option(max_keys, Rule),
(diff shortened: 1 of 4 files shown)


elbrujohalcon (approved):
I worried at first if we were going to start reporting new warnings were there were none before, but after some careful review I think we're good. ✅

PR #639, opened 2026-05-28
Support OTP 29 in CI

williamthome:
# Description

Adds OTP 29 to the CI matrix.

OTP 29 adds new compiler warnings that some `test/examples` fixtures trigger on purpose, so they're suppressed in the test profile (`nowarn_deprecated_catch`, `nowarn_export_vars`, `nowarn_match_alias_pats`); `src` is unaffected.

Follow-up: maybe open a tracking issue per warning to fix the fixtures rather than suppress indefinitely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
-        otp_vsn: ['26', '27', '28']
+        otp_vsn: ['26', '27', '28', '29']
         rebar3_vsn: ['3.25']
         os: [ubuntu-24.04, windows-2022]
     steps:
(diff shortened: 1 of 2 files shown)


elbrujohalcon:
I think I would rather add full support for OTP29 all-at-once (i.e., removing rules that are now checked by the compiler, handling new warnings, etc.) with a single ticket and within `5.1.0` or `6.0.0`, not `5.0.4`.
But I would defer the decision to @bormilan, @NelsonVides, and @paulo-ferraz-oliveira since they are way better at SemVer-ing things than me.

elbrujohalcon:
Besides… the _tradition_ here is to remove a version (in this case OTP26, including all the `ifdef(?OTP_RELEASE >= 27)`s and such) if we add another one (OTP29).

PR #640, opened 2026-05-28
fix cutom rulesets

### LINKED ISSUE #632: Custom rulesets not working via rebar?
## Bug Description

I tried to define and use a custom ruleset in Elvis, but it's not being recognized, at least when coming via the rebar3_lint plugin. It seems that the config validation is done differently for some reason (going into `do_validate({config ...` instead of `do_validate({app ...`) and the custom rules are not found in the persistent term ... or something. It's quite hard to follow what the config code is doing. Not sure if it's a problem in elvis core or if the rebar plugin is calling it wrong.

## Reference

Reported by @richcarl at https://github.com/inaka/elvis/issues/608

bormilan:
# Description

Previously, custom rulesets worked when loading a config explicitly with `elvis_config:from_file/1`, but not when running through `elvis_core:rock({config_file, default})`. In that path, Elvis read the config but skipped the app-level validation step that registers `{rulesets, ...}`, so custom rulesets were later rejected as unknown or unavailable during rule execution.

## Changes

- Route discovered `elvis.config` and non-empty `{elvis, [...]}` entries from `rebar.config` through the same validation/loading path used by explicit config files.
- Preserve fallback behavior when no usable Elvis config file exists.
- Keep custom rulesets available during `rock/1` execution and clear them afterward to avoid leaking state between runs.
- Add a regression test proving a custom ruleset from default `elvis.config` is executed.

Closes #632 .

### DIFF
--- a/src/elvis_config.erl
+++ b/src/elvis_config.erl
@@ -135,24 +135,65 @@ default(Key) ->
             Value
     end.
 
+for(config) ->
+    maybe
+        {ok, {Source, AppConfig}} ?= default_app_config(),
+        case Source of
+            {file, File} ->
+                fetch_elvis_config_from(AppConfig, File);
+            fallback ->
+                from_static(config, {app, AppConfig})
+        end
+    else
+        {error, _} = Error -> Error
+    end;
 for(Key) ->
     maybe
-        AppDefault = default_for(app),
-        {ok, ElvisConfig} ?= consult_elvis_config("elvis.config"),
-        {ok, AppConfig} ?=
-            case ElvisConfig of
-                AppDefault ->
-                    % This might happen whether we fail to parse the file or it actually is []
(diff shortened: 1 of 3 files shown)


elbrujohalcon (approved):

paulo-ferraz-oliveira:
> Previously, custom rulesets worked when loading a config explicitly with elvis_config:from_file/1, but not when running through elvis_core:rock({config_file, default})

Is this the initial behaviour from when this was implemented, i.e. it was expected to skip in one case, not in the other?

My most recent changes might've broken something, but I'm curious as to the initial intent, and why tests didn't cover it. (no need to dig deep in this last curiosity)

elbrujohalcon:
I agree with @paulo-ferraz-oliveira … it's odd that custom rulesets were ignored in some scenarios.

PR #642, opened 2026-05-29
fix no files found causes an error

### LINKED ISSUE #631: No files found causes an error
## Bug Description

If you set up a standard elvis.config for your company's applications, and you want to check .hrl files so you add a section such as `#{ files => ["include/**/*.hrl", "src/**/*.hrl", "test/**/*.hrl"], ruleset => hrl_files }`, then Elvis will fail in an application that has no hrl files at all. It would be nice if you could at least add a key such as `allow_no_files = true`, so that individual apps don't need to be manually edited to comment out the hrl section.

## Reference

First reported by @richcarl at https://github.com/inaka/elvis/issues/609

bormilan:
## Summary

Add `allow_no_files` as an optional per-rule-group config key.

This lets shared `elvis.config` files include optional sections, such as `hrl_files`, without failing in applications that do not have matching files.

## Changes

- Add support for `allow_no_files => true` in config rule groups.
- Keep the default behavior unchanged: missing file matches still fail unless explicitly allowed.
- Validate `allow_no_files` as a boolean.
- Add regression coverage for empty `.hrl` glob matches.
- Document the new option in `README.md` and `MIGRATION.md`.

Closes #631 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,9 @@ documentation was updated.
 The `files` key is a list of glob patterns that tell `elvis_core` which files to analyse. Each
 pattern uses [`filelib:wildcard/1`](https://erlang.org/doc/man/filelib.html#wildcard-1), so you can
 use patterns like `"src/*.erl"` or `"apps/**/src/*.erl"`. Matching files are run through the
-pre-defined rules in the specified `ruleset`.
+pre-defined rules in the specified `ruleset`. By default, at least one glob must match a file; set
+`allow_no_files => true` on a config section to allow shared optional sections, such as header-file
+checks, to match no files.
 
 If you want to override the [pre-defined rules](#pre-defined-rules), for a given ruleset, you need
 to specify them in a `rules` key which is a list of items with the following structure
--- a/README.md
+++ b/README.md
@@ -151,9 +151,9 @@ documentation was updated.
 The `files` key is a list of glob patterns that tell `elvis_core` which files to analyse. Each
 pattern uses [`filelib:wildcard/1`](https://erlang.org/doc/man/filelib.html#wildcard-1), so you can
 use patterns like `"src/*.erl"` or `"apps/**/src/*.erl"`. Matching files are run through the
-pre-defined rules in the specified `ruleset`. By default, at least one glob must match a file; set
-`allow_no_files => true` on a config section to allow shared optional sections, such as header-file
-checks, to match no files.
+pre-defined rules in the specified `ruleset`. By default, at least one glob must match a file.
+To allow a config section to match no files - which is useful for shared optional sections like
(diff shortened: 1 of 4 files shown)


src/elvis_config.erl
```diff
@@ -762,6 +771,11 @@ all_files_globs_are_valid(FileGlobs) ->
             {error, "'files' is expected to be a non-empty list of non-empty strings."}
     end.
 
+is_config_boolean(_What, Value) when is_boolean(Value) ->
+    ok;
+is_config_boolean(What, _Value) ->
+    {error, io_lib:format("'~s' is expected to be a boolean.", [What])}.
+
```
bormilan, line 781: This should be refactored, since we already have `is_boolean`

elbrujohalcon (commented):
I think this is a good change, but I'll defer to @paulo-ferraz-oliveira for proper evaluation of this change since he's usually more strongly-opinionated around this area of the code.

paulo-ferraz-oliveira:
> without failing in applications that do not have matching files.

I think it's a good choice. I was under the impression I'd discussed this with @NelsonVides and somebody had implemented it, but apparently not 😄 I'll review... Thanks, @bormilan.

README.md
```diff
@@ -151,7 +151,9 @@ documentation was updated.
 The `files` key is a list of glob patterns that tell `elvis_core` which files to analyse. Each
 pattern uses [`filelib:wildcard/1`](https://erlang.org/doc/man/filelib.html#wildcard-1), so you can
 use patterns like `"src/*.erl"` or `"apps/**/src/*.erl"`. Matching files are run through the
-pre-defined rules in the specified `ruleset`.
+pre-defined rules in the specified `ruleset`. By default, at least one glob must match a file; set
+`allow_no_files => true` on a config section to allow shared optional sections, such as header-file
+checks, to match no files.
```
paulo-ferraz-oliveira:
```suggestion
pre-defined rules in the specified `ruleset`. By default, at least one glob must match a file.
To allow a config section to match no files - which is useful for shared optional sections like
header-file checks - set `allow_no_files => true`.
```

paulo-ferraz-oliveira: Minor readability improvement (?)

src/elvis_config.erl
paulo-ferraz-oliveira: I'd be Ok with naming it `is_boolean/2` simply, as it'd read better in e.g. `is_boolean(allow_no_files, AllowNoFiles)`.

paulo-ferraz-oliveira (approved):
Two minor (non-blocking) suggestions for potential improvement, but looks good to me.

elbrujohalcon (approved):

